PR #7, opened 2024-02-12
Increase nigiri time window to one year

vkrause:
This does not seem to have any adverse effects on the memory consumption or query performance on my local Switzerland example at least, but it's significantly more useful.

Going beyond one year brings little benefit as the GTFS schedules don't seem to be valid that far into the future anyway.


1Maxnet1:
> This does not seem to have any adverse effects on the memory consumption or query performance on my local Switzerland example at least, but it's significantly more useful.

Can confirm this. For a Germany + Belgium + Netherlands server I maintain, this had no recognizable effects. :+1:

PR #14, opened 2024-02-16
add docker support

exmatrikulator:


exmatrikulator:
I know, I saw the same problem(s).
Ansible is great, but I can't use it on my dev environment and it's hard to learn the first motis steps with this.
Yes, with docker there will be 2 option to run it and you have 2 options to maintain :(
No, the Ansible way should work as well.
It's not only the montis install path which is different, also the splitting ini.

I liked the idea that you can easy run a docker-compose with example data from Aachen out of the box.
The data is stored in volume and can be replaced by the europe.pbf and other feeds without changing the setup.

I needed it for my setup to look into motis, so I wanted to share it with you.
Maybe this is another/better way to deploy it.

chrisdebian:
Hi — I'm a new contributor and noticed this PR has been open since February 2024 with no resolution. Just wanted to gently ask whether there's a decision either way: merge (with any conditions), or close with a documented reason?

The reason I ask is that I'm planning to contribute a `CONTRIBUTING.md` and would like to document the infrastructure contribution scope clearly — knowing whether Docker configurations are in or out of scope for the maintained repo would help with that.

Thanks, Chris

jbruechert:
Our contributor documentation is at https://transitous.org/doc/.
Containers are available and are documented already.

PR #20, opened 2024-02-18
Initial support for GTFS-RT feeds

vkrause:
This introduces a new feed type "url", for things that MOTIS will fetch directly without us doing any pre-processing. This also gives feeds another type called 'spec' (following what transitland does), to indicate whether it's a GTFS or GTFS-RT feed (and later potentially others like GBFS and ParkenDD).

Add the Swiss national dataset to demonstrate this.

What's still missing is support for transitland GTFS-RT feeds.


src/metadata.py
```diff
@@ -66,12 +69,26 @@ def __init__(self, parsed: dict = None):
                         int(parsed["options"]["fetch-interval-days"])
 
 
+class UrlSource(Source):
+    url: str
+    authorization: str
```
jbruechert: (not a python expert myself)

I think it would make sense to init this to = None, otherwise the attribute doesn't really exist at runtime, and using it would throw an error.

motis/config.ini.j2
```diff
@@ -26,7 +26,8 @@ host=127.0.0.1
 [nigiri]
 first_day=TODAY
 num_days=365
-
+{% for feed in gtfsrt_feeds %}gtfsrt=rt-{{ feed.id }}|{{ feed.url }}|{{ feed.authorization }}
```
jbruechert: I get an error starting with the generated config file. I think the name of the rt feed needs to match the name of the static one.

```suggestion
{% for feed in gtfsrt_feeds %}gtfsrt={{ feed.id }}|{{ feed.url }}|{{ feed.authorization }}
```

vkrause: That's going to be a problem for the cases where this doesn't match as perfectly as in CH then though. E.g. VBN (https://www.vbn.de/service/entwicklerinfos/opendata-und-openservice) has GTFS-RT (as a rare exception in DE), but we have their GTFS data as part of the aggregated DELFI feed.

PR #26, opened 2024-02-20
motis-proxy: Allow to configure the list allowed endpoints

jbruechert:


vkrause:
Makes sense, +2. Might be useful to have the default set needed for the web ui in the example config?

PR #31, opened 2024-02-21
motis-proxy: Generate documentation of API

jbruechert:
This automatically serves API documentation for our filtered MOTIS API.

There is still lots of text to add, and the design could certainly be prettier. I have just barely customized the colors so far.

image


derhuerst:
If MOTIS supports serving an OpenAPI spec, I'd rather want to generate both `motis-proxy`'s data structures & documentation from that, given that MOTIS clearly is the source of truth here. What do you think?

(This is of course somewhat out of scope of this PR.)

derhuerst:
Maybe, with whatever tooling is available to generate Rust code from OpenAPI specs, we could patch this generated code code to remove those routes/requests that we want to block. The source of truth would then still be MOTIS, right?

PR #32, opened 2024-02-22
Update transland-atlas submodule

vkrause:
This brings the Swiss national feed update that should fix realtime data for us.


vkrause:
Ah, ok, I missed that and assumed this might be the reason why GTFS-RT is still largely failing to match the CH data...

PR #34, opened 2024-02-23
motis-proxy: Implement route rate limit

jbruechert:
Routing requests can be pretty heavy, so we should stop people from making too many, whether on purpose or not

### DIFF
--- a/ansible/roles/motis-proxy/files/Rocket.toml
+++ b/ansible/roles/motis-proxy/files/Rocket.toml
@@ -2,9 +2,11 @@
 #
 # SPDX-License-Identifier: AGPL-3.0-or-later
 
+[release]
+ip_header = "X-Forwarded-For"
+
 [release.proxy]
 # TODO: Add /address and /ppr/route once we have it set up
 allowed_endpoints = [ "/intermodal", "/guesser",
                       "/railviz/get_trains", "/railviz/get_trips", "/railviz/get_station",
                       "/lookup/schedule_info", "/gbfs/info", "/trip_to_connection" ]
-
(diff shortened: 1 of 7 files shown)


vkrause:
Certainly makes sense to have that infrastructure, where exactly to draw the line we have to see once this is actually used more widely I guess.

ansible/roles/motis-proxy/files/Rocket.toml
```diff
@@ -2,9 +2,11 @@
 #
 # SPDX-License-Identifier: AGPL-3.0-or-later
 
+[release]
+ip_header = "X-Forwarded-For"
```
derhuerst, line 6: Just FYI: RFC 7239 replaced the non-standardised `X-Forwarded-*` headers with the [`X-Forwarded` header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Forwarded#transitioning_from_x-forwarded-for_to_forwarded).

ansible/roles/nginx/templates/transitous.conf.j2
```diff
@@ -41,6 +41,8 @@ server {
     # MOTIS Proxy
     location /api/ {
         proxy_pass http://localhost:8000/;
+
+        proxy_set_header X-Forwarded-For $remote_addr;
```
derhuerst, line 45: Just FYI: RFC 7239 replaced the non-standardised `X-Forwarded-*` headers with the [`X-Forwarded` header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Forwarded#transitioning_from_x-forwarded-for_to_forwarded).

jbruechert: Cool, I'm open to replacing this. It's just editing the nginx config and the Rocket.toml

PR #35, opened 2024-02-23
Rootless devcontainer

PartTimeDataScientist:
This PR configures the `Containerfile` to run rootless and adds a basic `.devcontainer` configuration for VS Code.

This allows to open the cloned repository in VS Code and (with the Dev Containers extension installed) VS Code will automatically offer to open the project in a container where you can start developing.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,48 @@
+{
+	"name": "Transitous",
+	"build": {
+		"dockerfile": "../ci/container/Containerfile",
+		"context": ".."
+	},
+
+	"runArgs": [
+		"--userns=keep-id:uid=1000,gid=1000"
+	   ],
+
+	// Configure tool-specific properties.
+	"customizations": {
+		// Configure properties specific to VS Code.
+		"vscode": {
+			// Set *default* container specific settings.json values on container create.
+			"settings": { 
+				"python.defaultInterpreterPath": "/usr/local/bin/python",
+				"python.linting.enabled": true,
+				"python.linting.pylintEnabled": true,
+				"python.formatting.autopep8Path": "/usr/local/py-utils/bin/autopep8",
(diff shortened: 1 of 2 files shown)


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,48 @@
...
+		"--userns=keep-id:uid=1000,gid=1000"
+	   ],
+
+	// Configure tool-specific properties.
+	"customizations": {
+		// Configure properties specific to VS Code.
+		"vscode": {
+			// Set *default* container specific settings.json values on container create.
+			"settings": { 
+				"python.defaultInterpreterPath": "/usr/local/bin/python",
```
jbruechert: Shouldn't this be `/usr/bin/python` inside the container?

PartTimeDataScientist: You are right, this was a classical copy&paste error 🤷🏻‍♂️ 

I've corrected the binary path to `/usr/bin/python3` and removed the settings for all the tools that are not even installed in the container right now. 

We could think about if and which formatters and linters should be added although this *might* be an argument to separate the devcontainer image from the CI image on the long run...

PartTimeDataScientist:
> To fix the reuse check, it should be enough to create a file `.devcontainer/devcontainer.json.license` with
> 
> ```
> SPDX-FileCopyrightText: 2024 Name <email>
> SPDX-License-Identifier: CC0-1.0 / AGPL-3.0-or-later (choose one)
> ```

I have added the license information as "comments" in the first line. Works for VS Code

> To fix the reuse check, it should be enough to create a file `.devcontainer/devcontainer.json.license` with
> 
> ```
> SPDX-FileCopyrightText: 2024 Name <email>
> SPDX-License-Identifier: CC0-1.0 / AGPL-3.0-or-later (choose one)
> ```

I've added the license information as comments in the first lines. Works for VS Code - if the REUSE check is fine this should be ok as well I think. (I don't like "sidecar" files - plenty of hassle with raw files in photography... 😀)

PR #39, opened 2024-02-25
Switzerland: change assigned maintainer

pka:
As discussed with @vkrause


pka:
@derhuerst Changes are probably quite rare for this particular feed. I'm fine with PRs, but it makes maybe sense to give all feed maintainers the same permissions.

PR #41, opened 2024-02-25
Add data for Vigo, Spain

Gallaecio:
(not tested locally)

### DIFF
--- a/feeds/es.json
+++ b/feeds/es.json
@@ -1,0 +1,21 @@
+{
+    "maintainers": [
+        {
+            "name": "Adrian Chaves",
+            "github": "Gallaecio"
+        }
+    ],
+    "sources": [
+        {
+            "name": "VITRASA",
+            "type": "http",
+            "url": "https://datos.vigo.org/data/transporte/gtfs_vigo.zip",
+            "license": {
+                "spdx-identifier": "ODC-By-1.0"
+            },
+            "options": {
+                "fetch-interval-days": 1
+            }
+        }
+    ]
+}


feeds/es.json
```diff
@@ -0,0 +1,29 @@
+{
+    "maintainers": [
+        {
+            "name": "Adrian Chaves",
+            "github": "Gallaecio"
+        }
+    ],
+    "sources": [
+        {
+            "name": "Renfe — Alta velocidad, larga distancia y media distancia",
```
jbruechert: I should probably document this, but this should ideally not contain spaces, as it is used as a file name

```suggestion
            "name": "Renfe",
```

feeds/es.json
```diff
@@ -0,0 +1,29 @@
...
+            "type": "http",
+            "url": "https://ssl.renfe.com/gtransit/Fichero_AV_LD/google_transit.zip",
+            "license": {
+                "spdx-identifier": "CC-BY-4.0"
+            }
+        },
+        {
+            "name": "VITRASA",
+            "type": "http",
+            "url": "https://datos.vigo.org/data/transporte/gtfs_vigo.zip",
```
jbruechert, line 20: From the CI log, it looks like this feed contains a stop time for a non-existent trip.
Adding the fix option should help.

```suggestion
            "url": "https://datos.vigo.org/data/transporte/gtfs_vigo.zip",
            "fix": true,
```

feeds/es.json
Gallaecio: Oh, right, I read that, then forgot.

Should it be a unique name, e.g. `es-vigo`?

jbruechert: It only needs to be unique inside the json file, the final filename is "{name of the json file}-{name of the feed}"

Gallaecio:
It indeed looks like quite a few lines are missing, I’ll look into it eventually.

PR #44, opened 2024-02-26
Add IT sources

Fale:


feeds/it.json
```diff
@@ -0,0 +1,26 @@
...
+        {
+            "name": "Fabio Alessandro Locati",
+            "github": "fale"
+        }
+    ],
+    "sources": [
+        {
+            "name": "ATM",
+            "type": "http",
+            "url": "https://dati.comune.milano.it/gtfs.zip",
```
jbruechert: This is available on Transitland, as `f-u0nd-comunedimilano`. Please switch to the transitland-atlas type.
You can then remove the license info, it will automatically be fetched from there.

PR #47, opened 2024-02-26
italy: Add Trenitalia

jbruechert:
@Fale Would you be interested in looking after this as well, or should I add myself to maintainers?


Fale:
I don't mind looking at this as well, but - despite the name - this is not the GFTS of TreniItalia, but only the TreniItalia part in the Tuscany, Piedomont, and Sardinia regions.
I think we can merge this one, and then I can do additional searches to see if we can find the data for the other regions as well!

PR #50, opened 2024-02-26
ansible: Allow building a container

jbruechert:
The effort for this is really small, and it allows people to run our setup in their container based environments.


jbruechert:
We can revisit this if needed, closing for now

PR #54, opened 2024-02-26
motis: Enable duplicate matching

jbruechert:


felixguendling:
Loading performance with this option for timetables with large overlap is not really well tested. It works well for the MOTIS demo instance (where only a few long distance trains overlap).

This `vecvec<K, V>` data structure is not really made for many `push_back` operations. If you want to use this for large overlaps it might make sense to try a different data structure (like `mutable_fws_multimap`).

https://github.com/motis-project/nigiri/blob/d806245b6b313f143e1705a01df4dc7d9abbbe19/src/loader/build_footpaths.cc#L152

https://github.com/motis-project/nigiri/blob/d806245b6b313f143e1705a01df4dc7d9abbbe19/include/nigiri/timetable.h#L356

vkrause:
Loading times still seem ok here locally (but then we don't really have large overlaps yet). However it's not filtering the DE <-> NL Regionalexpress duplicates that triggered this originally either it seems.

Your optimization suggestion however seems to be the solution for a problem we have with GTFS-RT updates :)

felixguendling:
Maybe it makes sense to create a minimal example with those trips? Usually what I do is:
- click on the trains on the web UI, get the `trip_id` from the URL
- grep for the corresponding `trip_id` in the GTFS data and create a `stop_times.txt` with only times from this trip
- reuse all other files (`stops.txt`, `routes.txt`, etc. from the GTFS feed(s) and call `gtfs-tidy`
- voilà - you have a minimal testing example where you can print some output on where [the code](https://github.com/motis-project/nigiri/blob/bf9b9e98ef5ddf35a1d37defd0af55e459351f18/src/loader/build_footpaths.cc#L90-L218) bails out and decides that they can't be duplicates

Maybe it would even make sense to write a very simple command line tool to do this extraction (like `gtfs-extract feed.zip trip_id1 trip_id2 ...`) to be able to generate minimal examples and debug them quickly.

derhuerst:
remotely related discussion in the GTFS repo on referencing other "original"/proprietary/external trip IDs and/or route IDs: https://github.com/google/transit/pull/534

PR #59, opened 2024-02-28
Add Australian Queensland Translink Feeds

Steve-Tech:
Hi, I just added Australian Translink Feeds.

They have real-time GTFS-RT feeds too with vehicle tracking, but I wasn't sure if this project makes use of them.

Thanks,
Steve

### DIFF
--- a/feeds/au.json
+++ b/feeds/au.json
@@ -1,0 +1,208 @@
+{
+    "maintainers": [
+        {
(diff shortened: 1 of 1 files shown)


Steve-Tech:
Oh nevermind, I just found [o-r-translink](https://www.transit.land/operators/o-r-translink). I'll fix it up in a few hours.

PR #60, opened 2024-02-28
Create dk.json

claus-chr:
Add support for rejseplanen.dk - Danish trains and busses

### DIFF
--- a/feeds/dk.json
+++ b/feeds/dk.json
@@ -1,0 +1,18 @@
+@@ -0,0 +1,17 @@
+{
+    "maintainers": [
+        {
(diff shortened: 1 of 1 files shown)


feeds/dk.json
```diff
@@ -0,0 +1,18 @@
+@@ -0,0 +1,17 @@
```
derhuerst: This is probably a leftover of some Git operation. It does not belong in the file, right?

jbruechert:
```suggestion
```

claus-chr: You are right. I'm sorry I didn't spot it.

PR #71, opened 2024-03-02
Improve manual

Fale:


README.md
```diff
@@ -89,14 +89,20 @@ You can also use the container described below.
...
+
+Proceed by building the container:
 ```bash
 podman build ci/container/ -t transitous -f ci/container/Containerfile
 ```
 
 Enter the container:
 ```bash
-podman run -it -p 8080:8080 -v $PWD:/transitous -w /transitous transitous
+podman run -it --privileged -p 8080:8080 -v $PWD:/transitous -w /transitous transitous
```
jbruechert: What is this needed for?

README.md
```diff
@@ -89,14 +89,20 @@ You can also use the container described below.
 Running a local instance of the transitous setup can be useful for debugging.
 The easiest way is to use the same container image that we use for fetching and importing the data on the CI.
 
-First, build the container:
+First, ensure that you have the Git submodules:
+
+```bash
+git submodule init && git submodule update
```
jbruechert:
```suggestion
git submodule update --remote --checkout --init
```

This will pull the most up to date version independently of what is pinned in the git repo.
For transitland-atlas I think that makes sense, and we already do it on the CI.

README.md
Fale: To ensure that within the container the user has enough privileges (root). Since the `ci/fetch-feeds.py` is owned, if the container is run as a normal user you need to be root within the container to run it

README.md
Fale: Thanks for the suggestion :)

README.md
Fale: Actually, a different way might be better

jbruechert: I can run the script without that, also when the user id inside and outside the container doesn't match.
Since the permissions are set to -rwxr-xr-x, every account should be already allowed to execute it.

I think this might be something specific to your podman setup.

jbruechert: The podman docs say:
> Using --userns=auto when starting new containers will not work as long as any containers exist that were started with --userns=keep-id.

so that doesn't seem optimal either.

Are you using selinux btw?

Fale: What distro are you using? My assumption is that in your distro the `--userns=keep-id` gets added with an alias or config at every run. If you try to do an `ls -alhZ` in the container, who owns the files? if no `--userns=keep-id` is used, they should be owned by root

jbruechert: I'm using debian (testing).
```
transitous@c2032ff00955:/transitous$ ls -alhZ src/
total 24K
drwxr-xr-x 1 root root ?  182 Mar  1 17:38 .
drwxr-xr-x 1 root root ?  380 Mar  1 22:33 ..
drwxr-xr-x 1 root root ?  144 Feb 28 09:57 __pycache__
-rwxr-xr-x 1 root root ? 5.8K Mar  1 17:38 fetch.py
-rwxr-xr-x 1 root root ? 2.2K Feb 28 09:30 generate-attribution.py
-rwxr-xr-x 1 root root ? 2.3K Feb 28 09:30 generate-motis-config.py
-rw-r--r-- 1 root root ? 2.7K Feb 27 17:27 metadata.py
-rw-r--r-- 1 root root ? 1.5K Feb 28 09:30 transitland.py
```

The scripts can nevertheless be executed.

I'll test on a fedora setup with selinux enabled and report back if that works.

Fale: There are two different situations here:
1. Being on Debian, you have no SELinux running, so since `fetch.py` has `o+rx` it can be executed. If you have SELinux running that's not enough (hence the `:Z`)
2. The `transitous` user probably has enough privileges to write `/transitous/out` and `/transitous/downloads`, which is probably due to how those were setup on your system. Usually folders get created as 755 in $HOME which would not be enough. The `--userns=keep-id` allows to avoid adding `o+w` for those folders

jbruechert: That explains it. I'm find with adding this as long as it doesn't interfere with other containers set up differently.

Fale: It should not be a problem, since Podman should drop all parameters/options that are not applicable to the current system (eg: the `:Z` on a system without SELinux should have no effect)

PR #75, opened 2024-03-02
Add More Canadian Transitland Feeds

Pi-Cla:
Add every Transitland feed I could find which is not in B.C, Ontario, or Quebec (as those have a lot of feeds...)


nekohayo:
While we're here, this might be relevant to your interests: https://github.com/transitland/transitland-atlas/issues/1024

nekohayo:
Also, why not add the B.C., Ontario and Quebec feeds from the [Canadian feeds](https://transitfeeds.com/l/32-canada)? Those would be very useful IMHO, they are the most populous metropolitan areas.

In addition / as an alternative to the transitfeeds website listing, there is also an official (govt provided) overall view of the Québec province's various cities ("ville") offering feeds: https://www.rtcquebec.ca/en/open-data

PR #76, opened 2024-03-02
Add Piedmont feeds

Fale:


Fale:
Strange that it timesout. I've tested on my machine and all those urls work :(

Fale:
I would not be surprised if they blocked some IPs (such as GitHub Action ones) that maybe have been doing lots of requestes

Fale:
I guess it would be better to whitelist the specific dest urls or source IPs to ensure that it does not become an open proxy. I guess that it could be an idea to cache all urls, not only those problematic ones, since if we hit every single GTFS source at every PR/commit, many more might block us in the future (and we are probably nice to them even if they do not block us)

Fale:
> We already support If-Modified-Since and Last-Modified, we shouldn't use much resources on the provider's side.

You are assuming that the server provides good preflight info, I think. If that is the case, it will hardly work as expected, since Public sector (at least in some countries, such as Italy) has very - on average - very poor IT setups and configs

Fale:
I'll try, though I doubt it will work, since - from the URL - it seems like they have a software (probably that they bought from some third party company) that produces the GTFS on request.

This night I was thinking about this issue. I was thinking that probably the best long-term solution would be to create a smart proxy that is able to cache GTFS files, provide info on when the file was retrieved and serve the old version until a new one is successfully downloaded. This with the addition of a background batch process that refreshes the feeds every X amount of time, would probably solve 99.9% of those kind of issues.

Fale:
Yes, I was also thinking about an external project. If I have some time in the next few days I'll give a shot at it :)

PR #82, opened 2024-03-03
Add ÖBB feed for Austria

vkrause:
That's at least giving us long/medium distance coverage, everything local is still missing.

### DIFF
--- a/feeds/at.json
+++ b/feeds/at.json
@@ -1,0 +1,18 @@
+{
+    "maintainers": [
+        {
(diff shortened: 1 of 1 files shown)


felixguendling:
With MOTIS, it should be possible to load two timetables, one for `current_year` and one for `current_year+1`. Since the change date might be different for each region (Germany for example usually has mid of December), the feed would also need to provide a date for the timetable change (e.g. to match the real-time data on the correct timetable if both are loaded).

Issue #29, opened 2024-02-20
Create a website


jbruechert:
Approximate contents:
* Very short description of the service / slogan
* A search field to try our routing, should just forward the request to the motis web interface. This may need extending motis with query parameters for this purpose. Maybe a simplified version of what https://www.vy.no/en#journey-planner does
* Further down some API example, link to motis docs / our docs in case we have some by then.

PartTimeDataScientist:
The rendering of the website ist done in the Github actions. The gh-pages branch does not contain a docs folder at all...

PartTimeDataScientist:
If you set this up to deploy I think Github is trying to render a Jekyll site. We are talking about option 3 of this tutorial:

https://roelofjanelsinga.com/articles/how-to-set-up-automatically-deploy-website-github-pages/

PartTimeDataScientist:
> Oh sorry, I misread that as deploying the contents of the branch to /docs, changed it to / now. It's probably relatively easy to move to /docs, right?
> 
> I'd like to do something else for the front page in the future.

That worked, website is now available under the links from the README. A better front page should be easy to integrate:

In the front matter of your index.md, you need to specify the template to use:

```
---
title: Title
template: home.html
---
```
Source: https://stackoverflow.com/questions/63438788/how-to-add-a-landing-page-to-a-mkdocs-doc-site-using-mkdocs-material)

PR #96, opened 2024-03-05
Add website and license page to Github-Pages

### LINKED ISSUE #18: Generate license page
We can now generate a license summary as json using `./src/generate-attribution.py`.
In order to make it easy for apps to link to it, we should set up a static site generator to make this data more readable for humans.

PartTimeDataScientist:
This PR adds a basic mkdocs(-material) website (basically a copy of the README.md file) and modifies the generate-attribution.py file so that it also generates a licenses.md file for mkdocs automatically.

It also adds a GH-action to push the site to GH-pages on every push to main. For the result see: https://parttimedatascientist.github.io/transitous/

This PR should close #18 and #29.

### DIFF
--- a/feeds/it.json
+++ b/feeds/it.json
@@ -131,6 +131,24 @@
             "name": "Toscana-Trenitalia",
             "type": "transitland-atlas",
             "transitland-atlas-id": "f-sp-trenitaliaspa"
+        },
+        {
+            "name": "Veneto-Venezia-Bus",
+            "type": "http",
+            "url": "https://actv.avmspa.it/sites/default/files/attachments/opendata/automobilistico/actv_aut.zip",
+            "license": {
+                "spdx-identifier": "CC-BY-3.0",
+                "url": "https://dati.venezia.it/?q=licenza/cc"
+            }
+        },
+        {
+            "name": "Veneto-Venezia-Ferries",
+            "type": "http",
+            "url": "https://actv.avmspa.it/sites/default/files/attachments/opendata/navigazione/actv_nav.zip",
+            "license": {
+                "spdx-identifier": "CC-BY-3.0",
+                "url": "https://dati.venezia.it/?q=licenza/cc"
+            }
         }
     ]
 }


PartTimeDataScientist:
Forgot to mention: I need a hint on how to properly add the license information for the mkdocs-material theme (MIT-license) to SPDX without touching 13k files.

Is it possible to add a sidecar file referring to all files within a folder?

vkrause:
> Forgot to mention: I need a hint on how to properly add the license information for the mkdocs-material theme (MIT-license) to SPDX without touching 13k files.
> 
> Is it possible to add a sidecar file referring to all files within a folder?

I'd suggest the global DEF5 file for that (see https://reuse.software/spec/#dep5), that's currently used to set the license on the feed JSON files as well.

PartTimeDataScientist:
> I am a bit cautious of using mkdocs as it seems to be adding a lot of files (over 5k) which can not be manually verified by hand. Also the REUSE compliance check failing is another cause for concern as well as esbuild executing in a 289 line postinstall script https://socket.dev/npm/package/esbuild/files/0.20.1/install.js
> 
> Perhaps we could avoid these issues by using static site generators such as Zola or Hugo to generate webpages instead?

You could get away with a *much* smaller number of files if using "vanilla" mkdocs instead of mkdocs-material or by reducing the number of mkdocs-material extensions that are not currently used. I'd assume the same holds for Hugo (and I don't know about Zola).

Looking at the numbers: From the 15110 files in the mkdocs-site folder there are 13137 .svg icons from fontawesome, material and octicons in `material/templates/.icons`. That still leaves about 2000 other files but my guess is that you'll reach somewhat similar numbers once you use a theme (and some extensions) with `whatever-site-generator`.  Just checked another frequently used theme ([docsy](https://github.com/google/docsy)) for Hugo which has about 800 files.

So to me the question seems to be more about what is planned in the long run: Is it a styled documentation page or a rather simple site with just the basic (e.g. license and developer) information? Both is fine with me (and could be changed later), although I prefer somewhat appealing websites in general.

PR #111, opened 2024-03-08
Add British Columbian feeds and split Canadian feeds

Pi-Cla:
Feeds which are exclusively in:
- British Columbia go to ca-bc.json
- Ontario go to ca-on.json
- Quebec go to ca-qc.json
The rest will go to ca-other.json including VIA rail as it crosses regional boundaries within Canada


mlundblad:
One idea:

Using [cc].json for feeds covering multi state/provinces in these cases with big countries.

This way it will be sorted, in this case, like:

ca.json
ca-bc.json
ca-on.json
ca-qc.json

Inssted of having the -other one getting "mixed in".

PR #119, opened 2024-03-09
Add Finish national railway realtime feed

vkrause:
Needs MOTIS v0.11.26 to work.


feeds/de.json
```diff
@@ -17,6 +17,16 @@
             "options": {
                 "fetch-interval-days": 2
             }
+        },
+        {
+            "name": "DELFI",
+            "type": "url",
+            "spec": "gtfs-rt",
+            "url": "http://gtfsr.vbn.de/gtfsr_connect.bin",
```
jbruechert: This doesn't look like Finland :D

vkrause: oops, that was the wrong file obviously :D

Issue #121, opened 2024-03-10
[Australia] Add GTFS feed for Transperth, Western Australia


mexon:
Static GTFS data for Western Australia's main public transport operator Transperth is available here:

https://www.transperth.wa.gov.au/TimetablePDFs/GoogleTransit/Production/google_transit.zip

Licensing information is described here:

https://www.transperth.wa.gov.au/About/Spatial-Data-Access

I can't find any mention of public access to realtime data, even though they certainly do have this information available for their own app.

PR #122, opened 2024-03-10
Add West Australian Transperth feed (#121)

mexon:
Add feed for Transperth to the Australia feed file.

(I do not know if I'm supposed to make PRs like this, just thought it'd be cool to see my local public transport on the demo.)

### DIFF
--- a/feeds/au.json
+++ b/feeds/au.json
@@ -128,6 +128,11 @@
             "name": "Translink-WHT",
             "type": "transitland-atlas",
             "transitland-atlas-id": "f-rk3s-qconnect~whitsundaytransit"
+        },
+        {
+            "name": "Transperth",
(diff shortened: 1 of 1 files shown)


mexon:
> We require a (somewhat) local person as maintainer for each feed, so we potentially have someone who can check the plausibility of the data in case of issues with the feed. If you'd like to do that, please add yourself to the maintainers list at the beginning of the file.

I am somewhat local ;)  I grew up in Perth and can (and did!) check the plausibility of the data.  I live in Berlin now, but visit regularly and can double-check with people on the ground.

It feels slightly wrong to list myself as a maintainer of the whole file, which right now is dominated by Queensland services.  This is a big country!  But anyway, done.

PR #127, opened 2024-03-12
fr: Add CTS (Stasbourg)

mlundblad:
Also added Alexandre Franke, as he agreed
on co-maintaining the feed.

Asked about it in the GNOME Maps Matrix channel.


mlundblad:
> It looks like Paris is now completely removed, that isn't on purpose right?

Yeah,. I messade up a bit when synching fork and rebasing
Should be correct now.

PR #131, opened 2024-03-12
Add transitland-atlas-proxy

Fale:
I've created [gtfsproxy](https://github.com/Fale/gtfsproxy) which I deployed [here](https://gtfsproxy.fwan.it).

What do you think about this approach?


Fale:
Like this?

src/metadata.py
```diff
@@ -53,11 +53,13 @@ class TransitlandSource(Source):
     transitland_atlas_id: str
     options: HttpOptions = HttpOptions()
     url_override: str
+    proxy: bool
 
     def __init__(self, parsed: dict):
         super().__init__(parsed)
         self.transitland_atlas_id = parsed["transitland-atlas-id"]
         self.url_override = parsed.get("url-override", None)
+        self.proxy = parsed.get("proxy")
```
jbruechert:
```suggestion
        self.proxy = parsed.get("proxy", false)
```

Fale:
I've added TPER :). I have to make it slightly more flexible (hopefully by this weekend) to make it fully work with the open PRs

PR #145, opened 2024-03-15
Added VLC city metro

Bro66666:


feeds/es.json
```diff
@@ -41,6 +41,12 @@
             "type": "transitland-atlas",
             "transitland-atlas-id": "f-metro~malaga"
         },
+        {
+            "name": "valencia-metro",
+            "type": "url",
```
jbruechert: type: url is only for things that we can't fetch in advance (currently only realtime data).

Bro66666: Right. Will take that into account from now on. "http" doesn't work though

Bro66666:
> This feed seems to have broken https set up. We should probably add a flag the code to ignore this.

So should I close the request?

PR #158, opened 2024-03-18
Added more city bus routes

Bro66666:
Including up to city of Madrid. Removed older feeds


feeds/es.json
```diff
@@ -54,18 +54,6 @@
             "transitland-atlas-id": "f-ezj-informaciónoficial~consorcioregionaldetransportesdemadrid",
             "fix": true
         },
-        {
```
jbruechert, line 57: Is the removal of these feeds on purpose?

Bro66666: Yes, to avoid duplicates. The sources I am using now (from the Spanish Ministry of Transport) provide more reliable and unified feeds. Nearly all the old ones will end up being removed in favour of the ones provided by the ministry.

If you check the last two entries of the new bit, they correspond to the old feeds that were removed.

PR #159, opened 2024-03-18
Poland: add Gorzów Wielkopolski

Gnomasz:
Added Gorzów Wielkopolski's feed from Transitland to pl.json


1Maxnet1:
@Gnomasz Would you be okay with being added as a maintainer?

Otherwise I would propose to merge it anyway, as there are already 3 maintainers for Poland, who could have a look in case something breaks.

feeds/pl.json
```diff
@@ -158,6 +158,11 @@
                 "spdx-identifier": "CC-BY-4.0",
                 "url": "https://spdx.org/licenses/CC-BY-4.0.html"
             }
+        },
+        {
+            "name": "Gorzów Wielkopolski",
```
jbruechert, line 163:
```suggestion
            "name": "Gorzów-Wielkopolski",
```

Gnomasz:
> @Gnomasz Would you be okay with being added as a maintainer?
> 
> Otherwise I would propose to merge it anyway, as there are already 3 maintainers for Poland, who could have a look in case something breaks.

Hi,
I'm afraid that I won't be able to keep track of it.

jbruechert:
That's alright, I think we have enough people for that region :)

Now the problem is that the source of the feed seems to have vanished. So I'm just leaving this open in case that changes again.

Lach-anonym:
#1189 has been merged, so this one is obsolete

PR #161, opened 2024-03-18
Adding regional and city bus routes, trains and 1st ferry

Bro66666:
Moved ALSA bus feed so that the list is in more or less alphabetical order, and removed **duplicates**:

* Valencia urban buses
* Cercanias RENFE

### DIFF
--- a/feeds/es.json
+++ b/feeds/es.json
@@ -19,12 +19,6 @@
             }
         },
         {
-            "name": "renfe-cercanias",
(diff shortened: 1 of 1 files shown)


Bro66666:
> (I clicked the wrong button, didn't mean to remove the draft status)

Don't worry. I keep it as draft until I am sure all tests pass, just in case. All tests passed, so we're good.

PR #169, opened 2024-03-19
Use thread pool for fetch

### LINKED ISSUE #154: Make fetch failures non-fatal and spam maintainers instead
Requirements:
* A GitHub issue with, depending on the error the http body and error code, or the gtfstidy output should be opened.
* There should be at most one issue per failing feed. The easiest way to do that is probably prefixing the title with the feed id.
* It should not depend on GitHub Actions specifically, we are moving away from that.
* The feed maintainer (as specified in the json) should be assigned or mentioned in the issue.

This needs to be implemented in `src/fetch.py`, but the GitHub specific parts can of course be in a new file.

### LINKED ISSUE #166: Parellelize feed processing
gtfstidy is not using multi-threading itself, so running up to $(nproc) gtfstidy jobs in parallel would speed things up a lot.
The same applies to the downloads as well.

anbraten:
closes #166
closes #154


anbraten:
@jbruechert Thanks for finishing. Hope it helped.

PR #176, opened 2024-03-25
Enable access to the /lookup/geo_station endpoint

vkrause:
#172 added support for this but missed to actually enable access.


vkrause:
> Should be deployed now

Works, thank you!

PR #185, opened 2024-03-28
Update fr.json : Add Nouvelle Aquitaine Mobilités

kalon33:
Add Nouvelle-Aquitaine Mobilités static feed (public transportation feed for the whole Nouvelle Aquitaine, France) and associated RT feed from TBM (one agency aggregated in the Nouvelle-Aquitaine Mobilités feed).


kalon33:
@jbruechert noted and done!

Issue #198, opened 2024-03-31
Enable adr in place of address in MOTIS


kalon33:
Could you please enable `adr` module in place of `address` in MOTIS? It appeared in latest versions and allows to calculate itineraries using precise addresses (house numbers...) instead of only street names.

kalon33:
Ok, I will have a look at this. Should we also enable odr to have lighter routing than osrm (I imagine its weight on the storage side is the reason it has been disabled)?

kalon33:
Is this PR https://github.com/public-transport/transitous/pull/200 what you asked for @jbruechert?

felixguendling:
In order to compute door-to-door journeys, also a street routing (e.g. `osr`) needs to be enabled.

felixguendling:
As the `osr` routing data now fits into ~13GB RAM (for the whole planet), there is no need to memory map it anymore. I had hoped that locking the memory [via `mlock`](https://github.com/motis-project/osr/blob/master/include/osr/ways.h#L142-L160) would force it to be loaded but it looks like it only prevents unpaging memory once it's loaded. So maybe it would help to trigger a read operation on each memory page to load the data from disk to memory. Another option would be to not use memory mappings for routing data in the first place. I don't think this is a use case for faster storage.

PR #200, opened 2024-03-31
Update MOTIS version and enable adr module from latest MOTIS

kalon33:
Enable adr module in MOTIS, and update to latest version to be able to do so (https://github.com/public-transport/transitous/issues/198).


kalon33:
> Thanks, it would be nice to have a flavour of the config file which just loads the osm data, but no feeds, so we can import that separately (and less often)

You mean in a separate file (for instance `config.ini.osm`) or in the current template file with more conditions, and a `flavour=osm` for example?

kalon33:
@jbruechert I just committed something, but hadn't time to test it yet.

kalon33:
For info, I added osr module as suggested by @felixguendling in #198.

kalon33:
@jbruechert any news about this? Did everything went as expected on the resource needs side?

felixguendling:
Until then you can try it on https://europe.motis-project.de/
It doesn't update the GTFS datasets but it's a PoC.

kalon33:
@felixguendling @jbruechert Any idea what URL I should pass in the MOTIS Android app to be able to use Transitous? I'm struggling finding the proper one and the current web interface is not very usable on a smartphone.

GerdC:
@kalon33 
> Any idea what URL I should pass in the MOTIS Android app to be able to use Transitous? I'm struggling finding the proper one and the current web interface is not very usable on a smartphone.

As a proof of concept, I have Transitous running with by iOS app and I use this base URL: "https://routing.spline.de/api/?target=/"
Note the additional "?target=/" that https://europe.motis-project.de/ doesn't need.

kalon33:
> @kalon33 
> > Any idea what URL I should pass in the MOTIS Android app to be able to use Transitous? I'm struggling finding the proper one and the current web interface is not very usable on a smartphone.
> 
> As a proof of concept, I have Transitous running with by iOS app and I use this base URL: "https://routing.spline.de/api/?target=/"
> Note the additional "?target=/" that https://europe.motis-project.de/ doesn't need.

Thanks, but this doesn't seem to work for the Android app.

felixguendling:
> Thanks, but this doesn't seem to work for the Android app.

If we're talking about the "official" MOTIS Android app (in `ui/android`), it's probably some incompatible binary protocol change. The app uses the binary Flatbuffers protocol via websockets to communicate with the server. Since we use a very old Flatbuffers version (and updating would be very tedious), I would not recommend using this interface anymore.

We will soon start working on a new app. Until then, it's probably like MOTIS doesn't have any "official" app.

Usually the app breaks if I make a incompatible protocol change. I am not aware of any protocol change at the moment, but maybe I missed it. At least.. if that's the case, generating the regenerating the Java classes using `flatc` would probably help. But I'm not sure if it's worth it for the reasons mentioned above.

kalon33:
@jbruechert any news about this? I added a suggestion from @felixguendling to fully take advantage of the new colored map style, if activated.

PR #204, opened 2024-04-02
Update eu.json: Add Eurostar and Thalys trains.

kalon33:
Add Eurostar and Thalys trains to European feeds.


kalon33:
To my knowledge, individual realtime coverage for Eurostar isn't available. It has been requested to Eurostar, the feed provider on https://transport.data.gouv.fr/datasets/eurostar-gtfs but without any answer for now.

kalon33:
@jbruechert Thanks! How often are the feeds refreshed in Transitous once they are online?

kalon33:
Ok, not needed for me right now, but as for example SNCF feeds are updated daily and id changes (making a large number of mismatches with GTFS-RT feeds if not updated), I was wondering what the current refresh rate is :)

Issue #201, opened 2024-03-31
North American coverage status


mlundblad:
Tracking issue for coverage in North America (inspired by #114 )

- [] 🇧🇸 Bahamas
- [] 🇧🇿 Belize
- [] 🇨🇦 Canada: Via Rail
-- [] 🏴󠁣󠁡󠁡󠁢󠁿 Alberta
-- [] 🏴󠁣󠁡󠁢󠁣󠁿 British Columbia: #75 
-- [] 🏴󠁣󠁡󠁭󠁢󠁿 Manitoba
-- [] 🏴󠁣󠁡󠁮󠁢󠁿 New Brunswick
-- [] 🏴󠁣󠁡󠁮󠁬󠁿 Newfoundland and Labrador
-- [] 🏴󠁣󠁡󠁮󠁴󠁿 Northwest Territories
-- [] 🏴󠁣󠁡󠁮󠁳󠁿 Nova Scotia
-- [] 🏴󠁣󠁡󠁮󠁵󠁿 Nunavut
-- [] 🏴󠁣󠁡󠁯󠁮󠁿 Ontario: #58 
-- [] 🏴󠁣󠁡󠁰󠁥󠁿 Prince Edward Island
-- [] 🏴󠁣󠁡󠁱󠁣󠁿 Quebec: Montreal, Quebec City, Trois-Rivières
-- [] 🏴󠁣󠁡󠁳󠁫󠁿 Saskatchewan
-- [] 🏴󠁣󠁡󠁹󠁴󠁿 Yukon
- [] 🇨🇺 Cuba
- [] 🇩🇴 Dominican Republic
- [] 🇸🇻 El Salvador
- [] 🇬🇹 Guatemala
- [] 🇭🇹 Haiti
- [] 🇭🇳 Honduras
- [] 🇯🇲 Jamaica
- [] 🇲🇽 Mexico
- [] 🇵🇦 Panama
- [] 🇺🇸 United States: Amtrak #106 
-- [] 🏴󠁵󠁳󠁡󠁬󠁿 Alabama
-- [] 🏴󠁵󠁳󠁡󠁫󠁿 Alaska
-- [] 🇦🇸 American Samoa
-- [] 🏴󠁵󠁳󠁡󠁺󠁿 Arizona
-- [] 🏴󠁵󠁳󠁡󠁲󠁿 Arkansas
-- [] 🏴󠁵󠁳󠁣󠁡󠁿 California: Bay Area #162 
[...]

PR #210, opened 2024-04-05
Added the US State of Indiana to help with issue #201

jjesse:
The State of Indiana needs coverage per issue #201 so adding it here. Thanks for the start of this.


feeds/us-in.json
```diff
@@ -0,0 +1,65 @@
...
+        {
+            "name": "Jonathan Jesse",
+            "github": "jjesse"
+        }
+    ],
+    "sources": [
+        {
+            "name": "Bloomfield Transit",
+            "type": "transitland-atlas",
+            "transitland-atlas-id": "o-dnfq-bloomingtontransit"
```
jbruechert: these need to be feed ids, starting with `f~`. On the transit.land website, you can get from the operator to the feed by clicking on the "Feed" link.

jjesse: THanks will resolve sorry grabbed the wrong ones

PR #222, opened 2024-04-07
us-ca: Add more feeds for California

mlundblad:
. Los Angeles
- Sacramento
- San Diego
- Santa Cruz


hbruch:
In case you encounter any difficulties with the us-ca feeds, please open an issue in http://gituub.com/mfdz/GTFS-issues/issues. When you prefix the issue title with `us-ca`, this would help the people from Caltrans watching the repo.

PR #227, opened 2024-04-09
us-ca: Re-try previously added feeds with proxy

mlundblad:


mlundblad:
Santa Cruz now gives a 404 with proxy, and Sacramento fails to parse (also with fix: true), so still skip those.

PR #237, opened 2024-04-11
ansible: Update MOTIS to v0.11.30

kalon33:


felixguendling:
To enable the third style from the [release screenshots](https://github.com/motis-project/motis/releases/tag/v0.11.30) you would also need to change the tiles import profile to extract the information used in the more colorful style:

```ini
[tiles]
profile=/opt/motis/tiles-profiles/profile.lua
```

https://github.com/public-transport/transitous/blob/698858fe4649a267d805e7c799eb0baaac07891c/motis/config.ini.j2#L46C9-L46C49

The colorful style (`?style=color`) also works with `background.lua` import profile (see second screenshot) and the black&white style also works with the colorful `profile.lua` import profile (third screenshot). But to have all colors, you need to use `profile.lua` together with `?style=color`. I guess it's a matter of taste and where you want to use the tiles.

Indoor controls can be used here to select the displayed level (MOTIS web ui can only display level 0 atm):
https://osr.motis-project.de/

PR #255, opened 2024-04-22
feat(feeds): add public transport of valparaiso chile

CleoMenezesJr:


feeds/cl.json
```diff
@@ -0,0 +1,35 @@
+{
+  "maintainers": [
+    {
+      "name": "Cleo Menezes Jr.",
+      "github": "CleoMenezesJr"
+    }
+  ],
+  "resources": [
```
jbruechert:
```suggestion
  "sources": [
```

feeds/cl.json
```diff
@@ -0,0 +1,35 @@
...
+    {
+      "name": "buses-del-gran-valparaiso",
+      "type": "transitland-atlas",
+      "transitland-atlas-id": "o-66jh-busesdelgranvalparaisosa"
```
jbruechert: These need to start with f-, you can get from the operator to the corresponding feed in the transit.land web interface.

CleoMenezesJr: Sorry, but the onestop ID is the same for all operators:
https://www.transit.land/feeds/f-66jh-vinabussa~trolebuseschilesa~busesdelgranvalparaisosa~topt
Should I leave just one source?

jbruechert: That's actually good, then there is already one feed for all operators. Adding that will be enough then,.

CleoMenezesJr:
> Unfortunately it looks like this feed is fairly old and in the category "historic" on transit land, which our scripts so far can't use.
> 
> There is probably a current version of the dataset on the open data portal, but that seems to be down right now (or at least not accessible for me)

Is there nothing that can be done now? :/

PR #259, opened 2024-04-25
Add feed for Brittany (train, bus, metro, car, boat)

Gildas-GH:
I added a feed for Brittany public transport. It aggregates the feeds of each city in Brittany in a single one. There are some errors (12 stops and 828 stop times, most of them for train stop times, which is already included in SNCF TER feed) so I added the "fix" attribute.


mlundblad:
Thanks!

PR #268, opened 2024-05-03
Add Ontario feeds

espidev:


feeds/ca-on.json
```diff
@@ -7,145 +7,349 @@
...
+        },
+        {
+            "name": "Guelph-Transit",
+            "type": "transitland-atlas",
+            "transitland-atlas-id": "f-guelph~rt"
+        },
+        {
+            "name": "Hamilton-Street-Railway",
+            "type": "transitland-atlas",
+            "transitland-atlas-id": "f-dpxj-hamiltonstreetrailway"
```
jbruechert: The host of this feed seems to be consistently down right now.
I recommend leaving it out for now, and opening an issue to have a look later.
Optionally you can of course try to notify them.

espidev: I removed it for now

PR #325, opened 2024-05-19
max_footpath_length=20, link_stop_distance=100

felixguendling:
20min maximum footpath length
100m stop distance to link stops


felixguendling:
I confused both numbers last time.
This makes more sense. Otherwise I still had some 45min footpaths through Paris.
Should also speedup the preprocessing a bit.

PR #346, opened 2024-05-30
Add V-BUSZ and MKV to Hungary

thegrasshopper104:
Adds V-BUSZ Kft. (Veszprém) and MKV Zrt. (Miskolc) data in Hungary.
Both feeds are directly from the operators.

### DIFF
--- a/feeds/hu.json
+++ b/feeds/hu.json
@@ -25,6 +25,16 @@
             "name": "bkk",
             "type": "transitland-atlas",
             "transitland-atlas-id": "f-u2m-bkk"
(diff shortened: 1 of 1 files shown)


mlundblad:
Btw, is there some license information associated with those feeds?
In that case, I think it could be added to the metadata.

thegrasshopper104:
I couldn't find any specific license or usage restrictions.

Marcus Lundblad ***@***.***> ezt írta (időpont: 2024. máj.
30., Cs 18:35):

PR #355, opened 2024-06-05
feeds: Update NZ.json 

Hoverth:
Updated with baybus, realtime GTFS for Metro Christchurch, and Waikato's BUSIT feeds from the update to transitland -thanks Angus!


Hoverth:
Removed the realtime Metro Christchurch feed, didn't realise it needed the base feed to work, as the base feed needs an API key on the transitland end AFAIK.

PR #358, opened 2024-06-06
Added feed for Indonesia

sandervankasteel:
Currently only Jakarta, Bekasi and Tangerang


sandervankasteel:
> This could be an issue in the CI setup, I'll have a look later

I just had a check locally as well, it seems like the host is temporarily unavailable.

image

PR #364, opened 2024-06-10
Add us-ny region and feeds

kd2flz:


feeds/us-ny.json
```diff
@@ -36,6 +36,46 @@
       "type": "transitland-atlas",
       "transitland-atlas-id": "f-dr5r-nyctsubway"
     },
+    {
+      "name": "MTA_NYCSubway",
```
mlundblad: The import script fails on these names.
I think in other cases we use - as separators in feed IDs.

If I remember correctly, it is because an _ is used to separate the region with the per-region feed ID to create the downloaded final file name.

jbruechert: _ has some internal meaning to MOTIS, iirc to separate feed and station id.

kd2flz: Should I just try replacing the underscore with a dash?

mlundblad: Yeah, that should work!

PR #382, opened 2024-06-24
Update MOTIS to v0.12.8.

kalon33:
Update MOTIS to latest v0.12.8 version, with enhancements to bike transport, shortest footpaths and bugfixes in webUI notably.

### DIFF
--- a/ansible/roles/motis/tasks/main.yml
+++ b/ansible/roles/motis/tasks/main.yml
@@ -8,7 +8,7 @@
 
 - name: Unpack MOTIS
   unarchive:
-    src: https://github.com/motis-project/motis/releases/download/v0.12.5/motis-linux-amd64.tar.bz2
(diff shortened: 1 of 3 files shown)


felixguendling:
I guess it would make sense to add these two attributes to the transitous proxy:

https://github.com/motis-project/motis/blob/9eace97dd51c1093d6894b7ffba40f28ceba350d/protocol/intermodal/IntermodalRoutingRequest.fbs#L133-L134

Regarding bike transport: unfortunately, the data situation is not very good here. We will add the option to provide a default setting (allowed /  not allowed) for each clasz (or route_type - not decided yet) for each dataset which will be used in case a trip does not state whether bikes are allowed or not allowed.

felixguendling:
If I remember correctly, there are no `osr` changes that change the on-disk data format.

PR #385, opened 2024-06-26
Added some autobus companies in Lombardy

airon90:
Added Movibus, AirPullman, NET, STAR and STAV from https://www.agenziatpl.it/open-data/gtfs

I suppose those data must be checked regularly as the filename contains a date and they may change the date when updating the data.

### DIFF
--- a/feeds/it.json
+++ b/feeds/it.json
@@ -49,7 +49,7 @@
         {
             "name": "Lombardia-AirPullman",
             "type": "http",
-            "fix": true
+            "fix": true,
             "license": {
                 "spdx-identifier": "CC-BY-4.0"
             },
--- a/feeds/it.json
+++ b/feeds/it.json
@@ -47,10 +47,50 @@
             "fix": true
         },
         {
+            "name": "Lombardia-AirPullman",
+            "type": "http",
+            "license": {
+                "spdx-identifier": "CC-BY-4.0"
+            },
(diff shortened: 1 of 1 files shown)


mlundblad:
Yeah, it looks like those URLs will probably change…
I wonder if we should have some way to insert variables in the URL, like current date (but that might not catch the correct file).
Or maybe if it would be possible to "probe" the directory to list the file, and somehow take the newest one…

feeds/it.json
```diff
@@ -46,12 +46,52 @@
             "url": "https://www.amt.genova.it/amt/GTFS/GTFS_AMT_GENOVA.zip",
             "fix": true
         },
+        {
+            "name": "Lombardia-AirPullman",
+            "type": "http",
```
jbruechert, line 51:
```suggestion
            "type": "http",
            "fix": true
```

feeds/it.json
```diff
@@ -46,12 +46,53 @@
...
+            "fix": true
```
jbruechert:
```suggestion
            "fix": true,
```

mlundblad:
> I think we can merge it as it is now, but it would be nice to have a script to update the links. Something similar to what we have for france (`src/generate-france.py`). In this case it would be web scraping based (for example using BeautifulSoup).

Yeah, sounds good!

PR #414, opened 2024-07-13
Merging to fix issue #201

jjesse:
Sadly Kansas only has 1 feed that is valid and working will check back on more


mlundblad:
BTW, shouldn't this have been put under Arkansas (I think the feed is for Little Rock)?

jjesse:
Oh shoot  let mw double check and fix of needed but will be late tonight

PR #427, opened 2024-07-22
added Poznan region and wkulesza as maintainer

wkulesza:
Added Poznan GTFS as region. ZTM Poznan (Municipal Transport Authority in Poznan) is the original maintainer of the feed. Source is here: https://ztm.poznan.pl/pl/dla-deweloperow/gtfsFiles but also here in Transitland - https://www.transit.land/feeds/f-u3k4-miejskieprzedsi%C4%99biorstwokomunikacyjnespzoowpoznaniu~k%C3%B3rni so I added transitland-atlas-id instead of original url.

### DIFF
--- a/feeds/pl.json
+++ b/feeds/pl.json
@@ -4,6 +4,10 @@
             "name": "Jonah Brüchert",
             "github": "jbruechert"
         }
+        {
+            "name": "Wojciech Kulesza"
+            "github": "wkulesza"
+        }
     ],
     "sources": [
         {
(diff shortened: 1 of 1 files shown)


feeds/pl.json
```diff
@@ -112,5 +116,10 @@
             "type": "transitland-atlas",
             "transitland-atlas-id": "f-u3tu-elblag"
         }
```
jbruechert:
```suggestion
        },
```

feeds/pl.json
```diff
@@ -110,7 +114,12 @@
         {
             "name": "Elbląg",
             "type": "transitland-atlas",
-            "transitland-atlas-id": "f-u3tu-elblag"
+            "transitland-atlas-id": "f-u3tu-elblag",
```
jbruechert:
```suggestion
            "transitland-atlas-id": "f-u3tu-elblag"
```

wkulesza: Should i add or remove the semicolon ?

jbruechert: This one needs to be removed. There can't be a semicolon after the last attribute in an object (`{ ... }`)

PR #431, opened 2024-07-29
Add Minneapolis/St. Paul GTFS feed

vkrause:


mlundblad:
Hmm, these are in Minnesota, so I think us-mn.json would be more approriate.

vkrause:
> Hmm, these are in Minnesota, so I think us-mn.json would be more approriate.

oops, indeed, need to improve my my ISO 3166-2 code knowledge for the US a bit :D will fix

mlundblad:
Thanks!

Issue #437, opened 2024-08-07
Germany DELFI GTFS-RT is available!


traines-source:
The DELFI GTFS-RT is finally available! https://mobilithek.info/offers/755009281410899968

People say that the data quality is actually quite good (haven't investigated it myself), but it is still very incomplete. I think it can't harm to add it. However, it needs registration as an organisation and a client certificate must be sent on each request. 

I've successfully registered and am mirroring publicly at https://stc.traines.eu/mirror/german-delfi-gtfs-rt/latest.gtfs-rt.pbf (fetching every two minutes).
What do you think, shall I make a PR to use my mirror (at least provisionally, since apparently towards the end of the year this is supposed to become more open), or do we want to fetch it directly from mobilithek and add support for client certificate authentication (I suppose this is not supported yet). In this case I could also offer to issue a certificate via my mobilithek account.

PR #438, opened 2024-08-07
Add DELFI GTFS-RT

traines-source:
As per #437, adding my mirror of the new DELFI GTFS-RT feed. I haven't tested it locally, so let's see.

### DIFF
--- a/feeds/de.json
+++ b/feeds/de.json
@@ -3,6 +3,10 @@
         {
             "name": "Jonah Brüchert",
             "github": "jbruechert"
+        },
+        {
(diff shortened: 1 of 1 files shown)


traines-source:
Thanks for testing it! Yes allegedly coverage is going to improve over the coming months. Weirdly, the size of the feed varies a lot during the day (from 100KB to over 3MB within a few minutes). And right now, I'm getting a lot of Gateway Timeouts from Mobilithek... As far as I can tell, both are not problems on my side, so the system seems to be a bit brittle still...

PR #455, opened 2024-08-17
Add USA/Indiana/Indianapolis IndyGo realtime feed information

jscott0:
This is my first time contributing to Transitous and am having trouble navigating how to test the changes locally. Healthy skepticism is justified, but since the data is fetched from Transitland Atlas I'm hopeful this makes it more sound.


mlundblad:
LGTM!

Thanks!

PR #464, opened 2024-08-27
Add GNOME Maps to list of supporting apps

mlundblad:
Add GNOME Maps to the list of supporting apps.


website/content/_index.html
```diff
@@ -31,8 +31,9 @@ <h2>Apps That Use Transitous</h2>
 
 <div class="mt-4 app-card-row">
 {{< app-card name="KDE Itinerary" website="https://apps.kde.org/org.kde.itinerary" icon="images/apps/kde-itinerary.svg" >}}
-{{< app-card name="KTrip" website="https://apps.kde.org/org.kde.ktrip" icon="images/apps/ktrip.svg" >}}
+{pp-card name="KTrip" website="https://apps.kde.org/org.kde.ktrip" icon="images/apps/ktrip.svg" >}}
```
jbruechert: Looks like an accidental change

mlundblad: Yeah

I just fixed that now (I used the web interface, as I'm using my work laptop now, and didn't go through checking out the branch and everything, so it's not squashed, don't know if that's possible to do from the web).

PR #466, opened 2024-08-28
Add GTFS feed for several cities from Turkey

yasinkaraaslan:


mlundblad:
Some of these feeds fails to parse.
Maybe adding "fix": true to those failing might help (that would use the -fix option when running gtfstidy when importing).

yasinkaraaslan:
> Thanks!
> 
> I added the fix option for you on a number of feeds.

Thanks but it removes a lot of information on some of them (on izmir metro it's almost 95%), so it isn't really a fix. 
I'll try uploading refactored versions of the feeds to my website and if it works I might submit another pull request afterward

PR #469, opened 2024-08-28
Fixing broken Turkey feeds

yasinkaraaslan:
It's so stupid that we have to put up with this crap, I finally got something more or less working without too many dropped infos.
It's crazy that they probably post the data without even testing.

Also, is there a chance to remove --drop-too-fast-trips from the test script? 13% of trips from istanbul-iett is dropped because of that. It's probably again someone put some numbers wrong but I am so tired and don't want to fix 17000 different numbers by hand :D

### DIFF
--- a/feeds/tr.json
+++ b/feeds/tr.json
@@ -14,12 +14,14 @@
         {
             "name": "izmir-metro",
             "type": "transitland-atlas",
-            "transitland-atlas-id": "f-izmir~metro"
+            "transitland-atlas-id": "f-izmir~metro",
+            "fix": true
         },
         {
             "name": "izmir-izban",
             "type": "transitland-atlas",
-            "transitland-atlas-id": "f-izmir~rail"
+            "transitland-atlas-id": "f-izmir~rail",
+            "fix": true
         },
         {
             "name": "izmir-tram",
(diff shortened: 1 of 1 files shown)


yasinkaraaslan:
> btw, did you automate the floating point fixes or do you do them by hand? If so you'd probably have to do that again every year, not sure if that works so well.
> 
> Another option is to write the feed authors, maybe they are responsive and can fix this properly.

There was like 20 of them, so didn't bother writing any scripts. What took my time was the istanbul feeds, I had to correct errors one by one because they were pretty random. 
I guess this has to do it for now.

> I think the trips are most likely dropped because the coordinates are wrong, and that causes them to be much further apart and the trip becoming too fast. If these trips are not filtered, they create connections in the routing from completely wrong places and can potentially break all useful connections (because they are much faster and therefore preferred)

Supposedly they share the same feeds with google, I wonder how they get around these.

yasinkaraaslan:
oh no

yasinkaraaslan:
So sorry about that, it's the almighty cloudflare challenge. I think I failed to notice that because my ip was whitelisted. I disabled some stuff for /gtfs/* path. Can you rerun the checks again?

yasinkaraaslan:
Again, I am sorry about that. I would guess it is somehow possible to whitelist a path but couldn't figure out how

PR #472, opened 2024-08-29
Add North Carolina RTP region transit operators

deanapearce:


feeds/us-nc.json
```diff
@@ -0,0 +1,70 @@
...
+				"type": "transitland-atlas",
+				"transitland-atlas-id": "f-dnrg-cary~transit~nc~us"
+			},
+			{
+				"name": "GoCary",
+				"type": "transitland-atlas",
+				"transitland-atlas-id": "f-dnrg-cary~transit~nc~us~rt"
+			},
+			{
+				"name": "Chapel Hill Transit",
```
mlundblad: The names can not contain spaces (and if I remember correctly, underscores are also not possible, due to the output filename being <base file name>_<feed name>

So, replacing these spaces with "-" should work

mlundblad:
Thanks!

PR #476, opened 2024-08-31
Add initial Israel sources

yotamN:
I've sent a message to Busofash regarding their error in the GTFS file.

### DIFF
--- a/feeds/il.json
+++ b/feeds/il.json
@@ -9,7 +9,7 @@
         {
             "name": "busofash",
             "type": "http",
-            "url": "https://opendatasource.tel-aviv.gov.il/opendata_ducaments/gtfs.zip"
+            "url": "https://transitous.yotam.net/busofash.zip"
         },
         {
             "name": "busofash",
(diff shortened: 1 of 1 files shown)


vkrause:
The second one passes locally with the following change and #477:
```
        {
            "name": "Israel-MOT",
            "type": "transitland-atlas",
            "transitland-atlas-id": "f-sv-גלים~מועצהאזוריתגולן~אפיקים~דןצפון~דןבדרום~דןבארשבע~ירושלים",
            "http-options": {
                "ignore-tls-errors": true,
                "headers": {
                    "user-agent": "Mozilla/5.0 (X11; Linux x86_64; rv:128.0) Gecko/20100101 Firefox/128.0"
                }
            },
            "fix": true
        }
```

yotamN:
Will try your configuration but it seems that Transitland was last updated 5 months ago, not so up-to-date.

yotamN:
Still not sure why Israel-MOT is closing the connection so I skipped it for now

feeds/il.json
```diff
@@ -0,0 +1,50 @@
...
+                }
+            }
+        },
+        {
+            "name": "Israel-MOT",
+            "type": "http",
+            "url": "https://gtfs.mot.gov.il/gtfsfiles/israel-public-transportation.zip",
+            "http-options": {
+                "headers": {
+                    "user-agent": "Mozilla/5.0 (X11; Linux x86_64; rv:128.0) Gecko/20100101 Firefox/128.0"
```
mlundblad: Although the HTTP RFCs state that the header field names should be case-insensitive, is this always true in practice? I think I've mostly seen "User-Agent" being used.
Maybe it could be worth trying to changing to that?

vkrause: We might need to make https://github.com/public-transport/transitous/blob/main/src/fetch.py#L60 case-insensitive as well then.

jbruechert:
Thanks!

PR #480, opened 2024-09-02
Fix Cap Cotentin network URL

kalon33:
URL for Cap Cotentin was an old one. Fix it.

BTW @jbruechert do you know why most (except one) data has been changed to URL format when they where updated and working in Transitland Atlas format?

### DIFF
--- a/ansible/roles/motis/tasks/main.yml
+++ b/ansible/roles/motis/tasks/main.yml
@@ -8,7 +8,7 @@
 
 - name: Unpack MOTIS
   unarchive:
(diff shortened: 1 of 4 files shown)


kalon33:
Oups, seems that MOTIS update went with this PR. Feel free to discard it if you don't want it.

kalon33:
@jbruechert Thanks for your answer. Is the regeneration an automated process? Multiple feeds in fr.json are currently broken, expired, or pointing to old URLs.

kalon33:
Any idea with the script doesn't use the static URLs instead of the changing ones?

PR #481, opened 2024-09-02
generate-france: Use stable urls

jbruechert:


kalon33:
It seems to be working, only 4 errors in the test run (missing agency.txt, all from the same feed producer)

kalon33:
> Unfortunately these feeds are fairly important, but return something like
> 
> ```
> donnees;format;download
> Horaires des trains Transilien NeTEx;NeTEx;https://eu.ftp.opendatasoft.com/sncf/gtfs/transilien-netexfr.zip
> Horaires des trains Transilien GTFS;GTFS;https://eu.ftp.opendatasoft.com/sncf/gtfs/transilien-gtfs.zip
> ```
> 
> Maybe the logic for picking the right resource needs some improvement.
> 
> **EDIT** we are already picking the right resource, but "url" and "original_url" don't return the same.

Yep indeed, SNCF feeds are relevant for most trains in France...

I still don't see some data in transitous (Cap Cotentin related data notably) do you know why?

laem:
Hi, feel free to read this very similar (deno) script https://github.com/laem/gtfs/blob/master/buildConfig.ts

I had the same problem with some URLs returning index files instead of GTFS files...

This comment : https://github.com/laem/gtfs/blob/master/buildConfig.ts#L157

jbruechert:
Thanks, it would be best if it was resolved upstream. Until then, regenerating the file often enough helps partially.

jbruechert:
> Is this `Horaires des trains Transilien` the only problematic resource?

No, I think this happened for every resource with multiple files.

It might be worth talking to the developers of the open data portal if you have the energy / time to do it.
Handling of multiple files per resource is problematic in different ways (no automated way to match gtfs-rt to gtfs). Maybe they could be stricter about what they allow to be a single resource.

PR #493, opened 2024-09-06
Disable Istanbul feeds, add Kocaeli feed

yasinkaraaslan:
The istanbul-iett feed is broken, disabled all the Istanbul feeds for now for now. 

A Kocaeli feed has been added.


feeds/tr.json
```diff
@@ -31,18 +31,26 @@
...
+            "fix": true,
+            "skip": true
         },
         {
             "name": "istanbul-iett",
             "type": "http",
             "url": "https://dl.dropboxusercontent.com/scl/fi/t57vx1ivwq02x0xy51tum/istanbul-iett.zip?rlkey=qb0ien7ba5s2okt3or2lzc2nn",
-            "drop-too-fast-trips": false
+            "drop-too-fast-trips": false,
+            "skip": true
```
jbruechert:
```suggestion
            "skip": true,
            "skip-reason": "Most trips are not usable, causes routing to return nonsensical results."
```

PR #495, opened 2024-09-11
Update to MOTIS v0.12.13.

kalon33:
It has duplicate merging parameters that should (to my understanding) fix real time use when GTFS feeds overlaps.
   
We could add use to the configuration once this PR would be merged.


felixguendling:
before: `match_duplicates` matches duplicates from different GTFS sources (but not within the same)

https://github.com/public-transport/transitous/blob/66ebf9d73991ea5f868d1c0b6ecf1f47f2d4760a/motis/config.ini.j2#L36


now:

- `merge_dupes_intra_src` = deduplicate within the same GTFS source (not possible before)
- `merge_dupes_inter_src` = same as `match_duplicates` (deduplicate trips from different GTFS sources)

Additionally, we have Open Telemetry (OTEL) support ([MOTIS issue 541](https://github.com/motis-project/motis/pull/541)) and have a metrics endpoint for Prometheus, so it's possible now to create Grafana dashboards to monitor matching quality of real-time feeds to GTFS static sources, plot query runtimes, etc.

Be aware that if you use OTEL, routing queries will be logged (which enables us to identify problematic queries, but might not be what you want regarding privacy).

kalon33:
> before: `match_duplicates` matches duplicates from different GTFS sources (but not within the same)
> 
> https://github.com/public-transport/transitous/blob/66ebf9d73991ea5f868d1c0b6ecf1f47f2d4760a/motis/config.ini.j2#L36
> 
> now:
> 
>     * `merge_dupes_intra_src` = deduplicate within the same GTFS source (not possible before)
> 
>     * `merge_dupes_inter_src` = same as `match_duplicates` (deduplicate trips from different GTFS sources)
> 
> 
> Additionally, we have Open Telemetry (OTEL) support ([MOTIS issue 541](https://github.com/motis-project/motis/pull/541)) and have a metrics endpoint for Prometheus, so it's possible now to create Grafana dashboards to monitor matching quality of real-time feeds to GTFS static sources, plot query runtimes, etc.
> 
> Be aware that if you use OTEL, routing queries will be logged (which enables us to identify problematic queries, but might not be what you want regarding privacy).
> 
> _Edit_: To really make use of `merge_dupes_intra_src` you would probably need to turn of fixing feeds with `gtfs-tidy`. This makes sense especially for GTFS static feeds that also have one or more real-time feeds. With `gtfs-tidy` trips are removed completely while MOTIS will only deactivate one trip and point its `trip_id` to the remaining trip. With `gtfs-tidy` you never know if the trip that will get real-time updates will be removed.

Thanks for the details @felixguendling .

felixguendling:
If MOTIS doesn't handle something like missing files or other stuff (which was handled by `gtfs-tidy`, we can also make it more robust. For example the case of missing agencies is already handled (it creates an "UNKNOWN" agency) - so we could make the agency file optional (as an example).

kalon33:
Activating `merge_dupes_inter_src` would be nice as we have multiple feeds that are overlapping in the ones we use, and that sometimes prevent GTFS-RT data to be properly used.

felixguendling:
It doesn't seem to be activated in the configuration file:

https://github.com/public-transport/transitous/blob/5eff476e5ffd70b15cdbcd89da228559ac13d364/motis/config.ini.j2#L33-L42

The default is that it's turned off:
https://github.com/motis-project/motis/blob/0ba35bedb7f33db4bf250bdc417af94f99cc19f2/modules/nigiri/include/motis/nigiri/nigiri.h#L35-L36

felixguendling:
I made the change here: #507

I haven't had time to deploy it to our test instance on europe.motis-project.de - so if you don't want any surprises I can give you green light after I tested it. If you're adventurous, you can just merge it and see what happens. I hope it should be fine.

PR #518, opened 2024-09-19
Initial feeds for Pennsylvania

mlundblad:
Add SEPTA (Southeastern Pennsylvania Transportation Agency) for Philadelphia area.

### DIFF
--- a/feeds/us-pa.json
+++ b/feeds/us-pa.json
@@ -1,0 +1,30 @@
+{
+  "maintainers": [
+    {
+      "name": "Marcus Lundblad",
+      "github": "mlundblad"
+    }
+  ],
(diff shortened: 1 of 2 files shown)


mlundblad:
Reported an issue in GTFS-Issues repo: https://github.com/mfdz/GTFS-Issues/issues/154

It seems the download links use a form like gtfs_file.zip#<sub file> leading to the download of the "main" file (contain two archives). Maybe the intent is that the client should interpret #<sub file> as "extract <sub file> from ZIP and use that"

mlundblad:
Yeah, maybe I'll look into it. I guess it could be tacked on here, as this would already be a good test bed for it.

PR #524, opened 2024-09-24
Include rt sources in credits list

traines-source:
I think we should also include the sources/licenses of the (GTFS-)RT feeds, since occasionally the license differs from the static feed (e.g. for Germany).

Surely one could come up with something more sophisticated that only displays this information if it is different from the static feed; I opted to just add a sentence below the `Original file` and `Processed file` buttons of the respective static feed. But I'm open for suggestions :)

(Also, this lead to some refactoring, which is why the diff is so large...)

### DIFF
--- a/src/generate-attribution.py
+++ b/src/generate-attribution.py
@@ -23,15 +23,31 @@ if __name__ == "__main__":
 
     transitland_atlas = transitland.Atlas.load(Path("transitland-atlas/"))
 
-    attributions = []
+    attributions = {}
 
     for feed in sorted(feed_dir.glob("*.json")):
         parsed = {}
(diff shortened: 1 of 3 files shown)


jbruechert:
Thanks, makes sense!

Please commit the updated license.json file in the website subdirectory so the website can be built locally without actually downloading all the feeds.
On the ci the file will be overwritten with current data.

traines-source:
Ahem I don't want to download all the feeds either :) But I added some example RT references to the existing `license.json`, which is backwards compatible anyways.

(For testing without having the feeds, I commented out lines 88 to 97 from `generate-attribution.py`).

PR #536, opened 2024-09-30
Izmir eshot fix

yasinkaraaslan:
There was a problem with fetching this feed so I uploaded it to dropbox.


jbruechert:
The issue was probably temporary. Mirroring it manually is a lot more work long term, I'd just wait if it works again on the next run

mlundblad:
Oops, I just merged it…
Yeah, maybe we could revert it, and wait and see if the issue persists?

jbruechert:
Yeah, would prefer to revert it for now.

PR #559, opened 2024-10-17
Add TEC Rt Feed

### LINKED ISSUE #558: Add RT Feed for BE-tec
Hello 👋 
Thanks for maintaining this, 
it seems that transitland have just added the RT version of tec

https://github.com/transitland/transitland-atlas/pull/1263

new id is "f-u0g-tec~rt" ... not 100% sure how to add it, but i would be happy to help if i can

eMerzh:
fix #558

not 100% sure it's good, but here is the attempt

### DIFF
--- a/feeds/be.json
+++ b/feeds/be.json
@@ -40,6 +40,12 @@
             "type": "transitland-atlas",
             "transitland-atlas-id": "f-u0g-tec",
             "fix": true
+        },
+        {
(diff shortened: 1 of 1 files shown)


jbruechert:
`fix` does not have any effect on realtime feeds, so please remove it.

Apart from that, looks good!

jbruechert:
Thanks!

Issue #114, opened 2024-03-09
European coverage status


vkrause:
Tracking issue for coverage in Europe

* [ ] :albania: Albania: unknown
* [ ] :andorra: Andorra: unknown
* [ ] :austria: Austria: national railway included, local transport available with contract?
* [x] :belgium: Belgium
* [ ] :bosnia_herzegovina: Bosnia and Herzegovina: unknown
* [ ] :bulgaria: Bulgaria: unknown
* [ ] :croatia: Croatia: national railway included (#109), local transport unknown
* [ ] :cyprus: Cyprus: unknown
* [ ] :czech_republic: Czechia: unknown
* [x] :denmark: Denmark 
* [x] :estonia: Estonia
* [ ] :finland: Finland: data available
* [ ] :fr: France: national railways and individual local ones done, many local ones still to be added
* [x] :de: Germany
* [ ] :greece: Greece: unknown
* [ ] :hungary: Hungary: unknown
* [ ] :iceland: Iceland: data available
* [x] :ireland: Ireland 
* [ ] :it: Italy: more feeds to be added? #76 #77
* [x] :latvia: Latvia
* [ ] :liechtenstein: Liechtenstein: unknown
* [x] :lithuania: Lithuania
* [x] :luxembourg: Luxembourg
* [ ] :malta: Malta: unknown
* [ ] :moldova: Moldova: unknown
* [ ] :monaco: Monaco: unknown
* [ ] :montenegro: Montenegro: unknown
* [x] :netherlands: Netherlands
* [ ] :macedonia: Macedonia: unknwon
* [ ] :poland: Poland: national railway included, partial local coverage?
* [ ] :portugal: Portugal: unknown
* [ ] :romania: Romania: unknown
* [ ] :san_marino: San Marino: unknown
* [ ] :serbia: Serbia: unknown
* [ ] :slovakia: Slovakia: #108, local transport unknown
* [ ] :slovenia: Slovenia: unknown
* [ ] :es: Spain: national railway included, partial local coverage?
* [x] :sweden: Sweden
* [x] :switzerland: Switzerland
* [ ] :gb: United Kingdom: Great Britain done, Northern Ireland todo

PartTimeDataScientist:
Nice overview! 😍

Would it be an idea to ask the maintainers of the feeds to add high level area IDs e.g. the stable identifiers from https://www.whosonfirst.org/?

Doing so would allow us to auto-generate a coverage map (separated by continent, country static, real-rime,...)

[...]

PR #560, opened 2024-10-17
Add Cotral (Lazio, IT)

Altonss:
Improve Italian coverage #114.
Important feeds in Italy are still missing (Frecciarossa, Italo, Itabus for example).


jbruechert:
Thanks!

PR #578, opened 2024-11-01
docs: add common patterns section[...]

Hoverth:
[...]with an example for using a api key with a transitland feed


jbruechert:
Thanks!

PR #606, opened 2024-11-08
Add Malaysia

Altonss:


jbruechert:
Probably they don't provide trip updates (pre-calculated delay) but only the raw vehicle positions.
I hope that we'll be able to work with them at some point. I think it's okay to just leave it this way until then.

PR #609, opened 2024-11-09
feat(feeds): NZ.json - added BUSIT and Baybus

Hoverth:
Thanks to whoever added Baybus to Transitland!

This should be the last of the larger cities, all the others don't have publicly available GTFS feeds as far as I can tell, so some emailing will be in order for the future.


jbruechert:
Thanks!

PR #612, opened 2024-11-10
Improve transitous bot

Altonss:
I haven't tested this on the real transitous repo yet, but this should maybe improve the raised issues by linking to the feed file directly.
However, with the current setup the commit hash doesn't get updated if the issue gets updated afaik.


jbruechert:
Looks good, I didn't review it so far because it was still in draft status

Issue #79, opened 2024-03-02
Add feeds for major metropolitan US cities (Los Angeles, San Francisco, New York City, Boston, Washington, etc.)


nekohayo:
Those are very populous cities, and also tend to be tourist attractions, so it might make sense to add them first? I suspect these cities all have open transit data. Looking at OpenMobilityData [transit feeds for the USA](https://transitfeeds.com/l/31-united-states), I can see:

* [New York](https://transitfeeds.com/search?q=new+york)
* [Los Angeles](https://transitfeeds.com/l/154-los-angeles-ca-usa) and surrounding suburbs and counties (as seen in the map and sidebar on the right) - the Los Angeles metropolitan area seems to be a big pile of counties in a trenchcoat, each with their own mini transit system
* [Boston](https://transitfeeds.com/search?q=boston) and [Massachusetts](https://transitfeeds.com/search?q=massachusetts)
* [San Francisco](https://transitfeeds.com/search?q=san+francisco) and the [Bay Area Transit Authority](https://transitfeeds.com/p/bay-area-transit-authority)
* [Seattle](https://transitfeeds.com/search?q=seattle)
* others?

sjwhitak:
I've been working with New England's feeds since I was planning to make an app like this for the New England region before I found this project. The data I found was:
* Connecticut's CTtransit GTFS files [here](https://www.cttransit.com/about/developers) (multiple static GTFS links in that URL) and [live updates](https://cttprdtmgtfs.ctttrpcloud.com/TMGTFSRealTimeWebService/Vehicle/VehiclePositions.pb)
* Rhode Island'S RIPTA GTFS [static](https://ripta.com/RIPTA-GTFS.zip) and [live updates](http://realtime.ripta.com:81/), 
* Boston's [MBTA](https://www.mbta.com/developers/gtfs) GTFS files, [static](https://cdn.mbta.com/MBTA_GTFS.zip) and [live updates](https://cdn.mbta.com/realtime/VehiclePositions.pb). 

I found when testing with these live feeds, the updates occur MBTA: 5 sec, RIPTA: 20 sec, CTtransit: 5 sec. CTtransit seems to have removed this document, but they used to have a rate limit of 30 seconds, documented [here](https://www.cttransit.com/sites/default/files/PDF_files/CTtransit_DeveloperDocumentation_Real-Time_v0.9.2_PreLaunch.pdf) but they may have scraped that requirement. I wasn't getting rate limited with 3 second updates. 

[...]

PR #640, opened 2024-11-13
Improve Pennsylvania [US]

Altonss:
Improves #201 and #79 
I'm merging on main, because the feed limit seems to have been raised.


jbruechert:
Thanks!

PR #647, opened 2024-11-19
Add more LA County bus and regional rail operators

DefenestratingWizard:
Added entries for:
- Burbank Bus
- ~~Glendale Beeline~~
- Santa Clarita Transit
- AVTA
- Metrolink

this is my first PR so I may have made a mistake

### DIFF
--- a/feeds/us-ca.json
+++ b/feeds/us-ca.json
@@ -96,9 +96,17 @@
             "transitland-atlas-id": "f-antelope~valley~transit~authority"
         },
         {
+            "name": "Glendale-Beeline",
+            "type": "transitland-atlas",
+            "transitland-atlas-id": "f-9q5f-glendalebeeline",
+            "skip": true,
+            "skip-reason": "403 error code when downloading, no documentation from agency on fix"
+        },
+        {
             "name": "Burbank-Bus",
             "type": "transitland-atlas",
-            "transitland-atlas-id": "f-9q5f7-burbankbus"
+            "transitland-atlas-id": "f-9q5f7-burbankbus",
+            "fix": true
         },
         {
             "name": "Santa-Clarita-Transit",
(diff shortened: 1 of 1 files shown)


feeds/us-ca.json
```diff
@@ -90,6 +90,30 @@
...
         },
+        {
+            "name": "Antelope-Valley-Transit-Authority",
+            "type": "transitland-atlas",
+            "transitland-atlas-id": "f-antelope~valley~transit~authority"
+        },
+        {
+            "name": "Glendale-Beeline",
+            "type": "transitland-atlas",
+            "transitland-atlas-id": "f-9q5f-glendalebeeline"
```
mlundblad: It looks like this gives a 403 (Permission denied).
Maybe it needs an addional authentication header with an API key, or something like that?

DefenestratingWizard: it seems to allow download when the feed url is opened in firefox. I can't seem to find information about an API key or required headers (or any info other than the transitland entries) so maybe it makes sense to remove Beeline for now.

feeds/us-ca.json
```diff
@@ -90,6 +90,30 @@
...
+        },
+        {
+            "name": "Burbank-Bus",
```
mlundblad, line 106: This one seems to stumble on agency.txt (something with the language code).
Maybe adding:

"fix": true
would resolve this (I'm not entirely sure)

feeds/us-ca.json
jbruechert: The server only seems to answer with the data if the client advertises support for compressed transfers. Apparently we don't do that right now, we should fix that.
If you don't want to do it right now, I suggest adding skip: true with a skip-reason for now so we can add it later.

mlundblad:
Thanks!
Looks good!
(there's a couple of other feeds failing now, but those are unrelated to these)

feeds/us-ca.json
mlundblad: Yeah, it's marked as skipped for now.

PR #650, opened 2024-11-22
Add gtfs-rt feeds, fix skip reason on glendale beeline

DefenestratingWizard:
Added gtfs-rt feeds for burbank bus, avta, and santa clarita transit as they seem to have rt feeds available without api keys

Changed skip reason on Glendale beeline to add text about compression mentioned in an earlier review


mlundblad:
Thanks!

PR #653, opened 2024-11-24
Add new operators to hu

thegrasshopper104:
Adds some major Hungarian cities' operators.
The original source for SZKT is http://szegedimenetrend.hu/google_transit.zip


jbruechert:
Thanks!

PR #699, opened 2024-12-14
France: improve handling of multiple feeds in one dataset

### LINKED ISSUE #692: Issue in generate_france.py, several GTFS in one dataset
After analysing `generate_france.py` and the resulting `fr.json`, I noticed some weird changes at regeneration, especially some feed like `arrets-horaires-et-parcours-theoriques-gtfs-du-reseau-routier-regional-de-transport-scolaire-et-interurbain-59-nord` where I noticed strange changes.
The issue is caused by some datasets containing several GTFS files (for several subregions of the dataset for example). We should better handle this:
- [ ] import all GTFS files for each dataset (filter out possible exceptions to this rule if needed)
- [ ] stable ordering/import such that links don't switch randomly at regeneration if one datasets contains several GTFS

Altonss:
Improve handling of multiple feeds (GTFS, GTFS-RT) in one dataset, this fixes #692 .

### DIFF
--- a/feeds/fr.json
+++ b/feeds/fr.json
@@ -65,7 +65,7 @@
             }
         },
         {
-            "name": "eurostar-gtfs--Eurostar_GTFS_a_partir_du_15/10/2024",
+            "name": "eurostar-gtfs--Eurostar-GTFS-a-partir-du-15-10-2024",
             "type": "http",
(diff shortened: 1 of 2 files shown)


jbruechert:
Thanks!

PR #708, opened 2024-12-16
Fix generate-france.py

Altonss:
A file name changed...


jbruechert:
There is also this issue:
```
Warning: Found duplicate source name: fr_reseau-urbain-la-navette-commune-de-gaillac--gtfs-gaillac.zip
```

jbruechert:
I think using the id is fine

PR #709, opened 2024-12-16
France: improve handling of multiple feeds in one dataset

Altonss:
Second attempt at improving handling of multiple feeds (GTFS, GTFS-RT) in one dataset.

This should be better:
- use id of resource as identifier (instead of title) : no duplicate GTFS names should happen anymore (and should be more stable too)
- added verification before adding a GTFS-RT feed : check if a corresponding GTFS feed having the same name exists, otherwise raise a warning :)


jbruechert:
Thank you!

Issue #748, opened 2025-01-03
Better filtering for SNCF/Eurostar trains


Altonss:
Most TGVs/Ouigo seem to come from sbb.ch, with the new filtering capabilities we should maybe filter this out? :thinking:
Same goes for Eurostar, which seems to be even worse: some trains seem to appear 3 times (2 times under thalys, and 1 time as Eurostar from NS data).

As long as the official data source is working fine, it makes more sense to use it. The only challenge with this filtering is ensuring that no important data is missed.

jbruechert:
For Eurostar the problem is made worse by the separate feeds for Eurostar / Thalys, which at the last time I checked still survived the merger. Maybe they have started to unify it finally?

jbruechert:
Whether it's in eu.json or fr.json is a tangential issue, most importantly we should probably drop thalys from eu.json

jbruechert:
But in general I agree, if it's clear which source is the best one (up to date, realtime, quality) then feel free to add agency removal to the other sources.

jbruechert:
This would basically boil down to improved (fuzzy) merging support in MOTIS, right? That would be great in any case

felixguendling:
If anyone is interested in improving fuzzy merging in nigiri, I am happy to answer questions. We could also setup a meeting if someone wants to work on it.

felixguendling:
Maybe it's sufficient to change the threshold for merging here?
https://github.com/motis-project/nigiri/blob/master/src/loader/merge_duplicates.cc#L141

PR #749, opened 2025-01-03
Drop SNCF from CH feed

Altonss:
Improves #748


jbruechert:
They probably block GitHub Actions / azure, should be fine on the actual import server I hope.

PR #754, opened 2025-01-04
Hide source details for each region by default

Altonss:


jbruechert:
Thanks!

jbruechert:
I think that was fallout from me breaking a type check recently: https://github.com/public-transport/transitous/commit/3d563b784d0fe961b41c854d5a16748a1d282895

We will see whether that was it after the next import + website build.

jbruechert:
Looks good again now

Issue #757, opened 2025-01-05
Switch to new sources for GB


Altonss:
Following the discussion in #758, here an interesting project to generate GTFS data from official sources for rail :
 https://github.com/thomasforth/ATOCCIF2GTFS

jbruechert:
This looks fairly actively maintained: https://github.com/planarnetwork/dtd2mysql

Unfortunately they no longer publish ready made GTFS files due to issues with AWS fees. Someone else would need to regularly run this.

PR #758, opened 2025-01-05
Update gb.json

Altonss:
Add official feed for bus dft, improves #757

### DIFF
--- a/feeds/gb.json
+++ b/feeds/gb.json
@@ -13,6 +13,11 @@
             "fix": true
         },
         {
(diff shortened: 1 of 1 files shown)


jbruechert:
Thanks!

To add some background: so far I wasn't able to find the upstream source of the busmaps feed we have used so far. My guess is that busmaps did not update the data in time, so having access to the upstream source would help a lot.

jbruechert:
The old feed is updated again now at least, but it would still be good to replace it.

PR #779, opened 2025-01-09
Add AVV-Aachen to de.json

dancojocaru2000:
Added [Aachener Verkehrsverbund](https://avv.de)'s opendata GTFS feed.

I decided to use the name AVV-Aachen in case [AVV-Augsburg](https://avv-augsburg.de) will be added eventually as well.

### DIFF
--- a/feeds/de.json
+++ b/feeds/de.json
@@ -7,6 +7,10 @@
         {
             "name": "Robin Durner",
             "github": "traines-source"
+        },
+        {
+            "name": "Dan Cojocaru",
+            "github": "dancojocaru2000"
         }
     ],
     "sources": [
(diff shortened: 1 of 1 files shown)


jbruechert:
Hi, thanks for the pull request!

What's your reason for adding this feed in addition to the DELFI feed for all of Germany?
If it has more details (like route colors) or realtime updates are available, that absolutely makes sense.

felixguendling:
Last time I checked it had shapes. But it also contains lots of trains in NRW that would maybe be hard to filter. Only makes sense if the duplicate matcher in MOTIS can handle it.

dancojocaru2000:
Indeed, the AVV feed contains accurate shapes. 

I think the following is a feed without the trains, maybe using this would be useful to avoid conflicts: `https://opendata.avv.de/current_GTFS/AVV_GTFS_Masten_ohne_SPNV_GlobaleID_oFremd.zip`

jbruechert:
> > I think the following is a feed without the trains, maybe using this would be useful to avoid conflicts: `https://opendata.avv.de/current_GTFS/AVV_GTFS_Masten_ohne_SPNV_GlobaleID_oFremd.zip`
> 
> Should we try out with this feed @jbruechert ?

Sounds good

dancojocaru2000:
Awesome, closing in favour of #822 then

PR #788, opened 2025-01-11
Try using proxy for green-bay [US-WI]

Altonss:


jbruechert:
Yes, will merge when I'm at home so I can fix it up more easily in case it still does not work

PR #798, opened 2025-01-13
Update us-ca.json

Altonss:
Improves #201


mlundblad:
> Looking good, the only error is related to #766

OK, yeah, let's merge this and then we can look at the other issue…

PR #817, opened 2025-01-27
feat(feeds): NZ.json - added Horizons Regional Council via URL

Hoverth:
There isn't a transitland entry for this yet, so adding it by URL for the moment.


mlundblad:
Thanks!

PR #818, opened 2025-01-27
Add feeds for Washington state [US] transit authorities

cheeseandcereal:
This adds all of the feeds for transit authorities for washington state that exist on transitland which are listed here: https://mrsc.org/explore-topics/government-organization/special-districts/local-transit-authorities

This includes some missing GTFS-RT feeds which were previously missing

This also fixes the indentation of the file, as some of it was previously indented with either 4 or 5 spaces at different parts of the file.

Improves #201

### DIFF
--- a/feeds/us-wa.json
+++ b/feeds/us-wa.json
@@ -3,59 +3,239 @@
         {
             "name": "Marcus Lundblad",
             "github": "mlundblad"
+        },
+        {
+            "name": "Adam C",
+            "github": "cheeseandcereal"
         }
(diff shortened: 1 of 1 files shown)


mlundblad:
Could you add yourself as a maintainer of this feed (there is no obligation that comes with this having to fix issues, but it's good to have a record of who has been contributing to each feed).

cheeseandcereal:
Updated

jbruechert:
Thanks!

PR #850, opened 2025-02-07
Fix reuse compliance and GitHub action

1Maxnet1:
- converted the outdated .dep5 configuration of the reuse linter to the REUSE.toml
- updated the github action to the current state, so it does work properly.
- added an annotation for the robots.txt in the REUSE.toml to fix the pipeline


jbruechert:
Thanks!

PR #857, opened 2025-02-08
fetch: Don't merge identical routes

jbruechert:
Could interfere with gtfs-rt matching

### DIFF
--- a/src/fetch.py
+++ b/src/fetch.py
@@ -238,7 +238,7 @@ class Fetcher:
                    "--fix-zip",
                    "--check-null-coords",
                    "--empty-agency-url-repl", "https://transitous.org",
-                   "--remove-red-routes", "--remove-red-services", "--remove-red-stops",
+                   "--remove-red-services",
                    "--output", str(temp_file)]
         if source.fix:
             command.append("--fix")


traines-source:
Can't really imagine this is the reason. E.g. in Munich rn RT coverage is very good on my instance and nothing at api.transitous.org. Same in Hannover (with the DELFI feed), although there api.transitous.org uses the VBN feed, which then also doesn't seem to be updated correctly. I suppose the matching rate in the logs will be very low, while on my instance it's >99% currently.

traines-source:
Mmh when I'm trying to import the 2025-02-03 DELFI feed version over the 2025-01-07 version, `motis import` doesn't do anything. Or does it now always need a clean data directory? Investigating.

jbruechert:
On the CI I also delete the old files because motis doesn't notice the changes for GTFS files. I'm not sure if this is intentional, might be worth reporting.

jbruechert:
Stats for VBN:

Feb 08 14:52:09 crunchy motis[2144110]: 2025-02-08T13:52:09Z | [info][motis.rt                      ] rt update stats for tag=de-VBN, url=http://gtfsr.vbn.de/gtfsr_connect.bin: total_entities=161, total_entities_success=161 (100%)

So first that looked like an issue on their side, but stats for DELFI are also weird:

Feb 08 17:04:15 crunchy motis[2144110]: 2025-02-08T16:04:15Z | [info][motis.rt                      ] rt update stats for tag=de-DELFI, url=https://stc.traines.eu/mirror/german-delfi-gtfs-rt/latest.gtfs-rt.pbf: total_entities=72, total_entities_success=72 (100%)

traines-source:
Weird. I get sth like 100k entities for DELFI, as expected (with low match because I'm on the old static feed now, still investigating that, but should then be unrelated):
```
2025-02-08T16:56:28Z | [info][motis.rt                      ] rt update stats for tag=gtfs, url=http://stc.traines.eu/mirror/german-delfi-gtfs-rt/latest.gtfs-rt.pbf: total_entities=105544, total_entities_success=15986 (15.1463%), trip_resolve_error=88797 (84.1327%), unsupported_schedule_relationship=761 (0.721026%)
```
And according to my history, the feed was never as small that it could have been only 72 entities. Quite the opposite, I'm seeing in my logs that the transitous instance does not seem to fetch all bytes of the feed, here from the exact moment of your log above:
```
10.0.1.9 - - [08/Feb/2025:16:04:04 +0000] "GET /mirror/german-delfi-gtfs-rt/latest.gtfs-rt.pbf HTTP/1.1" 200 CACHE:- 49766990 "-" "Boost.Beast/356" "[other IP]"
10.0.1.9 - - [08/Feb/2025:16:04:15 +0000] "GET /mirror/german-delfi-gtfs-rt/latest.gtfs-rt.pbf HTTP/1.1" 200 CACHE:- 37879548 "-" "Boost.Beast/356" "[spline IP]"
```
I.e. some other consumer fetched the feed and downloaded 49766990 bytes, but when the transitous server downloaded it, it stopped after 37879548 bytes. Actually it always only downloads 25-30MB of the feed. I don't think that some kind of compression is enabled or that this is some kind of chunked transfer.
What I don't understand is how it gets 72 entities out of a presumably corrupted protobuf, or whatever is going on :thinking:

jbruechert:
At some point we had the problem of the rt fetching/matching taking more than the 60 seconds refresh interval, maybe that could explain the aborted fetches?

@felixguendling do you have some time and an idea what could be going on?

traines-source:
Ok something in the chain seems to always close the connection after about 9.5 seconds. I'll look into 1) why the transfer is so slow in the first place 2) whether I can fix the time limit in MOTIS or maybe my infrastructure. In both cases I don't understand why the VBN feed has the same problem despite only being 1MB in size and why this has ever worked before.

traines-source:
Ok I suspect that [this timeout](https://github.com/motis-project/motis/blob/cc33fb18d36f179a8b279ce336fab69d7790ffe2/src/http_req.cc#L78) actually kills the stream even if a packet has arrived, and this timeout is set to [10 seconds by default](https://github.com/motis-project/motis/blob/cc33fb18d36f179a8b279ce336fab69d7790ffe2/include/motis/config.h#L89). Since as far as I can tell the individual RT feeds are already fetched in separate threads simultaneously, I think it can't harm to increase this timeout (will send PR). At the same time, this might also explain why the download of an individual feed is so slow, because it downloads all feeds at the same time.

traines-source:
> Stats for VBN:

While yesterday evening VBN still looked bad, today it looks good. I suppose they additionally switched the static GTFS yesterday.

> On the CI I also delete the old files because motis doesn't notice the changes for GTFS files. I'm not sure if this is intentional, might be worth reporting.

This should be fixed with the latest motis release.

(Sorry for cluttering this PR, I think you know best how to proceed with the PR itself :)

jbruechert:
Since according to the spec, matching can happen based on the route_id, gtfsclean should probably not touch that.

> If the trip_id field can't be provided, then route_id, direction_id, start_date, and start_time must all be provided

PR #862, opened 2025-02-08
Fix Piemonte-Bus and deduplicate Piemonte-Trains

Fale:


feeds/it.json
```diff
@@ -123,15 +122,6 @@
             "name": "Piemonte-Trenitalia",
             "type": "transitland-atlas",
             "transitland-atlas-id": "f-u0j-trenitalia",
```
Altonss:
```suggestion
            "transitland-atlas-id": "f-u0j-trenitalia"
```

Fale: Good catch, thanks

Fale:
@jbruechert this should work on the main server since api.smartdatanet.it seems to not be blocking it. Would it be possible to check and merge?

Also, I think we need to figure out how to handle those cases where the CI is broken but prod is not, since they pollute CI tests

jbruechert:
api.smartdatanet.it is accessible on the main server

PR #871, opened 2025-02-08
Increase rt feed download timeout

traines-source:
As discussed in #857, increasing the timeout for RT downloads, since this timeout does seem to encompass the entire transfer and not only until the first packet.

(Haven't actually tested it by emulating a slow connection, but quite confident that this should improve things and not make anything else worse :)

### DIFF
--- a/motis/config.yml
+++ b/motis/config.yml
@@ -30,6 +30,7 @@ timetable:
   merge_dupes_inter_src: true
   link_stop_distance: 100
   update_interval: 60
+  http_timeout: 30
   incremental_rt_update: false
   max_footpath_length: 20
 # The datasets are filled in by ./src/generate-motis-config.py


jbruechert:
Thanks for investigating, this seems like a likely explanation

jbruechert:
Looks like it worked!

traines-source:
Seeing in my logs that the Transitous server needed up to 25 seconds to download the 70MB feed, I've investigated further and noticed that it's so slow due to apparently bad IPv6 peering between the DFN and my hosting provider Netcup (via twelve99.net). I've now removed the AAAA record for stc.traines.eu, via IPv4 the download "only" takes around 10 seconds (i.e. it is still useful that we increased the timeout).
We might want to keep in mind that it might be worth trying IPv4 only in case connections from the Transitous server fail/are unreliable (but ofc I don't really know whether the fault is more on Netcup's or on DFN's side).

jbruechert:
Eve worse the FU doesn't have native IPv6, this is a Hurricane Electric Tunnel.  Usually it is quite fast, but peering might be suboptimal.

PR #877, opened 2025-02-10
Fix it.json

Altonss:
Campania-ANM isn't available through proxy.


jbruechert:
It works on the server, I'd add fetch-interval-days to be safe and then it should be fine.

PR #909, opened 2025-02-22
Check for expired feeds

Altonss:


jbruechert:
I added some more checks for expired calendar.txt and calendar_dates.txt.
I hope I got the logic correct.

Issue #74, opened 2024-03-02
Should a GBFS feed be added for Montreal's bike sharing system?


nekohayo:
Montréal's "Bixi" bikesharing system does offer GTFS data: https://bixi.com/en/open-data/ (at the bottom)

Probably other big cities (Paris, Toronto, Lyon, etc.?) have similar systems, I think.

They're part of the public transit systems in a way, I'm not sure exactly how @mlundblad might integrate this in GNOME Maps from a UI perspective, though!

jbruechert:
I don't think so, now only the metadata needs to be extended to handle it. Should be easy enough if you know some python.

felixguendling:
I think at the moment the routing performance is still not optimal. Especially if there are many vehicle types in one area (because each one has to be routed separately). Also: the API is not stable yet (we plan to make some changes to make it more flexible). But adding feeds to transitous is still a good idea to get a first clue of what other problems might exist. Less testing = more bugs.

PR #935, opened 2025-02-24
Initial GBFS support

1Maxnet1:
This PR adds initial GBFS support as discussed in #74 
For now it only supports GBFS feeds from the transitland atlas and generates a working config.

However, for this specific feed I need to double check why it does not show up in search results, so do not merge for now.


jbruechert:
Nice work, thanks! I want to deploy this manually instead of relying on the import pipeline to do it, so I can react in case something breaks. I'll look into it in the next few days

PR #953, opened 2025-02-27
Replace proxy / use-origin options with automatic fallback

jbruechert:
Now the urls will be tried in the following order:
 - url-override (if set)
 - url (original url)
 - cache-url (Fales Transitland cache or the Mobility Database cache)

This should hopefully reduce the number of failed fetches.

**Some background**: For the Mobility Database, we currently (before this PR) use the cache by default, but there is a relevant number of feeds that they are blocked from fetching and we are not, so that didn't work out.

For reviewing, I recommend looking at the first commit individually to not see the automatically generated changes.


jbruechert:
Good point. I'd say we should just try, and if it takes unexpectedly long, we will have to revert and do what you proposed instead.

jbruechert:
>  Update of the related documentation

Interestingly, `proxy` apparently was not documented, so no change needed there. However I added the url precendence to the documentation.

PR #960, opened 2025-02-28
Add Bixi Montreal

### LINKED ISSUE #74: Should a GBFS feed be added for Montreal's bike sharing system?

Altonss:
Fix #74

### DIFF
--- a/feeds/ca-qc.json
+++ b/feeds/ca-qc.json
@@ -179,6 +179,11 @@
             "type": "transitland-atlas",
             "transitland-atlas-id": "f-societe~de~transport~de~trois-rivieres",
             "fix": true
+        },
+        {
+            "name": "Bixi-Montreal",
+            "type": "transitland-atlas",
+            "transitland-atlas-id": "f-bixi~montréal~montreal~gbfs"
         }
     ]
 }


1Maxnet1:
@nekohayo do want to add yourself as a maintainer here? Then you would get an automated ping if the data import breaks or recognizes some issues. As a local you might be able to help identify issues and find fixes more easily :)

jbruechert:
```
Error: Feed is expired, please consider removing or updating its source
Error: Could not fetch ca-qc-STTR: Feed is expired
```

This is another feed, but the CI currently can't know this

jbruechert:
> STM is considered expired, but when checking the data it seem not to be the case 🤔

Ah yes, they seem to be overblocking and have hit our proxy as well.
I can't even download the file at home.

PR #970, opened 2025-03-01
Check validity after gtfsclean processing

jbruechert:
This allows to check cases in which the feed is not well-formed before
gtfsclean, for example because the relevant files are in a subdirectory.
It also reduces the number of special cases our own parsing code needs
to handle well, like leading whitespace.


jbruechert:
Currently unsolved case: Feeds that are not yet valid. Currently the code will correctly skip them but not try again next time.

PR #971, opened 2025-03-03
Add Open Transport Meetup (OTM) recordings

### LINKED ISSUE #947: Link to the presentations about transitous
There were multiple presentation of transitous already (I'm aware of 2 at the OpenTransportMeetup, one at FOSDEM and one at CCC iirc). To make it easy to onboard, it would be nice to have them collected and linked at the homepage.

1Maxnet1:
First presentation is added at the beginning and Import Pipeline at the end. This is not only the chronological timeline but imho also makes sense as an order to watch the videos

Resolves #947 

@metters  What do you think?

### DIFF
--- a/website/content/_index.html
+++ b/website/content/_index.html
@@ -45,5 +45,7 @@ routing service that does not stop at borders.</p>
 
 <h2>Conference talks and presentations</h2>
 <p>Looking for an overview of how this project works? There are several public talks from conferences and meetups that you can watch.</p>
+<a class="btn btn-primary text-black" href="https://meet.systect.de/playback/presentation/2.3/d941f8412ff350b73e9084e0fd1da86394aff088-1708540821542">OTM Initial presentation</a>
 <a class="btn btn-primary text-black" href="https://media.ccc.de/v/38c3-transitous-offener-routingdienst-fr-ffentliche-verkehrsmittel">38C3 Video (in German)</a>
 <a class="btn btn-primary text-black" href="https://fosdem.org/2025/schedule/event/fosdem-2025-4105-gnome-maps-meets-transitous-meets-motis/">FOSDEM 2025 Video</a>
+<a class="btn btn-primary text-black" href="https://meet.systect.de/playback/presentation/2.3/d941f8412ff350b73e9084e0fd1da86394aff088-1739990087983">OTM Import pipeline</a>


metters:
> First presentation is added at the beginning and Import Pipeline at the end. This is not only the chronological timeline but imho also makes sense as an order to watch the videos
> 
> Resolves #947
> 
> @metters What do you think?

I like that it is order chronologically, but in my opinion we should add some padding between the buttons, because they look like illustrated below. I realized this already for two elements on mobile, but did not mind too much. Now with four of them we should make it look a bit better.

Screenshot 2025-03-03 at 21 02 00

Screenshot 2025-03-03 at 21 14 15


I suggest to add a small padding below the button. I don't know yet, where I should add this extra padding, maybe below
the `btn-primary` in static/custom.css, so it does not edit all other buttons?
```css
.btn-padding {
  margin-bottom: 5px;
}
```

The result looks like this:

Screenshot 2025-03-03 at 21 13 33

Screenshot 2025-03-03 at 21 14 05

Let me know if you want me to update the css like I just suggested or if it should be done in another way - or maybe just leave it like it is at the moment.

jbruechert:
Maybe it's a bit too many entries now for buttons actually

metters:
Sounds like we need a list

1Maxnet1:
> Sounds like we need a list

Agree. @metters If you are you interested into taking over, just replace the list with buttons and create a new PR. Then I'll close this one. Now you also got the links to the remaining recordings. Otherwise I will check when I find some time, but it seems to me that you would be better suited for this :)

metters:
Let me take care of it.

metters:
@1Maxnet1 @jbruechert 
Is this what you were thinking about? Except this external link icon, I refrained from adding fancy styling.

Screenshot 2025-03-05 at 21 22 30
Screenshot 2025-03-05 at 21 25 56

metters:
@1Maxnet1 @jbruechert 
I copied the code from this PR and used in for PR #973

metters:
> Minor nitpick: I'm not sure enough people know the "OTM" abbreviation, maybe we should use the full name ("Open Transport Meetup") instead?

I will update this

1Maxnet1:
Closing, as it is replaced by #973

PR #974, opened 2025-03-06
Add selector to create a config for europe

1Maxnet1:
I usually do not test the whole planet locally, but only certain countries or complete Europe. So this allows to generate a configuration file. The selection of countries is based on the list in #114 This could relatively easily be extended to allow for other areas, if there is demand. Also if you have any hints on how to do this more elegantly, I'm happy to hear it.


src/fetch.py
```diff
@@ -423,8 +423,17 @@ def fetch(self, metadata: Path):
 if __name__ == "__main__":
     fetcher = Fetcher()
 
-    metadata_file = sys.argv[1]
-    errors = fetcher.fetch(Path(metadata_file))
+    selector = sys.argv[1]
+    errors = 0
+    if selector == "europe":
```
jbruechert, line 497: We generally have ci/fetch-feeds.py for fetching multiple regions. Maybe create a new similar script for this. Having too many special cases always bears the risk of breaking some of them with every change

jbruechert (changes requested):

src/fetch.py
1Maxnet1: So instead of adding it here, creating a python script `generate-continent-config.py` ? Would be fine by me as well.

PR #975, opened 2025-03-07
Support custom HTTP headers for GBFS as well

vkrause:
Needed e.g. for the API key of the Finnish national aggregate feed.


jbruechert:
Looks good, I'll wait with merging until the import infrastructure is powered on again

PR #986, opened 2025-03-10
Update Finish feed to new Source

1Maxnet1:
The linked website mentions a deprecation of the current feed URL and wants users to migrate to the new one.

### DIFF
--- a/feeds/fi.json
+++ b/feeds/fi.json
@@ -11,12 +11,12 @@
     ],
     "sources": [
         {
-            "name": "matka",
+            "name": "fintraffic",
             "type": "http",
-            "url": "http://traffic.navici.com/tiedostot/gtfs.zip",
+            "url": "https://mobility.mobility-database.fintraffic.fi/static/finland_gtfs.zip",
             "license": {
                 "spdx-identifier": "CC-BY-4.0",
-                "url": "https://developer.matka.fi/pages/en/home.php"
+                "url": "https://www.fintraffic.fi/fi/digitaalisetpalvelut/fintrafficin-datapalvelut/liikkumisen-tietopalvelut/joukkoliikenteen-tietopalvelut/koontipalvelu"
             },
             "fix": true
         },
--- a/feeds/fi.json
+++ b/feeds/fi.json
@@ -45,67 +45,6 @@
             "name": "229",
             "type": "transitland-atlas",
             "transitland-atlas-id": "f-ued-oulunjoukkoliikenne"
-        },
-        {
-            "name": "231",
-            "type": "http",
-            "url": "https://tvv.fra1.digitaloceanspaces.com/231.zip",
-            "license": {
-                "spdx_identifier": "CC-BY-4.0",
-                "url": "https://opendata.waltti.fi/docs#gtfs-static-packages"
-            },
-            "fix": true
-        },
-        {
-            "name": "232",
-            "type": "http",
-            "url": "https://tvv.fra1.digitaloceanspaces.com/232.zip",
-            "license": {
-                "spdx_identifier": "CC-BY-4.0",
-                "url": "https://opendata.waltti.fi/docs#gtfs-static-packages"
-            },
-            "fix": true
-        },
-        {
-            "name": "237",
-            "type": "transitland-atlas",
-            "transitland-atlas-id": "f-rovaniemi~fi"
-        },
-        {
-            "name": "239",
-            "type": "transitland-atlas",
-            "transitland-atlas-id": "f-salo~fi",
-            "fix": true
-        },
-        {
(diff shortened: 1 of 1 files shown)


1Maxnet1:
As Volker pointed out in the chat, this needs further changes, as it contains duplicated feeds like e.g. The FlixBus data. Also the realtime fetching doesn't work as well as with local feeds (similarly to Germany). So marked this as draft for now.

jbruechert:
Looks like we don't have realtime data yet either, so I think it's good to go after adding a drop-agency-names for flixbus. Make sure to check the exact operator name in agency.txt

1Maxnet1:
> Looks like we don't have realtime data yet either, so I think it's good to go after adding a drop-agency-names for flixbus. 

We do have one gtfs-rt feed: digitransit. Should we remove the duplicated agencies from there as well?

> Make sure to check the exact operator name in agency.txt

Did copy them both from the dataset itself

jbruechert:
> We do have one gtfs-rt feed: digitransit. Should we remove the duplicated agencies from there as well?

Ah yes, that would make sense. You don't need to perfectly remove all duplicates, but having less to do for MOTIS and less trips that can fail to be deduplicated by MOTIS is useful.

vkrause:
The fintraffic feed is the national aggregate, ie. it replaces *all* our existing Finnish GTFS feeds (and probably adds more beyond that).

Overall this is good, as it save us some maintenance work. I have trouble getting the corresponding realtime feed to work though.

```
{
    "name": "fintraffic",
    "type": "url",
    "spec": "gtfs-rt",
    "url": "https://mobility-api.mobility-database.fintraffic.fi/gtfs-realtime/v2",
    "license": {
        "spdx-identifier": "CC-BY-4.0",
        "url": "https://mobility.mobility-database.fintraffic.fi/fi"
    },
    "headers": { "x-api-key": "y5PwtXqIfS61AzBElMS48aZXjKizpH2r90ISgMTd" }
}
```

That at least loads but matching rates are abysmal/DELFI-grade, a clear step back from the realtime feeds we currently have.

Btw, there's also an aggregated GBFS feed that works with the same key. I'd suggest to deal with that separately, just mentioning that here for reference.

```
{
    "name": "fintraffic-gbfs",
    "type": "url",
    "spec": "gbfs",
    "url": "https://mobility-api.mobility-database.fintraffic.fi/gbfs/v3/gbfs.json",
    "license": {
        "spdx-identifier": "CC-BY-4.0",
        "url": "https://mobility.mobility-database.fintraffic.fi/fi"
    },
    "headers": {
        "x-api-key": "y5PwtXqIfS61AzBElMS48aZXjKizpH2r90ISgMTd"
    }
}
```

With latest MOTIS that loads, but I yet have to actually verify that this gives us usable vehicles.

1Maxnet1:
> The fintraffic feed is the national aggregate, ie. it replaces all our existing Finnish GTFS feeds (and probably adds more beyond that).

I just checked the digitransit feed and the fintraffic feed. Only 3 of the agencies can be found in both, so not sure, how complete the national aggregated feed is and whether it replaces all of them (yet).

> Overall this is good, as it save us some maintenance work. I have trouble getting the corresponding realtime feed to work though.

What about doing this in a separate step? For now we could just replace the feed by the new one, remove the feeds that are included in the aggregated feed and for the one realtime feed we can use the workaround as used in de.json.

vkrause:
That works for me, I just wouldn't want to lose the realtime coverage of long distance trains we have right now.

1Maxnet1:
> That works for me, I just wouldn't want to lose the realtime coverage of long distance trains we have right now.

I removed most of teh remaining feeds. Kept one, as it contains an agency name that I could not find in the fintraffic feed. Can you double check that realtime still works as expected? Otherwise from my side this is ready for review / merge. I created #987 to keep track of the GTFS-RT and GBFS feeds, that can still be added

1Maxnet1:
> Realtime data looks the same as before, and all other places I checked have the same or more data now. With one exception though, in Helsinki we seem to miss busses and metros now.

Maybe we can re-add `hsl` feed then? Can have a look tomorrow whether this fixes it.

jbruechert:
Please run `git fetch origin && git rebase origin/main` and force push the result. Currently GitHub is very confused and shows the commits you (cherry picked?) from main as changes of this PR.

1Maxnet1:
I re-added the hsl feed and busses and metro are back in Helsinki.

Sadly I also decided to rebase, which the GitHub UI seems to interprate as a lot of changes. If I should clean that chaos up by removing rebased commits again, so it's easier to review let me know.

vkrause (approved):
Looks good now, thanks!

PR #992, opened 2025-03-11
Fix IT

Fale:


jbruechert:
We always use the latest version of transitland-atlas anyway, but updating it doesn't hurt :)

PR #1001, opened 2025-03-13
Add PMDP to cz.json

MartinNemi03:
Adding Pilsen and it's public transport to cz.json (based on https://github.com/transitland/transitland-atlas/pull/1470)

### DIFF
--- a/feeds/cz.json
+++ b/feeds/cz.json
@@ -49,6 +49,11 @@
             "name": "Bean-Shuttle",
             "type": "mobility-database",
             "mdb-id": 2036
+        },
+        {
+            "name": "PMDP",
+            "type": "transitland-atlas",
+            "transitland-atlas-id": "f-plzen"
         }
     ]
 }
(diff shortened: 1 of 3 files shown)


mlundblad:
I'm not sure if maybe "fix": true would fix the issue in this feed.

jbruechert:
So the problem here is that the file has a UTF-8 BOM at the beginning. I have no idea if this is allowed in GTFS, but gtfsclean doesn't seem to handle it. For now, you can use my own little csv parser, into which I quickly hacked in BOM support.  I'll leave a suggestion you can apply.

feeds/cz.json
```diff
@@ -49,6 +49,11 @@
             "name": "Bean-Shuttle",
             "type": "mobility-database",
             "mdb-id": 2036
+        },
+        {
+            "name": "PMDP",
+            "type": "transitland-atlas",
+            "transitland-atlas-id": "f-plzen"
```
jbruechert:
```suggestion
            "transitland-atlas-id": "f-plzen",
            "fix": true,
            "fix-csv-quotes": true
```

jbruechert:
Will have a look later to find out why everything is being dropped in postprocessing

jbruechert:
Thanks!

PR #1003, opened 2025-03-14
Fix compatibility with Python 3.9 - 3.11 in generate-france.py

1Maxnet1:
This is a minor fix to make the script also work on older Python versions.

### DIFF
--- a/src/generate-france.py
+++ b/src/generate-france.py
@@ -217,7 +217,7 @@ if __name__ == "__main__":
                                 )
                         else:
                             print(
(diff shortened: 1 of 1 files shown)


jbruechert:
Are you running debian oldstable or something? :D

Thanks for the fix :)

1Maxnet1:
> Are you running debian oldstable or something? :D

On my development machine, no there it's Manjaro Linux. However the server I'm currently testing on runs Ubuntu 22.04. which is still on 3.10. It was just easier to fix this little issue then upgrading to 24.04. now :)

Issue #988, opened 2025-03-10
[GBFS] Bixi Montreal not used for routing


Altonss:
With `Allow usage of sharing vehicles` activated, Bixi Montreal doesn't seem to be used for routing.

jbruechert:
I couldn't find anything suspicious in the logs on the server, but the server for regiorad Stuttgart is down right now. Maybe MOTIS doesn't handle this case to well yet. I'd recommend setting up MOTIS and only enable the bixi feed and later try to add the failing regiorad one as well to see whether that changes anything.

felixguendling:
Should be fixed in v2.0.32. Please let me know if you find any further issues.

pablohoch:
> cc [@felixguendling](https://github.com/felixguendling?rgh-link-date=2025-03-22T14%3A17%3A11.000Z) [@pablohoch](https://github.com/pablohoch?rgh-link-date=2025-03-22T14%3A17%3A11.000Z)
> 
> After quickly looking at the code, it seems that the `num_vehicles_available_key` might be set to the incorrect key for GBFS v1 feeds (which Bixi Montreal is), it should be `num_bikes_available` (like v2 feeds) I think. [motis-project/motis@`c835bfa`/src/gbfs/parser.cc#L247-L249](https://github.com/motis-project/motis/blob/c835bfa7813293db5fdeef2fbb5062885c392c9f/src/gbfs/parser.cc#L247-L249) Might this be the source of the bug?

Yes, thanks for finding that bug! I hadn't previously tested it with the Bixi Montreal feed, but it should work now in MOTIS v2.0.44.

felixguendling:
I introduced a bug in v2.0.44 with the geocoding.
Should be fixed in v2.0.45.

PR #1007, opened 2025-03-14
Add some Bikesharing feeds for DE-BW

1Maxnet1:
- Replace RegioRadStuttgart with Call-a-bike as it's included. Currently however both feeds are down :/, but I reached out to MobiData for investigation what is currently happening.
- Add Nextbike feeds from https://www.mobidata-bw.de/dataset/bikesh
Checked in a local instance, that all of them get picked up

### DIFF
--- a/feeds/de.json
+++ b/feeds/de.json
@@ -94,9 +94,34 @@
             }
         },
         {
-            "name": "RegioRadStuttgart",
+            "name": "CallaBike",
             "type": "transitland-atlas",
-            "transitland-atlas-id": "f-regioradstuttgart~stuttgart~gbfs"
+            "transitland-atlas-id": "f-call~a~bike~germany~gbfs"
+        },
+        {
+            "name": "KVV.nextbike",
+            "type": "transitland-atlas",
+            "transitland-atlas-id": "f-kvv~nextbike~germany~gbfs"
+        },
+        {
+            "name": "FreloFreiburg",
+            "type": "transitland-atlas",
+            "transitland-atlas-id": "f-frelo~freiburg~gbfs"
+        },
+        {
+            "name": "SAPWalldorf",
+            "type": "transitland-atlas",
+            "transitland-atlas-id": "f-sap~walldorf~gbfs"
+        },
+                {
+            "name": "EinfachMobil",
+            "type": "transitland-atlas",
+            "transitland-atlas-id": "f-einfachmobil~ortenaukreis~gbfs"
+        },
+        {
+            "name": "VRNnextbike",
+            "type": "transitland-atlas",
+            "transitland-atlas-id": "f-vrnnextbike~rhine~neckar~gbfs"
         }
     ]
 }


jbruechert:
I think it would make sense to wait until the new MOTIS release is deployed before adding new GBFS feeds, as currently they would fail to download anyway.

derhuerst:
MobiData-BW ingests sharing data from several providers that
- have a proprietary API (e.g. [Deer](https://github.com/mobidata-bw/x2gbfs/blob/f327011490c87b4e4690dc42c18228af7db3f9fe/x2gbfs/providers/deer.py)); or
- require authentication and/or a contract to access the data; or
- have errors, quirks and insufficient details in their data.

It republishes all ingested feeds [as a single entrypoint with a list of `feeds`](https://api.mobidata-bw.de/sharing/gbfs) or (better) [as a GBFS 3.0 `manifest.json` linking to the other feeds](https://api.mobidata-bw.de/sharing/gbfs/v3/manifest.json). You should be able to consume all of them in Transitous, with exceptions due to proprietary licensing (see `license_url`/`license_id` in each feed's `system_information`) of course.

If, for those feeds that provide open and spec-compliant GBFS, you don't want to have MobiData-BW as an intermediary, you could only consume the closed and/or broken providers via MobiData-BW.

cc @hbruch

felixguendling:
Since the Mobidata BW feed has been used during development, I would hope that it works :)

https://github.com/motis-project/motis/pull/621

Open an issue on motis-project/motis if not.

1Maxnet1:
> If, for those feeds that provide open and spec-compliant GBFS, you don't want to have MobiData-BW as an intermediary, you could only consume the closed and/or broken providers via MobiData-BW.

Thanks for the heads-up and additional information. I thought as well, whether or not it makes sense to use the republished / aggregated feeds of Mobidata-BW, instead of adding original / individual feeds. Here are my thoughts:

* using the aggregated feed of all feeds would mean we have some feeds in there, that are not specific for Germany, but from Switzerland and France. As the feeds are currently per country, this could lead to duplicating the feeds when adding them e.g at ch.json and de.json. Not sure though what the consequences would be, if e.g. we add the repubslished and original feeds in such a case. Would then all information be doubled?

* one basic reason why I choose to add the individual feeds in #935 was that adapting the script to read the id from transitland-atlas was easy. However the aggregated feed there is not available in transitland-atlas, the individual feeds are. This can be fixed ofc.

* adding Mobidata-BW as intermediate republisher adds one layer of failure. At least currently the feeds seem to have an issue to be updated, as the last update is from 04.03. while we have no other possibility for feeds that are created by Mobidata e.g. Deer Carsharing the original nextbike feeds added in this PR are still updated and continue working even if issues at Mobidata happen.

* you mentioned already the proprietary licensing,so these feed would need to be filtered out if we choose an aggregated feed.

* While aggregation involves some steps for GTFS datasets (removing duplicates etc.) from my understanding GBFS feeds should be disjunct (as they are from different providers. So I do not see that big of an advantage in using the aggregated feed compared to adding the individual entries in `de.json`

* if individual feeds
> have errors, quirks and insufficient details in their data.

We can still use the republished feeds of MobiData-BW until the issues are resolved upstream

But I don't decide this alone. Would appreciate some other opinions on this. Just wrote down what came up during implementing #935

derhuerst:
> * one basic reason why I choose to add the individual feeds in [Initial GBFS support #935](https://github.com/public-transport/transitous/pull/935) was that adapting the script to read the id from transitland-atlas was easy. However the aggregated feed there is not available in transitland-atlas, the individual feeds are. This can be fixed ofc.

This is something MobiData-BW and Transitland should consider. 👍

(@hbruch & @ThorstenFroehlinghaus, let's discuss this somewhere else and give an update here later.)

> * adding Mobidata-BW as intermediate republisher adds one layer of failure.

Indeed. This week, we've had some major uptime issues. Before that, it mostly worked.

> At least currently the feeds seem to have an issue to be updated, as the last update is from 04.03. […].

Which one? 😮

Could it be that you checked the `last_updated` field of a `gbfs.json` file? For example, `https://api.mobidata-bw.de/sharing/gbfs/v3/deer/gbfs` currently has `2025-03-04T08:28:57.000+00:00` while `https://api.mobidata-bw.de/sharing/gbfs/v3/deer/station_status` has `2025-03-14T14:35:07.000+00:00`.

> * While aggregation involves some steps for GTFS datasets (removing duplicates etc.) from my understanding GBFS feeds should be disjunct (as they are from different providers. So I do not see that big of an advantage in using the aggregated feed compared to adding the individual entries in `de.json`

*The aggregation itself* in MobiData-BW provides no additional value to Transitous, true.

However, the error fixing & "quality of life" improvements (e.g. [in our HTTP proxy](https://github.com/mobidata-bw/ipl-proxy/tree/77cfab4e5c10f168102f31102eb36b6ffc9cf404/app/converters) and [in x2gbfs](https://github.com/mobidata-bw/x2gbfs/tree/f327011490c87b4e4690dc42c18228af7db3f9fe/docs/mappings), which are two components behind it) might be beneficial: Transitous/MOTIS won't have to add special-casing to the code just because a single feed doesn't comply with the spec or is lacking essential information.

1Maxnet1:
> Which one? 😮

The aggregated one as per homepage here: https://www.mobidata-bw.de/dataset/bikesh
I didn't investigate any further (yet) , I just wondered why the RegioRad feed does not show up when I run transitous locally and found that the feed is marked as last updated 04.03. and then stopped my investigation for the moment, hoping that it will be back soon and asking via mail.

> Could it be that you checked the last_updated field of a gbfs.json file? For example, https://api.mobidata-bw.de/sharing/gbfs/v3/deer/gbfs currently has 2025-03-04T08:28:57.000+00:00 while https://api.mobidata-bw.de/sharing/gbfs/v3/deer/station_status has 2025-03-14T14:35:07.000+00:00.

Yes I checked that as well for the nextbike feeds I added in this PR and the RegioRad/Call-A-Bike feeds and saw that CallABike/RegioRad is stuck at 04.03.

> If it turns out that some feeds are indeed not updating, please create an issue in [ipl-orchestration](https://github.com/mobidata-bw/ipl-orchestration). 🙏

Done: https://github.com/mobidata-bw/ipl-orchestration/issues/348

> However, the error fixing & "quality of life" improvements (e.g. [in our HTTP proxy](https://github.com/mobidata-bw/ipl-proxy/tree/77cfab4e5c10f168102f31102eb36b6ffc9cf404/app/converters) and [in x2gbfs](https://github.com/mobidata-bw/x2gbfs/tree/f327011490c87b4e4690dc42c18228af7db3f9fe/docs/mappings), which are two components behind it) might be beneficial: Transitous/MOTIS won't have to add special-casing to the code just because a single feed doesn't comply with the spec or is lacking essential information. Long-term, this might make the maintenance of Transitous' DE list of data sources more sustainable.

Agree, I would check individual feeds and then evaluate where the Quality of Live improvements outway the risk of a failure on the MobiData side and am happy to work on improving these (btw awesome tools) further. However, Transitous and MOTIS need some robustness for issues sooner or later, as we probably do not only want to add GBFS data from DE-BW. In e.g. #988 this already lead to a fix on the MOTIS side.

jbruechert:
Has any conclusion been reached here?
I don't have a strong preference, feel free to merge if you think it is fine as it is.

1Maxnet1:
> Has any conclusion been reached here? I don't have a strong preference, feel free to merge if you think it is fine as it is.

I merge it as is. If we find any issues with individual GBFS feeds that are also reproduced at MobiDataBW, we can still switch later, if beneficial. MobiData also has a Barcamp at the beginning of April which I'll attend, so maybe I can discuss this topic there in person as well :)

One more found: Call-A-Bike does not include the RegioRadStuttgart, so I re-add it.

PR #1008, opened 2025-03-14
Add Sardegna feed from transitland

fbarbe00:


mlundblad:
We already have the  "Sardegna-Trenitalia" feed (with a "direct" URL).
Should this replace that feed?
I'm guessing otherwise it will duplicate that one?

fbarbe00:
The two URLs differ. I downloaded both GTFS files and they also differed in file size, but I haven't had time to look deeper into it. The transit land one had more content


________________________________

feeds/it.json
```diff
@@ -245,6 +246,11 @@
                 "spdx-identifier": "CC-BY-4.0"
             }
         },
+        {
+            "name": "Sardegna-Trenitalia-1",
+            "type": "transitland-atlas",
+            "transitland-atlas-id": "f-spn-trenitalia"
+        },
```
Altonss, line 253: Instead of adding this new one, you can just move the existing one to the transitland feed.

PR #1021, opened 2025-03-20
Step down as fr maintainer

afranke:
I never really understood how to handle issues and there are enough other maintainers now.


jbruechert:
Aright, thanks for your work so far!

PR #1024, opened 2025-03-21
Regenerate FR

Altonss:


jbruechert:
idea:
We could generate an `url` and a `url-override` field, where `url` points to the mirrored gov portal url (even though that doesn't always work), and `url-override` to the upstream url. That should provide a bit more stable fetching overall.

jbruechert:
Just be aware that if there are expired feeds that we previously didn't have at all, motis import will likely fail because the current logic is too not store the file at all if it is expired I think.

Apart from that feel free to merge. We can always change that logic if it causes problems. Apart from that expired feeds are probably a sign that the items on the portal are no longer used.

PR #1030, opened 2025-03-25
Update Motis to v2.0.45

Altonss:
@jbruechert I don't know if it the correct process for bumping Motis version (I don't know If this PR should also update CI and ansible :thinking: ), so feel free to close the PR if it's not right.


jbruechert:
That's correct. After merging you need to create a tag, and the CI will build a release with an optimized MOTIS executable.

PR #1044, opened 2025-04-02
Update generate-france.py to support GBFS feeds + regenerate FR

### LINKED ISSUE #1029: Add GBFS support in France generator
Technical improvement needed:
- [x] Support `url` feed types for GBFS data

Altonss:
Fix #1029

### DIFF
--- a/feeds/fr.json
+++ b/feeds/fr.json
@@ -24,27 +24,6 @@
     ],
     "sources": [
         {
-            "name": "horaires-des-trains-trenitalia-france",
-            "type": "http",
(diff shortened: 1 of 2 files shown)


jbruechert:
This amount of new GBFS feeds is likely to expose some new bugs, I think it would be good to test locally before merging.

jbruechert:
I tested this on the staging instance now, seems to at least not to crash anything

jbruechert:
> > I tested this on the staging instance now, seems to at least not to crash anything
> 
> If it works fine on staging, I think we could merge this soon?

Yes, just needs a rebase but good to go apart from that.

jbruechert:
Thank you!

PR #1054, opened 2025-04-06
Url override for OC Bus in California

kylerchin:
This resolves https://github.com/public-transport/transitous/issues/1050


kylerchin:
@jbruechert

jbruechert:
Thank you!

PR #1058, opened 2025-04-12
Add traewelling to app card row

MrKrisKrisu:


website/static/images/apps/traewelling.svg
Altonss, line 1: Shouldn't the SVG shape be square?

jbruechert: image

Looks reasonable on the website, I think it's fine

jbruechert:
Thanks!

PR #1061, opened 2025-04-12
Add Münchner Verkehrsgesellschaft (MVG) as de feed

Alpakat:
Hey,
this is my first time working with GTFS / Transitous. Hope I got everything right.

I noticed that the current headsigns for the Munich U-Bahn all show the wrong destination `Parkhaus Blumengroßmarkt`. For example, https://api.transitous.org/api/v1/trip?tripId=20250412_20%3A13_de-DELFI_2783325967 has `Parkhaus Blumengroßmarkt` as the headsign, even though it should be `Mangfallplatz`.
This is due to DELFI providing the wrong data: `de:mvg:1|U1|U:_1,2,Parkhaus Blumengroßmarkt,,0,,161945,2783325967`

The feed at https://mobilitydatabase.org/feeds/gtfs/mdb-2333 contains the correct headsigns. So I tried to replace it.

Hope this helps!
Thank you.

### DIFF
--- a/feeds/de.json
+++ b/feeds/de.json
@@ -130,7 +130,7 @@
             "transitland-atlas-id": "f-vrnnextbike~rhine~neckar~gbfs"
         },
         {
-            "name": "Münchner Verkehrsgesellschaft (MVG)",
+            "name": "Münchner-Verkehrsgesellschaft(MVG)",
             "type": "mobility-database",
             "mdb-id": 2333
         }
--- a/feeds/de.json
+++ b/feeds/de.json
@@ -31,7 +31,8 @@
                 "Braunschweiger Verkehrs-GmbH",
                 "Kraftverkehrsgesellschaft mbH Braunschweig",
                 "SWO Mobil GmbH",
-                "FlixBus-de"
+                "FlixBus-de",
+                "U-Bahn München"
             ],
             "http-options": {
                 "fetch-interval-days": 2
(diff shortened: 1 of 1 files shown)


jbruechert:
Thank you for working on this! Unfortunately that would mean to lose realtime coverage in Munich, which is also not so good. Are there more such quality issues that would make a better but only static schedule more useful than the current state?

Alpakat:
Hey,
thank you for your quick response!
Because I use transitous through Träwelling, I think it would be more useful to know, which U Bahn to select for check in, rather than knowing if it is one minute late. Also because U Bahn trains don't really have delays or train numbers.
What needs to be tested is, which busses are included in the MVG feed and therefore need to be excluded from the DELFI feed.

However I understand that one data source that is mostly correct and has real time data is better, than this.
I already opend an issue with GTFS to try to get this resolved with the DELFI feed.
I don't know how fast they will be.
Maybe a hotfix, taking the name of the destination of the trip and replacing the heading with it could be implemented, but I don't know it this is beyond the scope of this project, because I think it gets messy pretty quick.

So let's hope this quickly gets resolved upstream.

feeds/de.json
```diff
@@ -127,6 +128,11 @@
             "name": "VRNnextbike",
             "type": "transitland-atlas",
             "transitland-atlas-id": "f-vrnnextbike~rhine~neckar~gbfs"
+        },
+        {
+            "name": "Münchner Verkehrsgesellschaft (MVG)",
```
Altonss:
```suggestion
            "name": "Münchner-Verkehrsgesellschaft(MVG)",
```
`name` cannot contain spaces

PR #1062, opened 2025-04-13
Improve attribution

traines-source:
Adds link to https://transitous.org/sources/ in the UI and as a response header of every API call (this needs not-yet-released motis 2.0.49, but shouldn't break anything rn). Adds attribution to OSM to the sources page.


jbruechert:
Thanks!

PR #1064, opened 2025-04-13
Add licenses + regenerate FR

Altonss:
Quick fix to add the spdx license identifiers to FR (so we can merge safely before #1044).
This should also improve downstream issues by providing more accurate license info, see https://github.com/Traewelling/traewelling/issues/3338 .

Small things I am not 100% sure about (cc @jbruechert):
- Etalab Open License 1.0 can be safely marked as Etalab Open License 2.0, they should be compatible so shouldn't be an issue, see https://www.etalab.gouv.fr/licence-ouverte-open-licence/. Here citation of compatibility mention of the 2.0 version : `This licence has been designed to be compatible with any free licence that at least
requires an acknowledgement of authorship, and specifically with the previous
version of this licence as well as with the following licences`
- The ODbL-1.0 used on the https://transport.data.gouv.fr portal, seems to have an additionnal conditions to make the 4.4 Share Alike less restrictive https://doc.transport.data.gouv.fr/le-point-d-acces-national/cadre-juridique/conditions-dutilisation-des-donnees/licence-odbl, so the more restrictive ODbL-1.0 identifier should be fine IMHO (as it is also what the API shows anyway (`odc-odbl`))

### DIFF
--- a/feeds/fr.json
+++ b/feeds/fr.json
@@ -29,7 +29,8 @@
             "url": "https://thello.axelor.com/public/gtfs/gtfs.zip",
             "fix": true,
             "license": {
-                "url": "https://transport.data.gouv.fr/datasets/horaires-des-trains-trenitalia-france"
+                "url": "https://transport.data.gouv.fr/datasets/horaires-des-trains-trenitalia-france",
+                "spdx-identifier": "etalab-2.0"
             }
         },
         {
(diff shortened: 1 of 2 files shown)


jbruechert:
I was working on the same thing xD

jbruechert (approved):

jbruechert:
Feel free to merge when you are ready.

PR #1083, opened 2025-04-21
Add GTFS.de feed with Poland-Germany trains

kasmar00:
**Why?**
PKP-Intercity feed has international trains ending at last station in Poland. Trains from Poland to Berlin are in the gtfs.de DB feed (but are not in the main german DELFI feed). It feels like a quick fix can be made, to have data for those trains.

Similar issue exists on trains for example from Poland to Czechia, but at the moment transitous has data for those trains in Czechia and uggests a transfer between them (while not ideal, no data is missing).

**How?**
Use `f-germany~long~distance~rail` (gtfs.de long distance feed) with all agencies but PKP Intercity filtered out.
This will cause only small duplication (double the PL-DE trains on Polish side), but will allow to navigate on full routes and on route parts inside Germany.

Small change was required to make `drop-agency-names` propagate properly.

### DIFF
--- a/src/transitland.py
+++ b/src/transitland.py
@@ -43,6 +43,7 @@ class Atlas:
             result.drop_shapes = source.drop_shapes
             result.fix_csv_quotes = source.fix_csv_quotes
             result.display_name_options = source.display_name_options
+            result.drop_agency_names = source.drop_agency_names
 
             if source.url_override:
                 result.url_override = source.url_override
--- a/feeds/pl.json
+++ b/feeds/pl.json
@@ -20,6 +20,26 @@
             "transitland-atlas-id": "f-pkp~intercity~pl"
         },
         {
+            "name": "PKP-Intercity-Germany",
+            "type": "transitland-atlas",
+            "transitland-atlas-id": "f-germany~long~distance~rail",
+            "drop_agency-names": [
+                "DB Fernverkehr (Codesharing)",
+                "HZZP",
+                "DB Fernverkehr AG",
+                "ZSSK",
+                "ÖBB",
+                "Ceske Drahy",
+                "MAV",
+                "Nederlandse Spoorwegen",
+                "Trenitalia",
+                "SBB",
+                "Dänische Staatsbahnen",
+                "SNCF",
+                "SÜWEX"
+            ]
+        },
+        {
             "name": "PolRegio",
             "type": "transitland-atlas",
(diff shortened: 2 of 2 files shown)


derhuerst:
@OpenDataVBB/GTFS Would it be possible for the VBB to include those Polish-German trains that stop anywhere in Berlin/Brandenburg?

kasmar00:
> Would it be possible for the VBB to include those Polish-German trains that stop anywhere in Berlin/Brandenburg?

That was my other though (as a train from Warsaw to Munich is included in the DELFI datasource - but it's a different Land), but I didn't know where to start 😅

jbruechert:
I'm not generally opposed to your approach, but I'm slightly surprised that it works. As far as I know GTFS.de only does DELFI NeTEx to GTFS conversion after all, which means we should already have pretty much the same data, or DELFI does something weird in their own converter.
Obviously we can go with your solution at least as a temporary measure, and talk to the DELFI people at their next event.

hbruch:
DELFI NeTEx contains more routes (lines in NeTEx parlance) than DELFI provides via GTFS. I.e. when generating the GTFS, they filter lines tagged with `Notices` with `PublicCode` eq `GTGSP` or eq `HFGSP ` (see https://github.com/mfdz/GTFS-Issues/issues/170 for details). 
GTFS.de seems to ignore these tags. 

The NeTEx feed contains lines with destination `Warszawa` in the following sub directories:

* DBDB_NZ____
* DBDB_51____
* DBDB_80____

jbruechert:
Okay, then let's go with this for now.

src/transitland.py
```diff
@@ -43,6 +43,7 @@ def source_by_id(self, source: TransitlandSource) -> Optional[Source]:
             result.drop_shapes = source.drop_shapes
             result.fix_csv_quotes = source.fix_csv_quotes
             result.display_name_options = source.display_name_options
+            result.drop_agency_names = source.drop_agency_names
```
jbruechert, line 46: Thanks for this fix, this design is a significant source of bugs lately, ideas on how to improve it are very welcome.

felixguendling: If it helps, dropping all trips from a specific agency could also be implemented in MOTIS and become part of the config.yml.

jbruechert: The actual dropping part is fairly reliable, but internal conversion from database sources to http sources requires manually copying each new attribute, which I always forget to add in one place or another.

PR #1082, opened 2025-04-20
Update feeds for Poland

kasmar00:
- Add RT for Warsaw Public Transport
- Add Warszawska Kolej Dojazdowa
- Add Koleje Śląskie and Łódzka Kolej Aglomeracyjna

PR #1090, opened 2025-04-22
Add koleje wielkopolskie and cleanup pl

kasmar00:
- Added Koleje Wielkopolskie
- Cleanup Łódzka Kolej Aglomeracyjna and Koleje Śląskie (followup from #1082)
- Pull transitland-atlas to latest version (required for other changes)


jbruechert:
Btw, the CI always uses the latest version of transitland-atlas, but it makes sense for local testing to update the submodule nevertheless.

PR #1096, opened 2025-04-24
Add some German carsharing providers

1Maxnet1:
Adding some German carsharing providers and especially some from Baden-Württemberg. I tested the setup locally and the server starts without any issues. Also in my tests they got picked up for routing, although I did not test all individual providers.


mlundblad:
Thanks!

PR #1105, opened 2025-04-27
Add feed for Hyderabad Metro

Neo2308:
* Add region for Telangana (India)
* Add feed for hyderabad metro


jbruechert:
Thank you!

PR #1108, opened 2025-04-27
Update license for CZPTT feed in cz.json

MartinNemi03:
Recently, I asked lead maintainer of CZPTT feed on their jrutil GitLab repo about the license and while the answer wasn't as conclusive as I'd like it to be _(if only the ministry of transport had a proper license attached to source data)_, I believe the response goes give more than enough reason to make this a CC0.

https://gitlab.com/dvdkon/jrutil/-/issues/13#note_2469249024


jbruechert:
Sounds reasonable to me, but I'll wait a bit for others to add their interpretations.

PR #1111, opened 2025-04-28
Yet another pl.json update

Lach-anonym:


feeds/pl.json
```diff
@@ -205,7 +203,12 @@
         {
             "name": "ZTP-Kraków",
             "type": "mobility-database",
-            "mdb-id": "mdb-2339"
+            "mdb-id": "mdb-1326"
+        },
+        {
+            "name": "ZTP-Kraków",
+            "type": "mobility-database",
+            "mdb-id": "mdb-2598"
```
jbruechert, line 214: This is no realtime feed so it can not have the same name as the previous one. Do you know what the difference between them is?

Lach-anonym: ZTP Kraków maintains seperate files for buses (each for MPK [A] and Mobilis [M] operators) and for trams (T). There are no static feeds for all of those due to technical reasons on ZTP side (but for realtime - yes).

jbruechert: They all look static only?
GTFS doesn't have a realtime-only variant.

Lach-anonym: Yes, all of them are static
I've sent the RT links to the Mobility Database, as soon as the feed gets added there, I'll update the file again

jbruechert: Can you change the name of the feed then so this can be merged please?
(Two static feeds can't have the same name)

Lach-anonym: Ah, whoops, I just wanted to keep my fork updated, didn't think it would notify people
I'll look into that

feeds/pl.json
```diff
@@ -146,10 +146,9 @@
         },
         {
             "name": "Elbląg",
-            "type": "transitland-atlas",
-            "transitland-atlas-id": "f-u3tu-elblag",
-            "skip": true,
-            "skip-reason": "Feed unavailable"
+            "type": "mobility-database",
+            "mdb-id": "mdb-2597",
+            "spdx_identifier": "CC0-1.0"
```
jbruechert: This currently does not work for databases, and it is the wrong json key anyway.

Lach-anonym: so it is not possible yet to implement the licenses for mdb yet?

jbruechert:
Thanks!

Lach-anonym:
No data from Elbląg seem to appear in Transitous, is it because of the SPDX identifier? If so, should I change it to a direct link?

jbruechert:
We don't use the spdx-identifier for anything apart from documentation so far. 
Looks like the import job failed for three days in a row

PR #1115, opened 2025-04-30
Update the MOTIS version in the containerfile

1Maxnet1:
When someone follows the instructions for running an instance locally, it will miss out the recent bugfixes, so I think it makes sense to update the MOTIS version here as well

### DIFF
--- a/ci/container/Containerfile
+++ b/ci/container/Containerfile
@@ -7,7 +7,7 @@ RUN GOPROXY=direct GOBIN=/usr/local/bin/ go install github.com/public-transport/
 
 FROM docker.io/debian:bookworm-slim
 
-ARG MOTIS_VERSION=v2.0.34
+ARG MOTIS_VERSION=v2.0.59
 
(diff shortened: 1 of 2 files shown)


jbruechert:
I think it makes sense to keep it on sync with the actually deployed version, which is 2.0.34.
Otherwise you can't really use it to test wether your changes will work on the main instance.

1Maxnet1:
> I think it makes sense to keep it on sync with the actually deployed version, which is 2.0.34. Otherwise you can't really use it to test wether your changes will work on the main instance.

Agree, then I misunderstood the latest updates. Thought it is already at .59 Sorry for the confusion

jbruechert:
.59 is deployed on the staging instance and works fine there. 

Now I just need to re-import OSM before deploying it to the main instance as well.

jbruechert:
.59 is deployed on the staging instance and works fine there. 

Now I just need to re-import OSM before deploying it to the main instance as well.

PR #1116, opened 2025-05-02
Multiple updates for ro.json

FloreaCostinMario:
- List cleaned up and ordered alphabetically, all feeds now start with capital letters
- Moved most feeds over from http to their existing mobility db entries
- Oradea static feed added


jbruechert:
Thanks!

PR #1129, opened 2025-05-08
Rename IN-TG to IN-TS

1Maxnet1:
As this is the official ISO-Code according to https://en.wikipedia.org/wiki/ISO_3166-2:IN Otherwise source generation script fails on this

@Neo2308 as you are listed as maintainer


jbruechert:
see https://github.com/public-transport/transitous/pull/1106

jbruechert:
you can however switch the container image to using pip to install the python dependencies, then this should be fine.

1Maxnet1:
Thanks for the hint.

> you can however switch the container image to using pip to install the python dependencies, then this should be fine.

Yeah, will have a look at it :+1:

PR #1130, opened 2025-05-09
Add Transportflow iOS app to website

hannsadrian:
Hey everyone, I've built an iOS App called "Transportflow" using the Transitous API and thought it would be great if it could be mentioned on the project website. That's what this pr contains :)

PR #1138, opened 2025-05-13
Update Transportflow App Website

hannsadrian:
Like [discussed](https://github.com/public-transport/transitous/pull/1130#issuecomment-2868705710) in #1130, I've now set up a proper support/product page for my App Transportflow. I've updated the links in the App Store to not point to transitous.org and would like to link the product page in transitous as well.


jbruechert:
Thanks!

PR #1139, opened 2025-05-13
Croatia license update

Lach-anonym:
Source: https://data.rijeka.hr/dataset/at-vozni-redovi/resource/aed95e3c-47fd-4689-b0f3-6e60f573af73


Lach-anonym:
Also I've looked at the Sisak and it seems fine to me - calendar works, there are routes and trips, but I'm not 100% sure

PR #1141, opened 2025-05-14
Bring back B-Bus feed

Lach-anonym:
Ironically Visi Maršrutai doesn't have all the routes, I've checked the Panėvežys area and found out the suburban lines were missing. So I've decided to add the B-Bus company group feed again, but this time in the proper country.

### DIFF
--- a/feeds/lt.json
+++ b/feeds/lt.json
@@ -80,8 +80,8 @@
         },
         {
             "name": "B-Bus",
-            "type": "http",
-            "url": "https://stops.lt/LVA/lva/gtfs.zip"
+            "type": "mobility-database",
+            "mdb-id": "mdb-2385"
         }
     ]
 }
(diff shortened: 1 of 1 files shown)


feeds/lt.json
```diff
@@ -77,6 +77,11 @@
             "name": "traukiniai",
             "type": "http",
             "url": "https://stops.lt/traukiniai/traukiniai/gtfs.zip"
+        },
+        {
+            "name": "B-Bus",
+            "type": "http",
+            "url": "https://stops.lt/LVA/lva/gtfs.zip"
```
jbruechert: Why not use the mobility database here like before?
https://mobilitydatabase.org/feeds/gtfs/mdb-2385

Lach-anonym: For the time the license fetching from MDb is resolved (and given that with some miracle this feed will get a license), I think it's good to leave it in http form. But if it's not right, I can transform it and other feeds to MDb.

jbruechert: License urls are already copied from there, so it does have advantages already.

Lach-anonym: I see, I'll convert the other feeds later then

PR #1143, opened 2025-05-14
Feed update pack

Lach-anonym:
Mainly licenses were added, Ireland was overhauled and now we have regional buses around Cracow


feeds/ie.json
```diff
@@ -8,9 +8,20 @@
...
+            "url": "https://www.transportforireland.ie/transitData/Data/GTFS_All.zip",
+            "license": {
+                "spdx-identifier": "CC-BY-SA-4.0"
+            }
+        },
+        {
+            "name": "transport-for-ireland",
+            "type": "url",
+            "spec": "gtfs-rt",
+            "url": "https://www.transportforireland.ie/transitData/Data/GTFS_Realtime.zip",
```
jbruechert: GTFS-RT can never be a zip file, this is something else

Lach-anonym: I see now, this is the GTFS static file only for operators that have realtime data. The GTFS-RT can be obtained by API from here: https://developer.nationaltransport.ie/api-details#api=gtfsr&operation=TripUpdates

Lach-anonym:
I guess I'll leave it like that in case someone wants to try getting the GTFS-RT

feeds/ie.json
jbruechert: Yes, if you can get a token for it we can add it as well

jbruechert:
We don't need the static version two times then though.

jbruechert:
Thanks!

PR #1144, opened 2025-05-15
Add python client to libraries

1Maxnet1:


felixguendling:
This looks all like the old MOTIS API?

https://gitlab.com/motis-datagon/test-client-motis/-/blob/development/src/pymotis/handler/motis_trip_handler.py?ref_type=heads

I think we should only collect MOTISv2 compatible clients.

1Maxnet1:
> This looks all like the old MOTIS API?
> 
> https://gitlab.com/motis-datagon/test-client-motis/-/blob/development/src/pymotis/handler/motis_trip_handler.py?ref_type=heads
> 
> I think we should only collect MOTISv2 compatible clients.

No it's v1 and v2. As we did not migrate all servers at my workplace over to v2, I didn't remove the support for the old API yet.

1Maxnet1:
> I didn't find anything related to v2 and the docs also only explain v1: https://gitlab.com/motis-datagon/test-client-motis/-/tree/development/docs/source?ref_type=heads
> 
> I think this client might lead to frustration. Is there a link that might avoid confusion? At least I wasn't capable to find anything related to v2 fast. And I would expect the average curious Python programmer not expect that the "main" version of the client advertised on Transitous isn't compatible to Transitous at all.

I got confused between MOTIS version and API version. Will remove the client for now and might re-add it back, when I checked what's the issue exact issue is. Maybe the documentation confuses me as well, because https://transitous.org/api/ still claims v1. The python client does supports old API (when MOTIS was 0.x, but not the rewrite) and the new v1 API. The documentation is indeed not yet updated, because no one else except me used it yet.

felixguendling:
To clarify:

- MOTIS v0.x = legacy (not relevant anymore because Transitous doesn't run servers with 0.x)
- MOTIS v2.x
  - API v1: initial API version
  - API v2: polyline precision changed

I think this list should only contain clients for MOTIS >= v2.x or whatever software Transitous has deployed.


> because no one else except me used it yet

I guess the purpose of this list is to have others using your software. So even having no documentation would probably be better than a wrong/outdated documentation misleading people to waste time with code that will not work with the current Transitous server.

I think however, it would be great to have a Python client. But since we have an OpenAPI description, I would maybe try to generate one. If it works and generates good code, this would have the advantage that it's always up-to-date (compatibility + documentation).

Issue #1154, opened 2025-05-21
[KR] Add nationwide GTFS feed for South Korea


parkss404:
I would like to contribute a GTFS static feed that provides comprehensive **nationwide public transit coverage in South Korea.**
The dataset includes multiple transit such as subway, rail, bus, etc.

Feed Source
- **Filename**: GTFS_DataSet.zip
- **Hosted at**: [Dropbox Direct Download](https://www.dropbox.com/scl/fi/6crb68t52o18nrgn89ma6/kr.zip?rlkey=6y76uyq0x6p29l805dgiwjfg7&e=1&dl=1)
- **License**: https://www.ktdb.go.kr
- **Timezone**: Asia/Seoul

The feed has been successfully validated locally using fetch.py and gtfsclean.

jbruechert:
Hi, thanks for the link!

Since you seem to have done all the work already, would you mind opening a pull request?

PR #1155, opened 2025-05-21
Add GTFS feed for South Korea

parkss404:
[KR] Add nationwide GTFS feed for South Korea #1154

The feed was added to feeds/kr.json.


jbruechert:
Thanks!

PR #1163, opened 2025-05-24
Update cz.json IDSJMK realtime

hoermalmeister:
Add realtime gtfs and up to date timetables directly from operator

### DIFF
--- a/feeds/cz.json
+++ b/feeds/cz.json
@@ -7,6 +7,10 @@
         {
             "name": "Jonah Brüchert",
             "github": "jbruechert"
+        },
+        {
+            "name": "Petr Novák",
+            "github": "hoermalmeister"
         }
     ],
     "sources": [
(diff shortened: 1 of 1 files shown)


feeds/cz.json
```diff
@@ -32,8 +36,17 @@
...
-            "type": "transitland-atlas",
-            "transitland-atlas-id": "f-u2e-idsjmk"
+            "type": "url",
+            "spec": "gtfs-rt",
+            "url": "https://content.idsjmk.cz/kestazeni/gtfsReal.dat"
+        },
+        {
+            "name": "IDS-JMK",
+            "type": "http",
+            "url": "https://www.arcgis.com/sharing/rest/content/items/379d2e9a7907460c8ca7fda1f3e84328/data"
```
jbruechert: This is the same url that we were also getting from transitland. Using transitland is better, because we then don't need to deal with changes to the url.

jbruechert (changes requested):

feeds/cz.json
hoermalmeister: Sure, I thought it's not up to date there, because there are some diversions that are not accounted for in the feed, but it seems even the arcgis is not up to date yet. It seems they always update it at midnight on sunday, so I hope that will deal with it.

jbruechert:
Thanks!

PR #1165, opened 2025-05-25
Add Swiss service alerts feed

vkrause:


mlundblad:
Danke! :+1:

felixguendling:
Unfortunately they use different IDs than in the GTFS static data, so the matching rate is incredibly low. Hopefully it will be fixed in the future. I'm in touch with them about this issue.

PR #1166, opened 2025-05-25
add secondary instance + new API routes

traines-source:


ansible/hosts
```diff
@@ -19,6 +21,14 @@ motis:
           38316362353534633137313732383231313366616635666161316537643732663564383666363839
           63363534313561313533
 
+    goldberg:
+      ansible_ssh_user: algolocal
+      ansible_sudo_user: root
+      ansible_become: true
+      ansible_host: goldberg
+      motis_no_cow: true
+      motis_update_time: "4:00:00" # UTC
```
jbruechert: crunchy is currently configured with a Europe/Berlin timezone, so this would end up happening at the same time.

traines-source: Ah yes I forgot that winter exists :D
Also: Watch out for the permission changes I applied (again), because currently no motis group is created, but we can change that

jbruechert: Oh right, I probably made some questionable change for some previous setup at some point, and the motis group kept existing on the server.

PR #1170, opened 2025-05-26
Update France script

Lach-anonym:
I've checked the databases for errors written in the script file and removed the skips where I couldn't find the problems. I can't guarantee it will work flawlessly (especially the authentication for Rhôneexpress). Please check if some feeds still confuse MOTIS. I am afraid some navette feeds will expire very soon and they will be active again for winter


src/generate-france.py
```diff
@@ -331,15 +317,42 @@ def contains_name(out, name_to_check):
...
         {
             "name": "lyon-tcl",
             "type": "http",
             "url": "https://gtech-transit-prod.apigee.net/v1/google/gtfs/odbl/lyon_tcl.zip?apikey=BasyG6OFZXgXnzWdQLTwJFGcGmeOs204&secret=gNo6F5PhQpsGRBCK"
-        }
+        },
+        {
+            "name": "lyon-rhoneexpress",
+            "type": "http",
+            "url": "https://download.data.grandlyon.com/files/rdata/rx_rhonexpress.rxtheorique/GTFS_RX.ZIP?apikey=BasyG6OFZXgXnzWdQLTwJFGcGmeOs204&secret=gNo6F5PhQpsGRBCK"
```
jbruechert: This returns "Unauthorized"

Lach-anonym: Right, I hoped the same key would work... reverted

jbruechert:
hi @lzgirlcat! Thanks for creating all of these feeds!
Are you in some way related to https://cdn.zbiorkom.live/ or is that just another third-party service using your feeds?
I'm just asking in order to make sure the correct original URL can be shown on our sources page :)

lzgirlcat:
zbiorkom.live is created and managed by my friend, my feeds are used primarily by it.
The feeds hosted on cdn.zbiorkom.live are just copies of my feeds, cached for internal reasons.
Ideally the dataset should link directly to my feeds, because the zbiorkom.live cache only updates daily ~4:00 CEST.

lzgirlcat:
I also maintain an alternative feed for Białystok, bc the official one isn't maintained properly. It's not automatically updated, and in february the trip ids didn't match the ones from RT, breaking live tracking.

Lach-anonym:
Yesterday the Naolib data has disappeared and I thought it's the authority failure... but after looking into the script I've noticed the feed is skipped and thus the data was like uh, 3 months old or so? The outdated dataset is used, on the métropole website there is a new resource and for now I've made a simple fix for it

Lach-anonym:
Should the feeds with SSL error be skipped? Sadly I can't do that currently

traines-source:
Yeah it's a bit weird, can't reproduce these SSL errors locally and there are many other sites that use R11 Lets Encrypt certs, with the only difference maybe that these were reissued only a few days ago.
gtfsclean (more precisely gtfswriter in csvwriter.go:115) crashes on some other feeds because these contain txt files with free text that are interpreted as CSVs having more data columns than header columns when using `--keep-additional-fields`. Not sure if this should be fixed here, in gtfsclean, gtfsparser or gtfswriter

src/generate-france.py
```diff
@@ -122,17 +119,11 @@
...
-            "82630": "82309",
-        },
         "breizhgo-car": {
             "81804": "81463",
             "81805": "81466",
             "81806": "81461",
         },
-        "versions-des-horaires-theoriques-des-lignes-de-bus-et-de-metro-du-reseau-star-au-format-gtfs": {
-            "82161": "82951",
-        },
```
Altonss, line 135: Why having removed this logic? This is needed to keep GTFS - GTFS-RT association working (id's just need to be updated)

src/generate-france.py
```diff
@@ -122,17 +119,11 @@
 
     # Map for each dataset slug, if needed, the selected GTFS-RT id to the corresponding GTFS id
     gtfs_rt_select = {
-        "moca-communaute-de-communes-caux-austreberthe": {
-            "82630": "82309",
-        },
```
Altonss: Same question here? This was written to keep only one relevant gtfs-rt

src/generate-france.py
```diff
@@ -255,6 +246,10 @@
                 }
                 if dataset["slug"] in skip:
                     source["skip"] = True
+                expired = "metadata" in resource and "end_date" and "end_date" in resource["metadata"] and resource["metadata"]["end_date"] is not None and datetime.strptime(resource["metadata"]["end_date"], '%Y-%m-%d') < datetime.now()
+                if expired:
+                    print("Feed expired according to metadata, setting to skip=True:", resource["metadata"]["end_date"], feed_name)
+                    source["skip"] = True
```
Altonss, line 248: Is it necessary to skip expired feeds? Some feeds are "seasonal", and as such are always expired part of the year.

src/generate-france.py
traines-source: Ah yes thanks for catching that, I have fixed it, will push as soon as gtfsclean is updated

src/generate-france.py
traines-source: Yes so now at least they should be automatically unskipped again as soon as the script is run, right?

traines-source:
Ok so on my local import it looked like no major cities/networks were missing and RT matching also looked good. Only the feeds skipped due to SSL errors are a bit sad, still can't reproduce this locally. But I think they are not that important so from my perspective this is ready to merge.

src/generate-france.py
```diff
@@ -15,42 +16,36 @@
...
-        "arrets-horaires-et-circuit-de-la-lignes-yeu-continent-gtfs",  # agency.txt
-        "caen-la-mer-reseau-twisto-gtfs-siri",  # Temporary removal, 401 error
-        "reseau-de-transport-interurbain-mobigo-en-bourgogne-franche-comte",  # Temporary removal, resource not available
-        "gtfs-transport-horaires-des-lignes-de-la-communaute-de-communes-corse-du-sud-a-berlina",  # Temporary removal, 404 error
-        "gtfs-transport-horaires-des-lignes-de-la-communaute-dile-rousse-balagne-a-balanina",  # Temporary removal, 404 error
-        "navettes-aeroport-paris-beauvais-aerobus",  # Not GTFS format
-        "offre-de-transport-de-la-c-a-beaune-cote-sud-gtfs", # Missing and broken data
-        "gtfs-move-vendome",  # very low availability rate
-        "gtfs-static-et-real-time-transporteur-thalys" # Skip outdated Thalys data
+        "gtfs-static-et-real-time-transporteur-thalys", # Skip outdated Thalys data
```
Altonss:
```suggestion
```
This line can be removed I think, as the dataset has now been deleted.

feeds/fr.json
```diff
@@ -4554,7 +5085,7 @@
             }
         },
         {
-            "name": "horaires-theoriques-du-reseau-tgl",
+            "name": "horaires-theoriques-du-reseau-tgl--81899",
```
Altonss: Remove 81899 ressource, probably old data and leads to failed association to GTFS-RT

feeds/fr.json
```diff
@@ -4565,14 +5096,26 @@
             }
         },
         {
-            "name": "horaires-theoriques-du-reseau-tgl",
-            "type": "url",
-            "url": "https://proxy.transport.data.gouv.fr/resource/tgl-longwy-gtfs-rt-trip-update",
+            "name": "horaires-theoriques-du-reseau-tgl--83193",
```
Altonss: Remove 83193 resource, because it seems to be duplicate data of 83210, which leads to failed GTFS-RT association.

feeds/fr.json
```diff
@@ -6056,25 +6683,26 @@
             "spec": "gtfs-rt"
         },
         {
-            "name": "offre-transport-en-commun-du-reseau-transpor-gtfs",
+            "name": "offre-transport-en-commun-du-reseau-transpor-gtfs--82168",
```
Altonss: Remove 82168 ressource, seems to be duplicate/older data compared to 83260, and leads to failed gtfs-rt association.

Altonss:
> Only the feeds skipped due to SSL errors are a bit sad, still can't reproduce this locally.

Is it only an issue with the Github CI? Because in that case, I see no need to skip those feeds.

jbruechert:
> Is it only an issue with the Github CI? Because in that case, I see no need to skip those feeds.

If you can not reproduce the error in our CI container on your local machine either, then I think there is no need to skip these.

src/generate-france.py
```diff
@@ -122,17 +117,15 @@
 
     # Map for each dataset slug, if needed, the selected GTFS-RT id to the corresponding GTFS id
     gtfs_rt_select = {
-        "moca-communaute-de-communes-caux-austreberthe": {
-            "82630": "82309",
+         "versions-des-horaires-theoriques-des-lignes-de-bus-et-de-metro-du-reseau-star-au-format-gtfs": {
+            "82161": "83264",
+            "82162": "83264"
```
Altonss:
```suggestion
            "82161": "83263",
            "82162": "83263"
```
The current valid GTFS seems to be 83263 (see https://transport.data.gouv.fr/datasets/versions-des-horaires-theoriques-des-lignes-de-bus-et-de-metro-du-reseau-star-au-format-gtfs)

traines-source: ah yes and in two weeks we will have to change it again :/

Altonss: Yes sadly :( There is no ideal solution with this feed... :cry:

Altonss:
> If you can not reproduce the error in our CI container on your local machine either, then I think there is no need to skip these.

Haven't tried the CI container, but just downloading some of these GTFS files locally worked fine ;)

traines-source:
Yep it does work in my local CI container. Updated the propositions, SNCF indeed works with the single feed now, including RT.

jbruechert:
Works fine: https://routing-import.spline.de/repos/1/pipeline/943/4

Thanks!

PR #1178, opened 2025-05-28
Add MVG as Munich transport provider

mariunaise:
To improve the accuracy of itineraries in Munich, the MVG entry from the Mobility Database should be added. 

I am not quite sure on which license to use, the MVG website does not specify any information about the license of their timetable data.


jbruechert:
Closing due to no reply.

PR #1179, opened 2025-05-28
Add GTFS-Flex support

vkrause:
This is almost identical to static GTFS feeds, the only reason we need it to be a different type is that gtfsclean can't handle Flex feeds, so we need to skip that in post-processing.

Also, add the Swiss GTFS-Flex feed so we have something to test with.

### DIFF
--- a/src/fetch.py
+++ b/src/fetch.py
@@ -253,7 +253,7 @@ class Fetcher:
 
     # Returns whether something was downloaded
     def fetch_source(self, dest_path: Path, source: Source) -> bool:
-        if source.spec != "gtfs":
+        if source.spec != "gtfs" and source.spec != "gtfs-flex":
             return False
         match source:
             case TransitlandSource():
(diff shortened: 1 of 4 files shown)


src/fetch.py
```diff
@@ -337,6 +337,9 @@ def postprocess(self, source: Source,
         if source.fix_csv_quotes:
             subprocess.check_call(["./src/fix-csv-quotes.py", temp_file])
 
+        if source.spec == "gtfs-flex":
+            return
```
jbruechert: I'm not sure this is the best solution, as this also skips `os.rename(temp_file, output_path)` and the calendar validity check.

vkrause: Indeed, this only seemed to be working as I had files from a previous run still around that disabled this less aggressively.

jbruechert:
Btw have you tried whether `--keep-additional-fields` would work as well?
I'm pretty sure we will get into a situation where we need a gtfsclean feature + GTFS Flex.
The other option is of course to implement support for GTFS Flex in gtfsclean

vkrause:
> Btw have you tried whether `--keep-additional-fields` would work as well? I'm pretty sure we will get into a situation where we need a gtfsclean feature + GTFS Flex. The other option is of course to implement support for GTFS Flex in gtfsclean

I haven't found a way to make gtfsclean accept the feed and not destroy it in the process. I agree that longer term we'll likely need that, but for getting started with evaluating Flex support this is probably good enough (this is the feed Flex support in MOTIS has been developed against).

felixguendling:
I'm also not opposed to adding more robustness in the MOTIS parser to support/ignore invalid data.

PR #1182, opened 2025-05-29
Keep additional GTFS fields that gtfsclean doesn't know about

vkrause:
We need this e.g. for the `cars_allowed` field for ferries and car transport trains. It will also help with preserving Fares v2 related fields, although that needs more work to also preserve unknown files.

### DIFF
--- a/src/fetch.py
+++ b/src/fetch.py
@@ -344,6 +344,7 @@ class Fetcher:
                     "--check-null-coords",
                     "--empty-agency-url-repl", "https://transitous.org",
                     "--remove-red-services",
+                    "--keep-additional-fields",
                     "--output", str(temp_file)]
             if source.fix:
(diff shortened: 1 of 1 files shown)


jbruechert:
I'm a bit unsure if this alone is safe to merge without keeping unknown files as well.

vkrause:
https://github.com/public-transport/gtfsparser/pull/7 is the first part of that, the other two (write and gtfsclean) follow once this is in as they depend on the new API.

vkrause:
Sure, that's how I built/tested this locally, but I can't do PRs referencing not yet integrated dependencies :)

vkrause:
https://github.com/public-transport/gtfsclean/pull/11 is the final piece then, with that all unknown files are also preserved with `--keep-additional-fields`.

PR #1187, opened 2025-05-30
Poland further license updates and new feeds

Lach-anonym:
I've finally found the PKS Poznań feed link and also updated more licenses and added more feeds provided by @izgirlcat (thanks!)
(I swear at some point I'll tidy the file, so feeds are there alphabetically)

### DIFF
--- a/feeds/pl.json
+++ b/feeds/pl.json
@@ -260,8 +260,12 @@
         },
         {
             "name": "MPK-Łódź",
-            "type": "mobility-database",
-            "mdb-id": "tld-6079",
+            "type": "http",
+            "spec": "gtfs",
+            "url": "https://cdn.zbiorkom.live/gtfs/lodz.zip",
+            "license": {
+                "spdx-identifier": "MIT"
+            },
             "fix": true
         },
         {
@@ -275,7 +279,7 @@
             "spec": "gtfs",
             "url": "https://cdn.zbiorkom.live/gtfs/bialystok-wschod.zip",
             "license": {
-                "spdx-identifier": "MIT"
+                "spdx-identifier": "CC0-1.0"
             },
             "fix": true
         },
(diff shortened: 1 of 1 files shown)


Lach-anonym:
The error is... very random

jbruechert:
> I'm also wondering about still using Mobility Database here, because the license imported from there does not include or generate the SPDX identifier, which is quite crucial for other project I take part in

Then rather implement overriding the spdx-identifier for databases instead of workarounds. Using mobility database IDs makes it a lot easier to diff against it later ti find new feeds.

feeds/pl.json
```diff
@@ -260,8 +260,12 @@
         },
         {
             "name": "MPK-Łódź",
-            "type": "mobility-database",
-            "mdb-id": "tld-6079",
+            "type": "http",
+            "spec": "gtfs",
+            "url": "https://cdn.zbiorkom.live/gtfs/lodz.zip",
```
jbruechert, line 265: Is there actually something wrong with the previously used feed for Łódź?
What is the reason for the change?

feeds/pl.json
```diff
@@ -260,8 +260,12 @@
...
+            "license": {
+                "spdx-identifier": "MIT"
```
jbruechert, line 267: This is pretty surely wrong

feeds/pl.json
Lach-anonym: This feed actually has a license
I haven't found much for the official feed (although it's just made public)

feeds/pl.json
Lach-anonym: The source is here:

https://zbiorkom.live/tricity/blog/otwarte-dane-4VIz7Imv6dDIIy6j2e87Cs

jbruechert: Ah sorry, mixed it up with the feeds from girlc.at.

feeds/pl.json
jbruechert: I think in this case it makes sense, as the original server seems to be down.
However I am generally not happy with switching from official feeds to third party feeds.

Having all licenses properly documented is nice, but given many agencies just put their files somewhere without any conditions (and I'm pretty sure they have zero interest in suing people who use their files), I don't think it should be the primary concern.

PR #1190, opened 2025-06-01
Hungary feed update

Lach-anonym:
The source is https://napportal.kozut.hu/en/#/informations/toc and sites of the companies claiming it's just public info

### DIFF
--- a/feeds/hu.json
+++ b/feeds/hu.json
@@ -112,8 +112,9 @@
         },
         {
             "name": "Weekendbus",
-            "type": "http",
-            "url": "https://gtfs.menetbrand.com/download/weekendbus"
+            "type": "transitland-atlas",
+            "transitland-atlas-id": "f-u2mw-weekendbus",
+            "fix": true
         }
     ]
 }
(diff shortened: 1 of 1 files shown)


Lach-anonym:
The parser gets fooled because there is a date in feed version of Paks, so it thinks it's expired

jbruechert:
> The parser gets fooled because there is a date in feed version of Paks, so it thinks it's expired

The check uses the exact same fields that MOTIS uses as well, so if it says it's expired it will not have any effect. It does not use the feed version or url, just the calendar.txt, calendar_dates.txt and feed_info.txt.

jbruechert:
The latest date covered by the feed's calendar is 2025-05-30, so definitely in the past.

Lach-anonym:
> The latest date covered by the feed's calendar is 2025-05-30, so definitely in the past.

The feed_info.txt shows validity to 20250901, but indeed no trip can occur this way, very bizarre

So then I'll skip this feed until it gets an update

PR #1193, opened 2025-06-02
Update Romanian feed licenses

Lach-anonym:


feeds/ro.json
```diff
@@ -25,34 +25,52 @@
...
+                "spdx-identifier": "CC-BY-SA-4.0"
+            }
         },
         {
             "name": "Consiliul-Judetean-Vrancea",
             "type": "mobility-database",
-            "mdb-id": "mdb-1984"
+            "mdb-id": "mdb-1984",
+            "license": {
+            "spdx-identifier": "CC0-1.0"
```
jbruechert: formatting (you can use the bulk-edit-feeds.py script for automatic formatting. It's just a side effect but it works)

feeds/ro.json
```diff
@@ -25,34 +25,52 @@
...
+            "spdx-identifier": "ODbL-1.0"
+            }
         },
         {
             "name": "Bucuresti-Ilfov",
             "type": "mobility-database",
-            "mdb-id": "mdb-2098"
+            "mdb-id": "mdb-2098",
+            "license": {
+            "spdx-identifier": "CC0-1.0"
```
jbruechert: formatting

feeds/ro.json
```diff
@@ -25,34 +25,52 @@
         {
             "name": "Brasov",
             "type": "mobility-database",
-            "mdb-id": "mdb-2143"
+            "mdb-id": "mdb-2143",
+            "license": {
+            "spdx-identifier": "ODbL-1.0"
```
jbruechert: formatting

feeds/ro.json
Lach-anonym: fixed!

jonaes:
Hey @Lach-anonym, might the licence you added here for Iași, Cluj-Napoca and Chișinău also be applicable for the Feeds of Craiova (and [Pleven](https://mobilitydatabase.org/feeds/gtfs/mdb-2119) which isn't included here yet) which originate from external.gtfs.ro too?

jbruechert:
I added Pleven now, but I don't know about the license unfortunately.

Lach-anonym:
This has been so long time ago, I forgot how exactly I decided between CC-BY and CC0 (though I surely browsed through the sites for info), however I remember there is a government OGL license that is like one of the previous mentioned licenses

As Pleven is Bulgarian, I have no idea at this moment

jbruechert:
OGL is it's own license. spdx-identifiers should never be added here based on similar terms, but only if that is actually the license mentioned.

The only exception is CC-0 which is basically the international version of public domain (which in jurisdictions where it exists means no license).

PR #1206, opened 2025-06-10
added Calabria (italy)  data from mobility database

domewastaken:
Not much to say, however the gtfs lacks shapes.txt


jbruechert:
Thanks!

The lack of shapes.txt is unfortunately common. When there are no more important tasks left I would like to look into generating these automatically during import using pfaedle.

PR #1218, opened 2025-06-15
Update hr.json Rapska plovidba

hoermalmeister:
Added Rapska plovidba ferry feed


jbruechert:
Thanks! The error looks unrelated

PR #1221, opened 2025-06-15
Add GBFS for àVélo (Quebec City) and Accès Vélo (Saguenay)

wolfy1339:


wolfy1339:
Test failures are unrelated to the addition of this GBFS feed.

It's a problem with the STTR GTFS, of which there isn't an updated version available

jbruechert:
Thanks!

PR #1225, opened 2025-06-16
Create xk.json

hoermalmeister:
Kosovo railway operator

### DIFF
--- a/feeds/xk.json
+++ b/feeds/xk.json
@@ -1,0 +1,15 @@
+{
+    "maintainers": [
+        {
+            "name": "Petr Novák",
+            "github": "hoermalmeister"
+        }
+    ],
+    "sources": [
+        {
+            "name": "Trainkos",
+            "type": "http",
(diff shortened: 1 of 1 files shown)


jbruechert:
Where did you find this? I've been searching for such a feed for ages :)

hoermalmeister:
I wrote an email to trainkos whether the service in their timetables actually corresponds to reality and created the feed myself accordingly

jbruechert:
Ah, so it is updated manually?

hoermalmeister:
Indeed, unfortunately. I'm going to have to keep track of it. But with those few connections, it's not a huge burden

Lach-anonym:
On the other side there is Albania...

jbruechert:
> On the other side there is Albania...

There is a GTFS feed for Albania but it is unfortunately not public and nobody has been replying to me e-mails unfortunately.

Lach-anonym:
Oh indeed there is one for buses, but for trains there's nothing
In case of trains afair the only reliable enough timetable is the one at the station and it changes frequently

PR #1226, opened 2025-06-16
Ottawa's OCTranspo changed to direct feed

clintonthegeek:
I have amended the `ca-on.json` feed in two ways:
1) For the GTFS `.zip` file to be directly downloaded from OCTranspo's dedicated server.  to directly link to OC Transpo's official URL.
2) For the GTFS-RT data to be pulled from a proxy set up by JBB on the Transitous matrix channel, which obscures the secret API key necessary to download the data. The API key has been registered under my name and email address according to official processes and rules spelled out by [OCTranspo](https://www.octranspo.com/en/plan-your-trip/travel-tools/developers/dev-doc).

### DIFF
--- a/feeds/ca-on.json
+++ b/feeds/ca-on.json
@@ -231,14 +231,18 @@
         },
         {
             "name": "OCTranspo",
-            "type": "transitland-atlas",
-            "transitland-atlas-id": "f-f24-octranspo",
-            "fix": true
+            "type": "http",
+            "url": "https://oct-gtfs-emasagcnfmcgeham.z01.azurefd.net/public-access/GTFSExport.zip",
+            "http-options": {
+                "fetch-interval-days": 1
+            },
         },
         {
             "name": "OCTranspo",
-            "type": "transitland-atlas",
-            "transitland-atlas-id": "f-f24-octranspo~rt"
+            "type": "url",
+            "spec": "gtfs-rt",
+            "url": "https://jbb.ghsq.de/gtfs/octranspo/TripUpdates",
+
         },
         {
             "name": "PCConnect",
--- a/feeds/ca-on.json
+++ b/feeds/ca-on.json
@@ -4,6 +4,10 @@
             "name": "Marcus Lundblad",
             "github": "mlundblad"
         }
+        {
+            "name": "Clinton Ignatov (OC Transpo)",
+            "github": "clintonthegeek"
+        }
     ],
     "sources": [
         {
--- a/feeds/ca-on.json
+++ b/feeds/ca-on.json
@@ -3,7 +3,7 @@
         {
             "name": "Marcus Lundblad",
             "github": "mlundblad"
-        }
+        },
         {
             "name": "Clinton Ignatov (OC Transpo)",
             "github": "clintonthegeek"
--- a/feeds/ca-on.json
+++ b/feeds/ca-on.json
@@ -235,11 +235,9 @@
         },
         {
             "name": "OCTranspo",
-            "type": "http",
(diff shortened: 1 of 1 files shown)


feeds/ca-on.json
```diff
@@ -231,14 +235,17 @@
         },
         {
             "name": "OCTranspo",
-            "type": "transitland-atlas",
-            "transitland-atlas-id": "f-f24-octranspo",
-            "fix": true
+            "type": "http",
+            "url": "https://oct-gtfs-emasagcnfmcgeham.z01.azurefd.net/public-access/GTFSExport.zip",
+            "http-options": {
+                "fetch-interval-days": 1
```
jbruechert: The server seems to support the `Last-Modified` header to prevent fetching the file unnecessary often, is the manual limit really necessary?

feeds/ca-on.json
```diff
@@ -231,14 +235,17 @@
         },
         {
             "name": "OCTranspo",
-            "type": "transitland-atlas",
-            "transitland-atlas-id": "f-f24-octranspo",
-            "fix": true
+            "type": "http",
+            "url": "https://oct-gtfs-emasagcnfmcgeham.z01.azurefd.net/public-access/GTFSExport.zip",
```
jbruechert: The Transitland entry used before pointed to the same url already. Usually we prefer using the existing databases everywhere, as they make it easier to update the list of feeds from them without accidentally introducing duplicates. Is there a specific reason why you prefer the URL here?

clintonthegeek: I have not received realtime data from the Translitland entry. If it's better to keep it that way for the `GTFS` .zip file, then I understand.

If someone has in fact registered with OCTranspo and transitland is using a valid API key and fetching data, then  I'm just missing something on my end or if Transit Alerts remain unimplemented on either end, then yes my patch is redundant.

clintonthegeek: I should have checked this before. [Apparently there is an API key set up in the configuration file](https://github.com/transitland/transitland-atlas/blob/main/feeds/octranspo.com.dmfr.json), so if there is a registered key being privately stored and used by Transitious, then I have fundamentally misunderstood why I have don't receive any realtime data. Thanks for clarifying.

clintonthegeek: So apparently, using Transitous meta-data as is currently configured for GTFS-RT returns `GTFS-RT error (tag=ca-on-OCTranspo): unable to parse protobuf message: { "statusCode": 404, "message": "Resource not found" }`.

I have changed my pull request to keep the working proxy which transparently adds the API key, which should fix this issue, while reverting to transitous for the static .zip file. In the long term, some kind of infrastructure for handling private keys would be a longer-term solution.

jbruechert:
Thanks, it looks good now :)

feeds/ca-on.json
```diff
@@ -237,8 +241,9 @@
         },
         {
             "name": "OCTranspo",
-            "type": "transitland-atlas",
-            "transitland-atlas-id": "f-f24-octranspo~rt"
+            "type": "url",
+            "spec": "gtfs-rt",
+            "url": "https://jbb.ghsq.de/gtfs/octranspo/TripUpdates",
```
jbruechert:
```suggestion
            "url": "https://jbb.ghsq.de/gtfs/octranspo/TripUpdates"
```

feeds/ca-on.json
```diff
@@ -4,6 +4,10 @@
             "name": "Marcus Lundblad",
             "github": "mlundblad"
         }
```
jbruechert:
```suggestion
        },
```

PR #1233, opened 2025-06-22
Update lv.json

hoermalmeister:
Added official Vivi feed

### DIFF
--- a/feeds/lv.json
+++ b/feeds/lv.json
@@ -3,6 +3,10 @@
         {
             "name": "Jonah Brüchert",
             "github": "jbruechert"
+        },
+        {
+            "name": "Petr Novák",
+            "github": "hoermalmeister"
         }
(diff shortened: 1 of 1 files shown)


mlundblad:
Thanks!

jbruechert:
This is wrong, the data was already there in better quality and is probably now duplicated.

Lach-anonym:
Yet this feed doesn't have international trains!

jbruechert:
As far as I know vivi does not operate any international trains (except for Riga - Valga which is right at the border)

Lach-anonym:
Yes, but they cooperate with Elron and LTG on that three-country connection
I still don't know who is more to blame in this situation

jbruechert:
LTG Link does not publish any GTFS so the parts they operate are missing

Issue #1110, opened 2025-04-28
Liberty Island cruise issue


Lach-anonym:
https://github.com/user-attachments/assets/f2ad5f21-67ee-410a-aadf-d9545b17b92c
I've just noticed a boat flying over Uzbekistan

jbruechert:
I think wie might be lacking an unrealistic-speed filter for boats right now, usually that's the way we filter these kinds of things. That needs to be done in gtfsclean

Lach-anonym:
Also I can point out here the issue with city labeling at lower zooms

felixguendling:
> Also I can point out here the issue with city labeling at lower zooms

I'm not sure what you're referring to, but feel free to file a PR to fix this in MOTIS.

PR #1240, opened 2025-06-24
Fix boats flying over Uzbekistan

Lach-anonym:
#1110 fix

### DIFF
--- a/feeds/us-ny.json
+++ b/feeds/us-ny.json
@@ -156,8 +156,7 @@
         {
             "name": "Seastreak",
             "type": "transitland-atlas",
-            "transitland-atlas-id": "f-drk-seastreak",
-            "drop-shapes": true
+            "transitland-atlas-id": "f-drk-seastreak"
         },
         {
             "name": "NY-Waterway",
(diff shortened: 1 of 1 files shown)


feeds/us-ny.json
```diff
@@ -154,7 +156,8 @@
         {
             "name": "Seastreak",
             "type": "transitland-atlas",
-            "transitland-atlas-id": "f-drk-seastreak"
+            "transitland-atlas-id": "f-drk-seastreak",
+            "drop-shapes": true
```
jbruechert: Are we entirely sure this is the feed's fault and not a motis issue?

Lach-anonym: I haven't looked into it, just I wanted to pass the CI for now

jbruechert: This is unrelated to CI?

Also, we can of course merge PRs with failing CI, rather don't add unrelated changes.
The check is intended for new feeds, so it can't know that there is still a working cached download.

Lach-anonym: Alright then!

jbruechert:
@traines-source Could this be just a case of coordinates not being representable?

felixguendling:
image

Is this the ferry? It seems work as intended?

Lach-anonym:
It looks like in the meanwhile it got fixed (by feed maintainer?), so I'll close the PR

PR #1242, opened 2025-06-25
Add Transport for NSW

daviewales:


daviewales:
I've replaced the spaces in the source name with dashes.

daviewales:
I've enabled `"fix"`, and tested locally, and it seems to work now. (One stop was located at 0, 0.)

mlundblad:
Thanks!

PR #1244, opened 2025-06-26
Fix mc.json

hoermalmeister:
Repaired the link, motis didn't show the connections from the french server before anyway

### DIFF
--- a/feeds/mc.json
+++ b/feeds/mc.json
@@ -9,7 +9,7 @@
         {
             "name": "horaires-reseau-urbain-compagnie-des-autobus-de-monaco",
             "type": "http",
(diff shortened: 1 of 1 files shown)


jbruechert:
I had a look at the previous feed, and the problem there is that all coordinates are prefixed with a `+` character.
I would like to investigate whether that is something we should be able to handle or not first.

In general, I would like to keep the path data takes here somewhat understandable, so if you think the feed is so broken that fixing it is out of scope for Transitous, please at least document where the post-processing script is.

hoermalmeister:
Sure, I'm going to take at look at it when I'm home again. So is there a function that could handle the coordinates issue?

jbruechert:
I fixed this now by making the gtfsclean parser a bit more lenient.

PR #1245, opened 2025-06-26
Add AMAT Taranto feed

Lach-anonym:


Lach-anonym:
Funnily enough, this feed is used by Google and it also states it's expired

jbruechert:
The expiration check seems to be correct

Lach-anonym:
Again like in #1334, no feed_info.txt is present sadly

jbruechert:
This one only has calendar_dates.txt which makes it even more difficult as there is no service pattern to repeat.
It's not impossible to derive something from that, but the likelyhood that it makes sense becomes even lower.

Lach-anonym:
So it means Google presents incredibly low quality data

felixguendling:
> This one only has calendar_dates.txt

Doesn't gtfs-tidy have some kind of "compression" algorithm that derives weekly patterns from `calendar_dates.txt` to make the data more compact? Maybe gtfs-clean could also be used to add a `feed_info.txt` file?

jbruechert:
> > This one only has calendar_dates.txt
> 
> Doesn't gtfs-tidy have some kind of "compression" algorithm that derives weekly patterns from `calendar_dates.txt` to make the data more compact? Maybe gtfs-clean could also be used to add a `feed_info.txt` file?

Yes, it can do that. But the end date of each calendar pattern is currently the last service day of the pattern, which means the different services would not have a matching end date. That could be rounded up to the highest date that doesn't have a service.

Lach-anonym:
feed might be not expired anymore

Lach-anonym:
Looks like Taranto is not willing to change things, so I'll close this one

PR #1246, opened 2025-06-26
Add Pace (us-il) from Mobility Database

ethanc8:
Pace is a bus service operating in the Chicago suburbs. They provide a GTFS static feed, which can be downloaded from [their website](https://www.pacebus.com/route-timetable-data-services) and is included in [Mobility Database](https://mobilitydatabase.org/feeds/gtfs/mdb-2347) and [Transitland](https://mobilitydatabase.org/feeds/gtfs/mdb-2347).

### DIFF
--- a/feeds/us-il.json
+++ b/feeds/us-il.json
@@ -7,6 +7,10 @@
         {
             "name": "Kurt Kremitzki",
             "github": "kkremitzki"
+        },
+        {
+            "name": "Ethan C",
+            "github": "ethanc8"
         }
     ],
     "sources": [
@@ -21,6 +25,11 @@
             "http-options": {
                 "fetch-interval-days": 1
             }
+        },
+        {
+            "name": "Pace",
+            "type": "mobility-database",
+            "mdb-id": "mdb-2347"
         },
         {
             "name": "SMTD",


jbruechert:
I think gtfsclean is failing here because there is a non-csv .txt file in this feed. I'll try to find a solution

jbruechert:
We could add an option to ignore additional files I guess

mlundblad:
> We could add an option to ignore additional files I guess

Yeah, I guess the option, when enabled would ignore any non-required files.
Or, alternatively, if it could be made reliable, ignore any non-required file that "doesn't look like CSV"

jbruechert:
> Or, alternatively, if it could be made reliable, ignore any non-required file that "doesn't look like CSV"

I thought about it, but failed to come up with criteria to tell slightly malformed CSV and a text apart.
"having the same number of separators on every line" unfortunately also doesn't hold for some files we may want to read.

ethanc8:
Additionally, Pace has a GTFS-RT feed, but it needs to be cleaned, and the feed is undocumented. Catenary is currently using it, the URLs are on their transitland-atlas fork: https://github.com/catenarytransit/transitland-atlas/blob/a431b50e866e4d692014e36f3f8020d6ec49efee/feeds/pacebus.com.dmfr.json#L44

The issue that I know of is that the route IDs are different between the GTFS static and the GTFS-RT feeds; some of the routes have -367 appended to their IDs on the GTFS static but not on the GTFS-RT.

PR #1254, opened 2025-06-30
Create in-dl.json for adding Delhi Transit

Disaster2life:
Created an India Region, and added the DTC to it


mlundblad:
Other feeds in India has already been submitted in state-specific JSON files.

Maybe this should apply here as well?

I think in this case it would then be in-dl.json (for the NCT).

Lach-anonym (changes requested):
Greetings! Have you considered naming the file in-dl.json, according to https://en.m.wikipedia.org/wiki/ISO_3166-2:IN ?

Disaster2life:
Hi! yes indeed, `in-dl` makes much more sense, just searching for `in.json` hadnt yielded results so I named it that, I will make this change!

Disaster2life:
Hope this fits better with the current feeds and how hey are managed!

jbruechert:
Hi, looks good, please change the name to a string without spaces though :)

Disaster2life:
ooops-

Disaster2life:
So it seems the feed being expired seems to be throwing an error, if I understand correctly? not sure if that can be fixed?

jbruechert:
Hmm yes, unfortunately the Mobility Database website agrees on the feed being expired: https://mobilitydatabase.org/feeds/gtfs/mdb-1262

Lach-anonym:
Annoyingly, Google has the actual data apparently

jbruechert:
This website looks like it could have current data: https://otd.delhi.gov.in/data/static/

Disaster2life:
the OTD website doesn't, it says all the data was last updated on June 20th of 2024, and yes, Google, and the proprietary inhouse app they developed does have updated modern GTFS data, I believe they have mostly abandoned the OTD website in favour of a newer one (which I think is supposed to be more enterprise-y?) <https://delhi.transportstack.in/> I had avoided registering for it, but I guess if it gets us updated data?

Disaster2life:
Nope! updated later, the website claims it was updated last on the 2nd of December 2024, but their calendar.txt only puts the end of service at January of 2025, I guess I shall need to ask them very politely to update their GTFS data :/

Disaster2life:
I have filed an issue with them, hopefully we will have up to date data soon! the DTC (and DIMTS) introduced a new bus service recently and data about that has no access without using a proprietary source, I really wish to change that

jbruechert:
Let me know in case they update the data. I have already written some code to support post requests in our metadata, which seems to be necessary for that website.

Disaster2life:
Yeah if this is to be implemented, it would require sending over data to retrive data, I will prod and poke till I get an answer from the agency, and update this, better to turn this into a draf pull request till then?

jbruechert:
We could start by adding the DMRC feed which still seems to be valid

Disaster2life:
I was a bit concerned about using OTD instead of the transport stack website, but diffing the files, I have found that it is the same data even though both websites claim different last updated dates, ... and I am tied by the licence to inform both the providers to inform of the data being outdated and thus incorrect, so I shall be filing more tickets to both OTD and Transport Stack, which I think should allow us to use either if they do follow through and updated both.

But yes, both DMRC feeds are valid through to the end of year, so they have 5 months before they need to update it.

Would adding a "skip" and "skip-reason" to the DTC JSON be a good idea while we make the DMRC available till the data is updated?

jbruechert:
> Would adding a "skip" and "skip-reason" to the DTC JSON be a good idea while we make the DMRC available till the data is updated?

Sounds good!

jbruechert:
Thanks :)

Disaster2life:
ah it happened while I showered and did housekeeping work! thanks a lot for all this work! \o/ I will open a PR removing the skips as soon as I hear back from OTD and or Transit Stack!

PR #1257, opened 2025-07-02
Add Mińsk Mazowiecki feed

kasmar00:
Adds a new community maintained feed in Warsaw metro area. Thanks to @luigi22 for creating this feed.

### DIFF
--- a/feeds/pl.json
+++ b/feeds/pl.json
@@ -466,6 +466,15 @@
             }
         },
         {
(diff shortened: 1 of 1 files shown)


jbruechert:
Thanks!

Since our primary hosting location has problems right now, data can not be updated right now, so don't be surprised if nothing appears on the map tomorrow :)

Hopefully this will be fixed soon.

PR #1264, opened 2025-07-07
Update mc.json

hoermalmeister:
Just removing the + from the original stops.txt file


jbruechert:
This is already solved

PR #1265, opened 2025-07-07
[BG] Add Sofia feed

Dimitar5555:
The feed is based on the official data as provided on their website (https://sofiatraffic.bg/bg/public-transport) with minimal changes. I've added separate stops for each direction of the metro stations (they use a single code for the whole station) and I've used OpenStreetMap data for coordinates of stops (for metro, trams and trolleys).

I have a script that scrapes their website daily. The GTFS data is generated once per week.

Note that there are some data errors but it is outside of my control.


ignisf:
> Note that there are some data errors but it is outside of my control.

That is data consistency errors -- departures before arrivals for example -- which are sadly part of the official schedule.

jbruechert:
We probably need to improve our handling of unrealistic data in general to be able to make use of this feed. `"fix": true` will make it import it for now, but it will probably drop some stop times.
Maybe it makes more sense for now to add some heuristic on your side that works well for the specific data issues the timetable has, like removing times that are larger than the next time and setting `timepoint`  to `0` (approximate)

felixguendling:
> timepoint to 0 (approximate)

MOTIS doesn't distinguish this (and I'm also not sure how it should).

But MOTIS can deal with non-increasing stop times - no problem. It will just set each time to `max(this, previous)` to make sure there's no time travel.

Dimitar5555:
I dug a little deeper. It looks like their data is valid but for some routes some stops were in the wrong place. It is now fixed and the data should be error free.

Last manual validation of the data: https://gtfs-validator-results.mobilitydata.org/ae5c9896-9675-4689-b76a-e77f034bae4a/report.html

jbruechert:
> > timepoint to 0 (approximate)
> 
> MOTIS doesn't distinguish this (and I'm also not sure how it should).

From my reading of the spec, it is however necessary to set it if not providing a departure or arrival time.

> But MOTIS can deal with non-increasing stop times - no problem. It will just set each time to `max(this, previous)` to make sure there's no time travel.

gtfsclean however can not (yet)

jbruechert:
> I dug a little deeper. It looks like their data is valid but for some routes some stops were in the wrong place. It is now fixed and the data should be error free.

Thanks! There are still a few wrong stop positions most likely, at least that is the most likely cause of the too fast trips.
I think this is good to merge already though :)

felixguendling:
> timepoint to 0 (approximate)

MOTIS ignores this attribute (as it's not reliably set) but interpolates missing times anyway.


> gtfsclean however can not (yet)

Since MOTIS can deal with it already, I don't see any reason to put effort into adding this feature also to gtfsclean.

jbruechert:
> > timepoint to 0 (approximate)
> 
> MOTIS ignores this attribute (as it's not reliably set) but interpolates missing times anyway.
> 
> > gtfsclean however can not (yet)
> 
> Since MOTIS can deal with it already, I don't see any reason to put effort into adding this feature also to gtfsclean.

sure I guess but currently it will unconditionally error out which makes some kind of change necessary.

PR #1321, opened 2025-07-09
Shared Mobility feeds for Romania

FloreaCostinMario:
Added shared mobility feeds about half are provided by NextBike and the other half by another operator. The feeds by the later operator have generated [some errors on the mobility database ](https://storage.googleapis.com/mobilitydata-gbfs-snapshots-prod/gbfs-sibiu_bikecity/2.3/report_summary_20250709002718.json) however this shouldn't be an issues since it's caused by the way the operator defines the enums and the fix option should be able to fix this. (I am unable to test this since on my local machine I get the recent license errors).

There was also [another NextBike feed](https://mobilitydatabase.org/feeds/gbfs/gbfs-nextbike_rc) I wanted to add however for some reason it seems to have some elements in Germany? This is definitely not right so until I can investigate I'll hold it off.

### DIFF
--- a/feeds/ro.json
+++ b/feeds/ro.json
@@ -101,14 +101,12 @@
         {
             "name": "Dej-Velo-City",
             "type": "transitland-atlas",
-            "transitland-atlas-id": "f-dej~velo~city~gbfs",
-            "fix": true
+            "transitland-atlas-id": "f-dej~velo~city~gbfs"        
         },
         {
             "name": "Hunedoara-public-bike-system",
             "type": "transitland-atlas",
-            "transitland-atlas-id": "f-hunedoara~gbfs",
-            "fix": true
+            "transitland-atlas-id": "f-hunedoara~gbfs"
         },
         {
             "name": "Iași",
@@ -130,8 +128,7 @@
         {
             "name": "Moinesti-public-bike-system",
             "type": "transitland-atlas",
-            "transitland-atlas-id": "f-moinesti~gbfs",
-            "fix": true
+            "transitland-atlas-id": "f-moinesti~gbfs"
         },
         {
             "name": "Oradea",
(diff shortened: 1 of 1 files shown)


jbruechert:
Fix doesn't do anything for gbfs, so I'm not sure this will work.

The license error is probably just the missing license-expression python module.

feeds/ro.json
```diff
@@ -82,6 +98,18 @@
             "type": "mobility-database",
             "mdb-id": "mdb-2115"
         },
+        {
+            "name": "Dej-Velo-City",
+            "type": "transitland-atlas",
+            "transitland-atlas-id": "f-dej~velo~city~gbfs",
+            "fix": true
```
jbruechert: This is non-functional for gbfs

feeds/ro.json
```diff
@@ -82,6 +98,18 @@
...
+        },
+        {
+            "name": "Hunedoara-public-bike-system",
+            "type": "transitland-atlas",
+            "transitland-atlas-id": "f-hunedoara~gbfs",
+            "fix": true
```
jbruechert: same here

feeds/ro.json
```diff
@@ -109,11 +143,45 @@
             "type": "mobility-database",
             "mdb-id": "mdb-2108"
         },
+        {
+            "name": "Sântana-Velo",
+            "type": "transitland-atlas",
+            "transitland-atlas-id": "f-velo~sântana~gbfs",
+            "fix": true
```
jbruechert: same here

feeds/ro.json
```diff
@@ -137,6 +205,13 @@
             "type": "mobility-database",
             "mdb-id": "mdb-2107"
         },
+        {
+            "name": "Târgu-Mures-public-bike-system",
+            "type": "url",
+            "url": "https://targumures.publicbikesystem.net/customer/gbfs/v2/gbfs.json",
+            "spec": "gbfs",
+            "fix": true
```
jbruechert: same here

jbruechert (changes requested):

FloreaCostinMario:
Since the test import was successful I think it's safe to merge.

On another note, are gbfs feeds from the mobility database supported? Documentation only mentioned url or transitland.

jbruechert:
> On another note, are gbfs feeds from the mobility database supported? Documentation only mentioned url or transitland. 

Not yet, probably shouldn't be very hard to add. A few lines of python.

PR #1325, opened 2025-07-11
add Optima Express

jonaes:
adding unofficial Optima Express feed with timetable data scraped off their website.

The night train runs from Villach, AT to Edirne, TR, so I guess eu.json is most fitting.


jbruechert:
Thanks!

PR #1330, opened 2025-07-12
Improve how transitous deals with transitland feeds that require authorization

### LINKED ISSUE #1227: ci: Throw an error when encountering transitland entries that require authorization
- [ ] Throw an error when a transitland source is specified that requires authorization
- [ ] Add an abstract "api-key" property to transitland sources that allows to set a key, taking the header name from the metadata

krasch:
fixes #1227


jbruechert:
Thank you!

PR #1343, opened 2025-07-14
MOTIS config.yml: enable routed transfers + elevator updates

felixguendling:
Changes are only tested with the German DELFI dataset and not in production. So if we could deploy this on a staging instance (requires OSM) for some time, it would be better.

Changes:

- Enable osr_footpath which currently enables walking, wheelchair, and car transfers (bike could also be added in the same way): https://github.com/motis-project/motis/blob/566820bb099bac4e918336b4e349e3aef9ab098b/src/import.cc#L412-L430
- Enable elevator status update which updates wheelchair footpaths
- Not sure why `max_footpath_length` is set to 20min as this caused performance problems in big cities like Paris in the past. This PR changes it back to 15, which is also the default value.

### DIFF
--- a/motis/config.yml
+++ b/motis/config.yml
@@ -37,7 +37,7 @@ timetable:
   update_interval: 60
   http_timeout: 30
   incremental_rt_update: false
-  max_footpath_length: 15
+  max_footpath_length: 12
   extend_missing_footpaths: true
 # The datasets are filled in by ./src/generate-motis-config.py
   datasets:


felixguendling:
@traines-source should I add also `preprocess_max_matching_distance` here?

traines-source (commented):
I think we should currently not enable `preprocess_max_matching_distance`. Possibly as soon as we have separated the import pipelines of the two instances

traines-source (approved):

jbruechert:
Right now is probably not the ideal moment, but as soon as the TU Darmstadt backend server is back up, I can look into deploying this to the staging instance.

felixguendling:
I would bet that `max_footpath_length=20` significantly decreases performance. So if there's a longer delay, I would propose to merge things separately. Reducing this to `max_footpath_length=10` would probably bring even more speedup for dense city networks.

jbruechert:
Which part would you like to merge separately? Each of the changes?

felixguendling:
`timetable.max_footpath_length=12` would be nice to have rather sooner than later for performance reasons. But if the setup is done soon, there's no need.

traines-source:
I've added the `max_footpath_length` over in #1368, we can keep this PR here open until we've figured out why `osr_footpath` preprocessing takes so long on the planet.

traines-source:
Ok `osr_footpath` prolongs the timetable import by about 15-30 minutes. So I think we can enable it by merging this PR.

PR #1345, opened 2025-07-14
Add usage policy

vkrause:
As discussed during the hack weekend and thus especially looking for feedback from those of you who couldn't attend.

### DIFF
--- a/website/content/_index.html
+++ b/website/content/_index.html
@@ -27,6 +27,10 @@ routing service that does not stop at borders.</p>
 
 <h2>Apps That Use Transitous</h2>
 
+<!-- Rules for adding entries here:
+ - applications must be Free Software/Open Source
+ - non-application uses shown here must be non-profit
+-->
 <div class="mt-4 app-card-row">
 {{< app-card name="KDE Itinerary" website="https://apps.kde.org/itinerary/" icon="images/apps/kde-itinerary.svg" >}}
 {{< app-card name="KTrip" website="https://apps.kde.org/ktrip/" icon="images/apps/ktrip.svg" >}}
--- a/website/content/api.html
+++ b/website/content/api.html
@@ -12,7 +12,46 @@ SPDX-License-Identifier: AGPL-3.0-or-later
 -->
 
 The API provided by Transitous can be used for free, but we recommend that you let us know if you use it, so your use case can be considered in future decisions.
-You can do that by <a href="https://github.com/public-transport/transitous/blob/main/website/content/_index.html#L30">adding your app to our website</a> and/or letting us know in our <a href="https://matrix.to/#/%23transitous:matrix.spline.de">Matrix channel</a>.
+You can do that by <a href="https://github.com/public-transport/transitous/blob/b2d184b51cb4cf20fb48d0b76420ba57f4d3fce0/website/content/_index.html#L30">adding your app to our website</a> and/or letting us know in our <a href="https://matrix.to/#/%23transitous:matrix.spline.de">Matrix channel</a>.
+
+<h2>Usage Policy</h2>
+
+<p>Transitous is provided by a community of volunteers to fulfill the needs of Free and Open Source applications and other non-profit activities.
+We provide the service on a best-effort basis and reserve the right to stop providing it to any consumer at any time.</p>
+
+<h3>Resource Usage</h3>
+
+<p>Our hosting resources are limited.
+Please contact us before doing any potentially resource-intensive operations, such as many requests (also from different users) or primarily difficult to calculate requests (such as routing, isochrones).
+Generally, if your requests take long, they are difficult to calculate.</p>
+
+<p>We'll try to find ways to accommodate your use. Should we hit the limits of what we can handle with our current resources,
+there's also the option to <a href="/doc/#running-a-transitous-instance-locally">self-host your own instance of Transitous</a>.</p>
+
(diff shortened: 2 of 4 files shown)


website/content/_index.html
```diff
@@ -27,6 +27,10 @@
 
 <h2>Apps That Use Transitous</h2>
 
+<!-- Rules for adding entries here:
+ - applications must be Free Software/Open Source
+ - non-application uses shown here must be non-profit
+-->
```
Altonss, line 33: This means `Transportflow (iOS)` would have to be removed from the entries?

website/content/api.html
```diff
@@ -14,6 +14,45 @@
...
+
+<p>There is no need to scrape data using the API, you may download our entire source dataset <a href="https://github.com/public-transport/transitous/wiki/Using-Transitous-data">here</a>.
+
+<h3>API Usage</h3>
+
+<p>If you want to use the API for your free and open-source application, please:<p>
+
+<ul>
+<li>Contact us <a href="https://matrix.to/#/%23transitous:matrix.spline.de">in our Matrix Channel</a> to make sure we can handle your
+expected resource usage and so we are aware your usecases.</li>
```
derhuerst:
```suggestion
expected resource usage and so we are aware of your usecases. This also allows us to contact you in case we want to coordinate breaking changes.</li>
```

website/content/api.html
```diff
@@ -14,6 +14,45 @@
...
+<li>Make sure your send HTTP <code>User-Agent</code> or <code>X-Client-Identification</code> headers with every request, at least containing the following information:</li>
+<ul>
+    <li>Name of the application</li>
+    <li>Version of the client implementation</li>
+    <li>Contact e-mail address</li>
```
derhuerst: Do we have a decision on collecting [PII](https://en.wikipedia.org/wiki/Personal_data) in access logs? We could also ask people to provide a URL to a website of the app.

website/content/api.html
```diff
@@ -14,6 +14,45 @@
...
+</ul>
+<li>You may <a href="https://github.com/public-transport/transitous/blob/main/website/content/_index.html#L30">add your application to the Transitous website</a></li>
```
derhuerst: Using a permalink (pointing to a specific commit) here has the disadvantage that it will be outdated soon, but at least it doesn't confuse users by breaking (or pointing to an entirely different section of the file).

```suggestion
<li>You may <a href="https://github.com/public-transport/transitous/blob/b2d184b51cb4cf20fb48d0b76420ba57f4d3fce0/website/content/_index.html#L30">add your application to the Transitous website</a></li>
```

website/content/suppliers.html
```diff
@@ -0,0 +1,29 @@
...
+
+<!--
+SPDX-FileCopyrightText: 2025 Volker Krause <vkrause@kde.org>
+
+SPDX-License-Identifier: AGPL-3.0-or-later
+-->
+
+If you need support around public transport routing for a commercial usecase or
+are looking for hosted MOTIS instances, the following service providers might be
+of interest to you. Everyone listed here has substantially contributed to Transitous.
```
derhuerst, line 15: We might want to explain publicly *why* we want these "trusted suppliers" to contribute substantially.

```suggestion
of interest to you.
Because contributions to the voluntary Transitous project by for-profit entities is important to ensure Transitous' long-term sustainability, only providers that have contributed substantially will be listed here.
```

website/content/api.html
jbruechert: Makes sense, a URL is fine as well.

website/content/suppliers.html
jbruechert: I prefer the previous version. On this page, people probably primarily want to know why we trust these suppliers, and the reason is that we know they are capable through their contributions, right?

website/content/_index.html
vkrause: That is a bit of a (dark-ish) gray area indeed. We could at least argue that there's no for-profit intent AFAIU, but the app not being open source kinda goes against the idea here.

And for context, most of this was motivated by much more clearly commercial/for-profit requests we have gotten recently, Transportflow on its own isn't really a problem.

website/content/api.html
vkrause: A generic vendor contact email address isn't really PII, but I agree generalizing this to also allow a URL makes sense.

website/content/api.html
vkrause: Preexisting problem, I'll apply your fix to all occurrences.

website/content/suppliers.html
vkrause: The main motivation for this criteria was that it is the most effective and halfway objective filter we could think of to ensure the right companies/people end up on this list.

It's stated on the page mainly as a way to explain how we got to know them and their capabilities.

website/content/_index.html
jbruechert: @Altonss Is this fine with you so that the comment can be resolved, or do you have concerns left?

website/content/suppliers.html
jbruechert: @derhuerst Are you okay with leaving the current text, so that we can move forward with this?

website/content/_index.html
Altonss: No I don't have any more concerns. Just noticed the slight incoherence between the new rule and existing listing (maybe the rule could be updated to better reflect the reality).

felixguendling:
Is it ready to merge? I get asked questions about this topic and it would be nicer to point to a website than to a PR. This version is way better than nothing IMO. We can iterate later with PRs if someone wants to change something.

PR #1348, opened 2025-07-15
Use proper argument parsing in fetch.py

vkrause:
Provides better error messages and a somewhat useful help output.


jbruechert (approved):

src/fetch.py
```diff
@@ -29,6 +30,11 @@
 import csv
 
 
+parser = argparse.ArgumentParser(description='Transitous GTFS feed fetcher and post-processor.')
```
jbruechert: Please move this into the main "function"

PR #1350, opened 2025-07-16
Use proper command line argument parsing in generate-motis-config.py

vkrause:
Gives us better error handling and help messages.

For further improvements here (like defaulting to full mode) we will need to move away from using consecutive positional arguments though, which will require breaking changes. So that's a separate change.


jbruechert:
Thanks! Breaking changes are fine, the documentation and the CI are the only things that need to be changed

PR #1351, opened 2025-07-16
Update feeds from the Slovenian NAP

jbruechert:
They added a few new ones in the meantime that I requested access for. They also added proper license documentation that I copied here.


jbruechert:
> Typo ;) Edit: line 29, somehow it didn't add it here

Thanks, fixed

PR #1361, opened 2025-07-19
Add a section in the documentation about the used data types

vkrause:
Taken from https://volkerkrause.eu/2025/06/14/transitous-adding-data.html.

The information about diagnostics from there probably also makes sense here, but I'd put this into its own section below.


felixguendling (approved):

jbruechert:
Thanks!

PR #1362, opened 2025-07-19
Update hr.json

hoermalmeister:
added multiple city busses


jbruechert:
Are these created by hand?

hoermalmeister:
Sveta Nedelja is completely by hand, because I couldn't find any other solution, but with the others I mostly import the times directly from arriva's timetable and create the shapes and stop coordinates manually

PR #1365, opened 2025-07-20
Automatically set paths to MOTIS assets when MOTIS is found in PATH

vkrause:
This only works for the prebuilt MOTIS binaries so far, not for locally self-compiled versions as those have a different relative directory layout.


vkrause:
https://github.com/motis-project/motis/pull/945 would also make this work for local builds.

jbruechert (approved):

Issue #756, opened 2025-01-05
[website] In sources, move subdivisions into country sections (Canada and USA)


Altonss:
Moving the subdivisions sources into each country would make the list more readable.

cahfofpai:
I started working on this. Writing this as a comment, because I can not assign myself to this issue.

jbruechert:
The format of license.json doesn't need to stay the same, so feel free to modify it if it makes things easier / cleaner.

PR #1370, opened 2025-07-22
Generate two-level source region information

vkrause:
That's the first step to show this hierarchically on the source page.

For now this keeps compatibility with the old format as well, so this can be done in stages (I don't have the full dataset locally for generating this entirely to properly test the website part in one go).

See #756.

### DIFF
--- a/src/generate-attribution.py
+++ b/src/generate-attribution.py
@@ -34,8 +34,8 @@ def filter_duplicates(elems):
     return out
 
 
-def http_source_attribution(source: HttpSource) -> Optional[dict]:
-    attribution: dict = {}
+def http_source_attribution(source: HttpSource, region_data: dict) -> Optional[dict]:
+    attribution = region_data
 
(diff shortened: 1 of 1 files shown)


jbruechert:
Wouldn't it make more sense to add a new field for `country_name` and `subdivision_name` and keep everything else as it is?

vkrause:
> Wouldn't it make more sense to add a new field for `country_name` and `subdivision_name` and keep everything else as it is?

Isn't that essentially what this does (besides the differently named fields, and reshuffling of the code a bit to not have functions access local variables in main)?

jbruechert:
Pretty much, but I think it would make sense to a) name the new properties country and subdivision, and to separate our REGIONS fallback into country and subdivision as well.

vkrause:
That's cleaner indeed, done.

jbruechert:
Looks good now, thanks!

Issue #1366, opened 2025-07-20
Multiple stops with the same name in the Netherlands


sandervhoogdalem:
For example: KNMI, De Bilt. Has 2 different stopareas (two different operators) but on DRGL.nl and in OVinfo (OVApi App) they show them as one, not as 2 seperate stops. Is there a way to get both of them as one result?

https://api.transitous.org/api/v1/geocode?text=KNMI results in:

```
  {
    "type": "STOP",
    "tokens": [
      [0, 4]
    ],
    "name": "De Bilt, KNMI",
    "id": "nl-ovapi_stoparea:448940",
    "lat": 0.2161949,
    "lon": 0.0215046,
    "level": 0,
    "areas": [],
    "score": -3.98749995231628
  },
  {
    "type": "STOP",
    "tokens": [
      [0, 4]
    ],
    "name": "De Bilt, KNMI",
    "id": "nl-ovapi_stoparea:446327",
    "lat": 0.0825067,
    "lon": 0.0082069,
    "level": 0,
    "areas": [],
    "score": -3.98749995231628
  },
```

felixguendling:
They are very far from each other:
https://opencagedata.com/tools/distance-between-two-points#0.2161949,0.0215046,0.0825067,0.0082069

If they would be less than 500m from each other, they should become one entry.

sandervhoogdalem:
I am not sure what the problem is with the locations but I see normal numbers in de GTFS records that don't have a stoparea:

2694127,50100060,"De Bilt, KNMI",52.103018,5.18288,0,stoparea:446327,,0,,
2694128,50100070,"De Bilt, KNMI",52.102918,5.182443,0,stoparea:446327,,1,,
3161835,50100060,"De Bilt, KNMI",52.103063,5.182877,0,stoparea:448940,,0,,
3161836,50100070,"De Bilt, KNMI",52.102885,5.182323,0,stoparea:448940,,1,,

Only the stoparea gives another format: 
stoparea:446327, KNMI,"De Bilt, KNMI",0.08250667933491687,0.00820690657165479,1,,,0,,
stoparea:448940,,"De Bilt, KNMI",0.21619491286307055,0.0215045643153527,1,,,0,,

The distance between 2694127 and 3161835 is only 0.01 KM... (https://opencagedata.com/tools/distance-between-two-points#52.103018,5.18288,52.103063,5.182877)

[...]

Issue #1347, opened 2025-07-15
Stoparea mistakes in the netherlands


sandervhoogdalem:
I am using this API for a while and I found something that is very strange.

When I search for the stoptimes of "Dieverbrug, Busstation" I get departures from the API that are not even close from the busstop. In OVApi, multiple stops are bundled as stoparea:id, for this one stoparea:449779. When I look in the GTFS data, I have the following matching stopIds:

3188834, 3188835, 3188836, 3199173, 3199174 en 3199175

But when I look into the stoptimes (https://api.transitous.org/api/v1/stoptimes?stopId=nl-ovapi_stoparea:449779&n=30&radius=200?time=2025-07-21T07:24:00Z) with a lot of different stopIds. For example:

`{
      "place": {
        "name": "Groningen, Jan Luykenstraat",
        "stopId": "nl-ovapi_3196724",
        "lat": 53.1974,
        "lon": 6.568679,
        "level": 0,
        "arrival": "2025-07-15T07:26:00Z",
        "departure": "2025-07-15T07:26:00Z",
        "scheduledArrival": "2025-07-15T07:24:00Z",
        "scheduledDeparture": "2025-07-15T07:24:00Z",
        "vertexType": "TRANSIT"
      },
      "mode": "BUS",
      "realTime": true,
      "headsign": "Eelde",
      "agencyId": "QBUZZ",
      "agencyName": "Qbuzz",
      "agencyUrl": "http://www.qbuzz.nl",
      "routeColor": "ff00ff",
      "tripId": "20250715_09:19_nl-ovapi_299379385",
      "routeShortName": "9",
      "pickupDropoffType": "NORMAL",
      "cancelled": false,
      "source": "nl_ovapi.gtfs.zip/stop_times.txt:12503208:12503236"
    }`

This one is a line on the same contract (concession) but I also have results in other area's of the country:

`{
      "place": {
        "name": "Apeldoorn, Hoofdstraat",
        "stopId": "nl-ovapi_2747802",
        "lat": 52.21208,
        "lon": 5.963825,
        "level": 0,
        "arrival": "2025-07-15T07:26:00Z",
        "departure": "2025-07-15T07:26:00Z",
        "scheduledArrival": "2025-07-15T07:26:00Z",
        "scheduledDeparture": "2025-07-15T07:26:00Z",
        "vertexType": "TRANSIT"
      },
      "mode": "BUS",
      "realTime": true,
      "headsign": "Apeldoorn Station",
[...]

PR #1376, opened 2025-07-25
Update nl.json to solve #1347 and #1366

sandervhoogdalem:
Change the datasource from OVApi to OpenOV to solve issue #1347 and #1366.

With this change, the stoparea's are not longer available in the dataset. 

Issue #1366 will be solved because without the stoparea's, every stop with the same name will return their departures instead of just the one that is selected. 

Issue #1347 will also be solved because without the stoparea's, every stop will only return their departures and not the departures of locations on the other side of the country.

### DIFF
--- a/feeds/nl.json
+++ b/feeds/nl.json
@@ -18,8 +18,7 @@
             "license": {
                 "spdx-identifier": "CC0 1.0 Universal",
                 "url": "https://gtfs.openov.nl/LICENSE.TXT"
-            },
-        
+            }
         },
         {
             "name": "OpenOV",
@@ -59,6 +58,6 @@
             "type": "url",
             "spec": "gtfs-rt",
             "url": "https://gtfs.ovapi.nl/nl/trainUpdates.pb"
-        },
+        }
     ]
 }
--- a/feeds/nl.json
+++ b/feeds/nl.json
@@ -16,7 +16,7 @@
             "url" : "https://gtfs.openov.nl/gtfs-rt/gtfs-openov-nl.zip",
             "fix": true,
             "license": {
-                "spdx-identifier": "CC0 1.0 Universal",
+                "spdx-identifier": "CC0-1.0",
                 "url": "https://gtfs.openov.nl/LICENSE.TXT"
             }
         },
--- a/feeds/nl.json
+++ b/feeds/nl.json
@@ -44,6 +44,21 @@
             "type": "url",
             "spec": "gtfs-rt",
             "url": "https://gtfs.openov.nl/gtfs-rt/vehiclePositions.pb"
-        }
+        },
+        {
+            "name": "ovapi",
+            "type": "transitland-atlas",
+            "transitland-atlas-id": "f-u-nl",
+            "fix": true,
+            "keep-agency-names": [
+                "Eu Sleeper"
+            ]
+        },
+        {
+            "name": "ovapi",
+            "type": "url",
+            "spec": "gtfs-rt",
+            "url": "https://gtfs.ovapi.nl/nl/trainUpdates.pb"
+        },
     ]
 }
--- a/feeds/nl.json
+++ b/feeds/nl.json
@@ -3,32 +3,47 @@
         {
             "name": "Jonah Brüchert",
             "github": "jbruechert"
+        },
+        {
+            "name": "Sander van Hoogdalem",
+            "github": "sandervhoogdalem"
         }
     ],
     "sources": [
         {
-            "name": "ovapi",
-            "type": "transitland-atlas",
-            "transitland-atlas-id": "f-u-nl",
-            "fix": true
+            "name": "OpenOV",
+            "type": "http",
+            "url" : "https://gtfs.openov.nl/gtfs-rt/gtfs-openov-nl.zip",
+            "fix": true,
+            "license": {
+                "spdx-identifier": "CC0 1.0 Universal",
+                "url": "https://gtfs.openov.nl/LICENSE.TXT"
+            },
+        
         },
         {
-            "name": "ovapi",
+            "name": "OpenOV",
             "type": "url",
             "spec": "gtfs-rt",
-            "url": "https://gtfs.ovapi.nl/nl/trainUpdates.pb"
+            "url": "https://gtfs.openov.nl/gtfs-rt/trainUpdates.pb"
         },
         {
-            "name": "ovapi",
+            "name": "OpenOV",
             "type": "url",
             "spec": "gtfs-rt",
-            "url": "https://gtfs.ovapi.nl/nl/tripUpdates.pb"
+            "url": "https://gtfs.openov.nl/gtfs-rt/tripUpdates.pb"
         },
         {
-            "name": "ovapi",
+            "name": "OpenOV",
             "type": "url",
             "spec": "gtfs-rt",
-            "url": "https://gtfs.ovapi.nl/nl/alerts.pb"
+            "url": "https://gtfs.openov.nl/gtfs-rt/alerts.pb"
+        },
+        {
+            "name": "OpenOV",
+            "type": "url",
+            "spec": "gtfs-rt",
+            "url": "https://gtfs.openov.nl/gtfs-rt/vehiclePositions.pb"
         }
     ]
 }
--- a/src/fetch.py
+++ b/src/fetch.py
@@ -376,6 +376,11 @@ class Fetcher:
                 for agency in source.drop_agency_names:
                     command.append("--drop-agency-names")
                     command.append(agency)
+            if source.keep_agency_names:
+                for agency in source.keep_agency_names:
+                    command.append("--keep-agency-names")
+                    command.append(agency)
+                command.append("--delete-orphans")
             if source.display_name_options:
                 if source.display_name_options.copy_trip_names_matching:
                     command.append("--copy-trip-names-matching")
--- a/src/metadata.py
+++ b/src/metadata.py
@@ -51,6 +51,7 @@ class Source:
     drop_too_fast_trips: bool = True
     drop_shapes: bool = False
     drop_agency_names: List[str] = []
+    keep_agency_names: List[str] = []
     display_name_options: Optional[DisplayNameOptions] = None
 
     def __init__(self, parsed: Optional[dict] = None):
@@ -81,6 +82,8 @@ class Source:
                 self.drop_shapes = parsed["drop-shapes"]
             if "drop-agency-names" in parsed:
                 self.drop_agency_names = parsed["drop-agency-names"]
+            if "keep-agency-names" in parsed:
+                self.keep_agency_names = parsed["keep-agency-names"]
             if "display-name-options" in parsed:
                 self.display_name_options = \
                     DisplayNameOptions(parsed["display-name-options"])


jbruechert:
Hmm the set of agencies doesn't match exactly, for example we'd be losing European Sleeper this way.

sandervhoogdalem:
> Hmm the set of agencies doesn't match exactly, for example we'd be losing European Sleeper this way.

Is that the only one?

jbruechert:
> > Hmm the set of agencies doesn't match exactly, for example we'd be losing European Sleeper this way.
> 
> Is that the only one?

I haven't done an exact comparison but it looks like a larger difference than just one. Compare the agency.txt of the two feeds for details.

sandervhoogdalem:
> > > Hmm the set of agencies doesn't match exactly, for example we'd be losing European Sleeper this way.
> > 
> > 
> > Is that the only one?
> 
> I haven't done an exact comparison but it looks like a larger difference than just one. Compare the agency.txt of the two feeds for details.

I see 38 lines in the agency of OpenOV and 40 in OVApi. Only European Sleeper and Deutsche Bahn are missing, but Deutsche bahn is merged with NS International because all trains of DB in NL are operating under the flag NS International. So it's only European Sleeper. It's weird because European Sleeper is on the list of one of the sources of this GTFS files.

image

jbruechert:
Is there any way to contact the feed producer?

sandervhoogdalem:
> Is there any way to contact the feed producer?

I am trying to find a e-mail address or something at the moment

felixguendling:
Could we mask the agencies and take only the sleepers from one and the rest from the other?

sandervhoogdalem:
> Could we mask the agencies and take only the sleepers from one and the rest from the other?

Is that even possible? Doesn't that need extra parameters or something?

jbruechert:
> > Could we mask the agencies and take only the sleepers from one and the rest from the other?
> 
> Is that even possible? Doesn't that need extra parameters or something?

We already have "remove-agency-names" which would allow to do this if we do not find a better option, yes.

sandervhoogdalem:
> We already have "remove-agency-names" which would allow to do this if we do not find a better option, yes.

It's impossible to find a contact for the set but I placed a support request on the discord server. Is it a option to use the "remove-agency-names" until there is a solution?

With European Sleeper, we talk about 3 trains in both directions every week

sandervhoogdalem:
I have an answer about why European Sleeper is missing. Its called interoperability. 

Apple and Google Maps are using the dataset of OpenOV thats why European Sleeper is not in the OpenOV set.

felixguendling:
> Its called interoperability.

Not sure why removing a dataset makes it interoperable?
Do Apple and Google have the sleeper trains from another dataset?
Can we also access this dataset and add them separately?

sandervhoogdalem:
> Not sure why removing a dataset makes it interoperable? Do Apple and Google have the sleeper trains from another dataset? 

I don't know either but this is the answer that I've got. 


> Can we also access this dataset and add them separately?

I asked for the information about this

jbruechert:
> Can we also access this dataset and add them separately?

European Sleeper has so far ghosted me completely when I asked about this.

sandervhoogdalem:
> > Can we also access this dataset and add them separately?
> 
> European Sleeper has so far ghosted me completely when I asked about this.

Well I found something:

https://www.distribusion.com/news/distribusion-technologies-partners-with-european-sleeper-to-increase-global-online-sales

Maybe we can find a GTFS File there?

jbruechert:
> With European Sleeper, we talk about 3 trains in both directions every week

Yes, but trains people are actually using Transitous with, and we'd be getting complaints from the Träwelling crowds if we removed them.

sandervhoogdalem:
> Yes, but trains people are actually using Transitous with, and we'd be getting complaints from the Träwelling crowds if we removed them.

Makes sense. I am not able to find another dataset with european sleeper in it. I guess we have to use the flag for only european sleeper

sandervhoogdalem:
> > With European Sleeper, we talk about 3 trains in both directions every week
> 
> Yes, but trains people are actually using Transitous with, and we'd be getting complaints from the Träwelling crowds if we removed them.

They can fill it in manually if they want the train in their list too

jbruechert:
> > > Can we also access this dataset and add them separately?
> > 
> > European Sleeper has so far ghosted me completely when I asked about this.
> 
> Well I found something:
> 
> https://www.distribusion.com/news/distribusion-technologies-partners-with-european-sleeper-to-increase-global-online-sales
> 
> Maybe we can find a GTFS File there?

That might explain why nobody replied, probably they never see the feed and it goes directly to google. We probably need to ask distribusion directly.

sandervhoogdalem:
> > Well I found something:
> > https://www.distribusion.com/news/distribusion-technologies-partners-with-european-sleeper-to-increase-global-online-sales
> > Maybe we can find a GTFS File there?
> 
> That might explain why nobody replied, probably they never see the feed and it goes directly to google. We probably need to ask distribusion directly.

Their API is only for retailers and has a price so I guess this option won't be something for Transitous...

sandervhoogdalem:
I think the best we can do is add OVApi only for European Sleeper. But what are we going to do when the list of operators change? When we have to check everyday it will be impossible to maintain. Is there a flag for loading only one operator from a GTFS dataset @jbruechert ?

jbruechert:
Not really. In the worst case we end up with duplicated trips, which MOTIS can hopefully merge by itself given both datasets are based on the same source data.

felixguendling:
> In the worst case we end up with duplicated trips

The logic in MOTIS needs some serious work for duplicate detection and probably the SIRI matching logic is a very good starting point. But at the moment I would not want to bet a good duplicate detection in MOTIS :-/
I'd rather add an agency filter in MOTIS as that should be quite easy (although I though that this is what we already have in gtfs-clean).

jbruechert:
> > In the worst case we end up with duplicated trips
> 
> The logic in MOTIS needs some serious work for duplicate detection and probably the SIRI matching logic is a very good starting point. But at the moment I would not want to bet a good duplicate detection in MOTIS :-/ I'd rather add an agency filter in MOTIS as that should be quite easy (although I though that this is what we already have in gtfs-clean).

We currently can only remove agencies, not specify a list to keep. That would be relatively simple to add as well of course.

sandervhoogdalem:
> > > In the worst case we end up with duplicated trips
> > 
> > 
> > The logic in MOTIS needs some serious work for duplicate detection and probably the SIRI matching logic is a very good starting point. But at the moment I would not want to bet a good duplicate detection in MOTIS :-/ I'd rather add an agency filter in MOTIS as that should be quite easy (although I though that this is what we already have in gtfs-clean).
> 
> We currently can only remove agencies, not specify a list to keep. That would be relatively simple to add as well of course.

This is also nice when its possible.

jbruechert:
> > > > In the worst case we end up with duplicated trips
> > > 
> > > 
> > > The logic in MOTIS needs some serious work for duplicate detection and probably the SIRI matching logic is a very good starting point. But at the moment I would not want to bet a good duplicate detection in MOTIS :-/ I'd rather add an agency filter in MOTIS as that should be quite easy (although I though that this is what we already have in gtfs-clean).
> > 
> > 
> > We currently can only remove agencies, not specify a list to keep. That would be relatively simple to add as well of course.
> 
> This is also nice when its possible.

I already have to many sidequests in this project, but if you want to look into it, it should be a relatively minor change here: https://github.com/public-transport/gtfsclean

sandervhoogdalem:
Is the flag for removing agencies placed in gtfsclean @jbruechert?

sandervhoogdalem:
I already see this file, is there a reason why we can't use this? 

https://github.com/public-transport/gtfsclean/blob/master/processors/agencyduplicateremover.go

sandervhoogdalem:
Another option is using NeTEx files so we can remove the stopareas. But NeTEx is not supported at the moment right? @felixguendling

jbruechert:
> I already see this file, is there a reason why we can't use this? 
> 
> https://github.com/public-transport/gtfsclean/blob/master/processors/agencyduplicateremover.go

This is only per feed

sandervhoogdalem:
> > I already see this file, is there a reason why we can't use this?
> > https://github.com/public-transport/gtfsclean/blob/master/processors/agencyduplicateremover.go
> 
> This is only per feed

Ahh makes sense. But the flags are programmed in GTFSClean?

jbruechert:
gtfsclean is the tool actuall implementing most of the preprocessing funcctionality. The python script in this repo just calls it with the right flags, according to what is set in the json.

sandervhoogdalem:
> gtfsclean is the tool actuall implementing most of the preprocessing funcctionality. The python script in this repo just calls it with the right flags, according to what is set in the json.

Okay I will look into it tomorrow

sandervhoogdalem:
@jbruechert @felixguendling I've made a change on my device at the GTFS Clean repo so we can use a flag called "keep-agency-names" and I am ready to make a pull request there but I have a question. I tested with both removing and keeping and I noticed someting: stops.txt is not modified also not with the removing flag. How is motis handling stops that are duplicated from multiple feeds?

jbruechert:
Unused stops would be merged if there is a stop at the same position, but if some end up not being merged, they'd still appear in the geocoding endpoint.

sandervhoogdalem:
Good to know. For now its waiting until the gtfsclean PR is merged so I can update the nl.json for European Sleeper

jbruechert:
I think setting `--delete-orphans` should also remove the leftover stops.

sandervhoogdalem:
I 've added OVApi only for European Sleeper to the nl.json in the pull request now and set the flag for only European Sleeper. Could you check if I did it correct @jbruechert @felixguendling?

feeds/nl.json
```diff
@@ -3,32 +3,62 @@
...
             "type": "url",
             "spec": "gtfs-rt",
-            "url": "https://gtfs.ovapi.nl/nl/tripUpdates.pb"
+            "url": "https://gtfs.openov.nl/gtfs-rt/alerts.pb"
+        },
+        {
+            "name": "OpenOV",
+            "type": "url",
+            "spec": "gtfs-rt",
+            "url": "https://gtfs.openov.nl/gtfs-rt/vehiclePositions.pb"
```
felixguendling, line 45: MOTIS currently can't handle vehicle positions.
We can leave it here, so we don't have to add it again in case MOTIS gains functionality in this area.
It will just always be 100% ignored updates.

sandervhoogdalem: Ahh okay, good to know. I learn a lot these days from you guys

jbruechert:
> I 've added OVApi only for European Sleeper to the nl.json in the pull request now and set the flag for only European Sleeper. Could you check if I did it correct @jbruechert @felixguendling?

A small part is missing, that is adding the new json option to metadata.py and converting it to the correct gtfsclean argument in fetch.py. Also I'd propose to add `--delete-orphans` there in general.

sandervhoogdalem:
> > I 've added OVApi only for European Sleeper to the nl.json in the pull request now and set the flag for only European Sleeper. Could you check if I did it correct @jbruechert @felixguendling?
> 
> A small part is missing, that is adding the new json option to metadata.py and converting it to the correct gtfsclean argument in fetch.py. Also I'd propose to add `--delete-orphans` there in general.

Are you able to make a change for that or do I need to do it myself @jbruechert? And what do you mean with metadata.py? Do you have an example?

Edit: I found the files. I will make a change there as soon as I have time for it

sandervhoogdalem:
I saw a fail with a workflow run so I checked the JSON and saw a "," and a blank line. I removed them in a new commit. Could you guys check again? @jbruechert @felixguendling

feeds/nl.json
```diff
@@ -3,32 +3,61 @@
...
-            "name": "ovapi",
-            "type": "transitland-atlas",
-            "transitland-atlas-id": "f-u-nl",
-            "fix": true
+            "name": "OpenOV",
+            "type": "http",
+            "url" : "https://gtfs.openov.nl/gtfs-rt/gtfs-openov-nl.zip",
+            "fix": true,
+            "license": {
+                "spdx-identifier": "CC0 1.0 Universal",
```
jbruechert:
```suggestion
                "spdx-identifier": "CC0-1.0",
```

(See https://spdx.org/licenses/ for the list of valid identifiers)

sandervhoogdalem:
@jbruechert I updated the file with your suggested change

sandervhoogdalem:
@jbruechert @felixguendling i see a fail again but its not from NL.json right?

jbruechert:
Ah yes, that got confused because the branch is behind and pl.json was changed on main.

But everything looks good now, thanks!

PR #1377, opened 2025-07-25
DE: Using the DELFI SIRI feeds

traines-source:
I think it's good enough to switch to the matched SIRI feeds (updated every 5 minutes only). This also still includes the data from the DELFI GTFS-RT Feed (updated every minute).
Not adding Service Alerts for the time being due to https://github.com/mfdz/GTFS-Issues/issues/189.


jbruechert:
Nice, thanks!

PR #1380, opened 2025-07-26
Add RT Feeds from Goeuropa (Kombus, Środa, PKS Poznań)

kasmar00:
- Add RT feed for PKS Poznań
- Add RT feed for Kombus
- Add Środa Wielkopolska


jbruechert:
Thanks!

PR #1381, opened 2025-07-26
Add support for MOTIS extend_calendar option

jbruechert:


felixguendling:
Currently, there are strict requirements:

- the feed needs to have `feed_info.txt` with a `feed_end_date` set
- the service that should be looped needs to have `end_date == feed_end_date`

We'll have to see if that's too strict. For the feed I tested with it worked.

PR #1384, opened 2025-07-28
fix pl gtfs rt config

traines-source:


traines-source:
maybe the `pl-Gorzów-Wielkopolski` feed is a candidate for the loop-expired-feeds-feature?

kasmar00:
> maybe the `pl-Gorzów-Wielkopolski` feed is a candidate for the loop-expired-feeds-feature?

🫠 Seems like they have unstable links. Currently their "public information website" lists: https://bip.mzk-gorzow.com.pl/files/file_add/download/48_gtfs_gw.zip 🫠
I'll try to generate a feed with a static link for them.

jbruechert:
If there is some way to derive the current link, please use the "function" feature in the metadata.

PR #1388, opened 2025-07-29
Add two-level hierarchy to attribution page

### LINKED ISSUE #756: [website] In sources, move subdivisions into country sections (Canada and USA)

vkrause:
This puts subdivisions under the corresponding country now, making the page easier to navigate.

Fixes #756.


REUSE.toml
```diff
@@ -8,12 +8,6 @@ precedence = "aggregate"
 SPDX-FileCopyrightText = "none"
 SPDX-License-Identifier = "CC0-1.0"
 
-[[annotations]]
-path = "website/data/license.json"
```
jbruechert: Why no more test data?

vkrause: Oh, this is actually used somewhere? The data seemed unused and outdated (and thus confusing), that's why I removed it.

website/layouts/shortcodes/source-table.html
```diff
@@ -9,27 +9,43 @@
...
+    <details>
+      <summary><h2>{{ .country_name }}</h2></summary>
+  {{ end }}
+
+  {{ if ne $subdivision $lastSubdivision }}
+    {{ if ne $lastSubdivision "_initial_" }}
+      </details>
+    {{ end }}
+    {{ if $subdivision }}
     <details>
```
jbruechert:
```suggestion
    <details style="margin-left: 50px">
```

REUSE.toml
jbruechert: I just use it to be able to build the website without additional steps

vkrause: I see. I'll replace it with the current version instead then (as the website now needs the extra fields in there).

jbruechert:
Thanks!

PR #1389, opened 2025-07-29
Update us-ca.json

hoermalmeister:
Added govcbus Oxnard

### DIFF
--- a/feeds/us-ca.json
+++ b/feeds/us-ca.json
@@ -23,6 +23,11 @@
             "fix": true
         },
         {
+            "name": "GOVCbus",
+            "type": "http",
+            "url": "https://s3.transitpdf.com/files/uran/improved-gtfs-govcbus.zip"
+        },
+        {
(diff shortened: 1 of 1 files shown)


feeds/us-ca.json
```diff
@@ -22,6 +22,11 @@
             "transitland-atlas-id": "f-9qc-capitolcorridor",
             "fix": true
         },
+        {
+            "name": "GOVCbus",
+            "type": "http",
+            "url": "https://s3.transitpdf.com/files/uran/improved-gtfs-govcbus.zip"
```
jbruechert: This website tends to just mirror feeds from other places (sometimes with pretty long delays for updates), so it would be better to find the original link.

jbruechert: This transitland feed `f-9q5-govcbus` looks like the right thing.

hoermalmeister: Sure thing, I'll keep that in mind for future

PR #1390, opened 2025-07-29
Dynamic link for Gorzow Wlkp, add Olsztyn

kasmar00:
- function generating latest file for Gorzow Wielkopolski, as suggested in: https://github.com/public-transport/transitous/pull/1384#issuecomment-3131695666
- add Olsztyn (AFAIK last missing from those available on zbiorkom.live, at least for now)

### DIFF
--- a/feeds/pl.json
+++ b/feeds/pl.json
@@ -444,6 +444,16 @@
             }
         },
         {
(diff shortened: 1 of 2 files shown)


jbruechert:
Thanks!

PR #1392, opened 2025-07-29
Update pl.json feeds

Lach-anonym:
Actually found the last remaining feed from zbiorkom.live And also Skarżysko-Kamienna has appeared

### DIFF
--- a/feeds/pl.json
+++ b/feeds/pl.json
@@ -396,6 +396,16 @@
             "fix": true
         },
         {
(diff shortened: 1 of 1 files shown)


Lach-anonym:
KD feed requires authentication now

jbruechert:
> KD feed requires authentication now

I had a quick look, looks like a bug on our side and not like a change on theirs. I hope it's fixed now.

PR #1394, opened 2025-07-30
Sort attributions by human readable names rather than ISO 3166 codes

vkrause:
Makes the attribution list easier to navigate.

### DIFF
--- a/src/generate-attribution.py
+++ b/src/generate-attribution.py
@@ -172,6 +172,10 @@ def get_region_data(code: str) -> dict:
     return region_data
 
 
+def attribute_sort_key(attr):
+    return (attr[1].get("country_name", ""), attr[1].get("subdivision_name", ""), attr[1].get("human_name", ""), attr[1].get("filename", ""))
+
+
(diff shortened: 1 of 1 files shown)


src/generate-attribution.py
```diff
@@ -165,13 +165,13 @@ def get_region_data(code: str) -> dict:
         region_data["subdivision_code"] = code
         region_data["subdivision_name"] = subdivision.name if subdivision else SUBDIVISIONS[code]
 
-    # TODO backward compatibility for not yet adapted consumer code, remove eventually
-    region_data["region_code"] = region_data.get("subdivision_code", region_data["country_code"])
```
jbruechert: It looks like Träwelling is parsing this, so they need to be contacted before this can be removed.

jbruechert: I'll open an issue there.

vkrause: Oh, I had no idea. Dropped from this PR for now.

jbruechert:
Thanks!

PR #1395, opened 2025-07-30
Fix grouping of subdivisions

vkrause:
With the new sorting the empty subdivisions come first, which the previous code didn't handle properly.

Also, sort feed names case-insensitively.


jbruechert:
Thanks!

PR #1402, opened 2025-08-02
nz: add AT RT key and protobuf accept header

applecuckoo:
cc @Hoverth

The `Accept: application/x-protobuf` header is there since by default AT's APIs will return a JSON version of the feed.


jbruechert:
Thanks!

PR #1408, opened 2025-08-04
NZ: fix RT feeds for AT and Metlink

applecuckoo:
cc @Hoverth

This PR sets up the Metlink and AT RT feeds manually since the Transitland import overrides our custom `Accept` header.

also add myself as one of the maintainers

PR #1409, opened 2025-08-04
NZ: fix RT feeds for AT and Metlink (again)

applecuckoo:
succeeds #1408

cc @Hoverth

This PR sets up the Metlink and AT RT feeds manually since the Transitland import overrides our custom `Accept` header.

also add myself as one of the maintainers

### DIFF
--- a/feeds/nz.json
+++ b/feeds/nz.json
@@ -3,6 +3,10 @@
         {
             "name": "Thomas Dickson",
             "github": "Hoverth"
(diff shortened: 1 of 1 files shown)


Hoverth:
Looks good to me, yeah having to remember to delete past commits and restart from main can be a bit of a pain at times.

mlundblad:
Thanks!

PR #1414, opened 2025-08-05
nz: fix headers yet again

applecuckoo:
should be all good this time

### DIFF
--- a/feeds/nz.json
+++ b/feeds/nz.json
@@ -20,11 +20,9 @@
             "spec": "gtfs-rt",
             "type": "url",
             "url": "https://api.at.govt.nz/realtime/legacy",
-            "http-options": {
-                "headers": {
-                    "Accept": "application/x-protobuf",
-                    "Ocp-Apim-Subscription-Key": "5a8bd844b5c941a188a17d6ed968c070"
-                }
+            "headers": {
+                "Accept": "application/x-protobuf",
+                "Ocp-Apim-Subscription-Key": "5a8bd844b5c941a188a17d6ed968c070"
             }
         },
         {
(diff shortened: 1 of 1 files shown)


felixguendling:
Looks like this is not documented here:
https://transitous.org/doc/#realtime-feeds

vkrause:
> Looks like this is not documented here: https://transitous.org/doc/#realtime-feeds

It's in https://transitous.org/doc/#more-source-options, but the subtle difference between static and realtime feeds for this is indeed not very clear there (and ideally we'd unify that entirely). The Swiss feeds use this pattern as well for example.

felixguendling:
Yes, the option itself is listed but not when to use which. Maybe it makes sense to have one (complete) example for each case?

Having GTFS-Flex as a `spec` type is also a bit confusing as GTFS Flex has been just a part of the GTFS spec for some time now. Even if many feed producers happen to separate the feeds, there's technically no need to do it and it would even make more sense to have one feed that combines line based transit with on-demand transit (to share the same stop ids, agency ids, etc.). Since it seems only to be used to prevent `gtfs-clean` from removing the fields, it could probably also be made a flag (something like `no-fix: true`)?

Both not urgent and everything seems to run fine anyway (probably I'm the only one confused) - just some ideas :-)

jbruechert:
Is it possible the special casing of gtfs flex is already unneccessary since the feature to keep additional files was added for fares V2?

vkrause:
> Is it possible the special casing of gtfs flex is already unneccessary since the feature to keep additional files was added for fares V2?

Unfortunately not, the changed rules for `stop_id` are still missing at least.

PR #1425, opened 2025-08-07
Support Transitland feeds with multiple GTFS-RT URLs

### LINKED ISSUE #1328: Some transitland feeds have multiple URLs, transitous uses only one of them
For example in https://github.com/transitland/transitland-atlas/blob/adb732ec6ad02241171913f6e03f337cae9813ca/feeds/tu-vp.goswift.ly.dmfr.json#L776

```
"urls": {
        "realtime_vehicle_positions": "https://api.goswift.ly/real-time/lubbock-citibus/gtfs-rt-vehicle-positions",
        "realtime_trip_updates": "https://api.goswift.ly/real-time/lubbock-citibus/gtfs-rt-trip-updates"
      },
```

Looking at the code in `transitland.py`, it appears that only the `realtime_trip_updates`  URL would end up being used in transitous.

vkrause:
This gives us a whole bunch of additional GTFS-RT feeds, mostly service alerts but also a few more trip update ones.

This does not filter out vehicle position feeds, which would generate extra load for no benefit, so presumably we want to do that for now? There seem to be feeds with vehicle position data only currently being added, those would break as a result though.

Fixes #1328.

### DIFF
--- a/src/fetch.py
+++ b/src/fetch.py
@@ -267,12 +267,12 @@ class Fetcher:
     def resolve_database_sources(self, source: Source) -> Source:
         match source:
             case TransitlandSource():
-                http_source = self.transitland_atlas.source_by_id(source)
+                http_source = self.transitland_atlas.sources_by_id(source)
                 if not http_source:
                     eprint("Error: Could not resolve", source.transitland_atlas_id)
                     sys.exit(1)
 
-                return http_source
+                return http_source[0]  # multi-source feeds only occur for GTFS-RT, which we don't handle here
             case MobilityDatabaseSource():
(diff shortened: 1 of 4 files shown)


felixguendling:
I would hope that https://github.com/motis-project/nigiri/pull/241 is ready next week to merge. Then, it would make sense to add vehicle positions if there's no trip update feed. If there's a trip update feed, there's no need for vehicle positions.

To display vehicle positions on a map, the update rate would need to be very high in order to show a "close to real-time" position. I'm not sure if we should make so many requests (at least 1x/10s)?

PR #1427, opened 2025-08-08
cy realtime

hoermalmeister:
So I figured the gtfs-rt for cyprus but it seems the static is broken now, it doesn't include any stops or stop_times, it seems two days ago it was still fine


jbruechert:
Let's merge this and hope they fix their feeds again

PR #1428, opened 2025-08-08
cz:IDZK and GRAPP realtime

hoermalmeister:
added own source for fetching realtime data from PMDP api

### DIFF
--- a/feeds/cz.json
+++ b/feeds/cz.json
@@ -24,6 +24,12 @@
             "fix": true
         },
         {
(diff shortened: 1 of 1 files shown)


felixguendling:
The feed seems to be empty?

hoermalmeister:
Indeed, I ran into issues with the token, so I am working on it rn

hoermalmeister:
Ready for review

PR #1431, opened 2025-08-09
Update ba.json, hr.json, rs.json & si.json

vekejsn:
Fixes:
- Mostar GTFS URL

Changes:
- Srbijavoz GTFS to different merged source between OeBB HAFAS & w3.srbvoz.rs
- IJPP GTFS to use a processed version, that is more user friendly
- Liburnija Zadar GTFS-RT source

Adds:
- HŽPP GTFS-RT
- Split GTFS-RT
- Sisak GTFS-RT
- IJPP GTFS-RT
- Maribor GTFS & GTFS-RT
- Celje GTFS & GTFS-RT
- Srbijavoz GTFS-RT

### DIFF
--- a/feeds/ba.json
+++ b/feeds/ba.json
@@ -29,7 +29,8 @@
         {
             "name": "Mostar",
             "type": "http",
-            "url": "https://gitlab.com/vekejsn/gtfs-generators/-/jobs/artifacts/main/raw/mobus_gtfs.zip?job=mostar"
+            "url": "https://gitlab.com/vekejsn/gtfs-generators/-/jobs/artifacts/main/raw/mobus_gtfs.zip?job=mostar",
+            "fix": true
         }
     ]
 }
--- a/feeds/ba.json
+++ b/feeds/ba.json
@@ -29,7 +29,7 @@
         {
             "name": "Mostar",
             "type": "http",
-            "url": "https://gitlab.com/vekejsn/gtfs-generators/-/jobs/artifacts/main/raw/zadar_gtfs.zip?job=zadar"
+            "url": "https://gitlab.com/vekejsn/gtfs-generators/-/jobs/artifacts/main/raw/mobus_gtfs.zip?job=mostar"
         }
     ]
 }
--- a/feeds/hr.json
+++ b/feeds/hr.json
@@ -152,7 +152,7 @@
             "name": "zadar",
             "type": "url",
             "spec": "gtfs-rt",
-            "url": "https://owncloud.cesnet.cz/index.php/s/hoLRLvxxORnoPOw/download"
+            "url": "https://rt.gtfs.baguette.pirnet.si/gtfs-rt/liburnijaZadar/trip_updates.pb"
         },
         {
             "name": "karlovac",
--- a/feeds/hr.json
+++ b/feeds/hr.json
@@ -146,7 +146,7 @@
         {
             "name": "zadar",
             "type": "http",
-            "url": "https://gitlab.com/vekejsn/gtfs-generators/-/jobs/artifacts/main/raw/zadar_gtfs.zip?job=zadar",
+            "url": "https://gitlab.com/vekejsn/gtfs-generators/-/jobs/artifacts/main/raw/zadar_gtfs.zip?job=zadar"
         },
         {
             "name": "zadar",
--- a/feeds/hr.json
+++ b/feeds/hr.json
@@ -147,7 +147,6 @@
             "name": "zadar",
             "type": "http",
             "url": "https://gitlab.com/vekejsn/gtfs-generators/-/jobs/artifacts/main/raw/zadar_gtfs.zip?job=zadar",
-            "fix": true
         },
         {
             "name": "zadar",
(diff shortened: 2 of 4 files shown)


mlundblad:
Srbijavoz seems to be failing due to some trip arrival times are before the departure of the previous stop. Maybe adding "fix: true" would correct this.

jbruechert:
I'm a bit concerned that the Srbijavoz feed only goes very few days into the future.

I wonder if `https://w3.srbvoz.rs/redvoznje//api/vozdetalji1?idvoza=104&brojvoza=8325%20&datum=31-07-2025&stanicaod=16102&stanicado=16103` already contains enough information that we could convert it to SIRI and rely on MOTIS's fuzzy matching.

CC @traines-source

jbruechert:
@vekejsn Ideally only add fix were needed. This option is inherently lossy, and especially since you are converting the feeds yourself, it's probably usually better to notice the issues and fix them in the generator, than to silently remove the trips.

vekejsn:
> I'm a bit concerned that the Srbijavoz feed only goes very few days into the future.

Sadly it requires scraping the trips for each day, I can extend it, but every extra day requires ~400 requests (1 per stop, to check what trains are running that day). Would 14 days be enough or would you need more?

> it's probably usually better to notice the issues and fix them in the generator, than to silently remove the trips

I've removed some of them, the ones that are left remaining are the ones which have poor source data, so it's a pain to get them to be 100% valid. The Srbijavoz GTFS seems to import into MOTIS fine, even with the error on the single trip.

jbruechert:
> > I'm a bit concerned that the Srbijavoz feed only goes very few days into the future.
> 
> Sadly it requires scraping the trips for each day, I can extend it, but every extra day requires ~400 requests (to check what trains are running that day). Would 14 days be enough or would you need more?

I think it's hard to say how much is really needed, but planning trips > 2 month into the future is probably a relevant use case.
I'm wondering if we can "easily" match the real time information to the existing feed which has a bit larger timeframe.

> I've removed some of them, the ones that are left remaining are the ones which have poor source data, so it's a pain to get them to be 100% valid. The Srbijavoz GTFS seems to import into MOTIS fine, even with the error on the single trip.

Yes, but MOTIS will just use the latest time it has seen on the trip so far to avoid going back in time, that doesn't make the data correct.

Using `fix` for the unreliable sources makes sense :+1:

vekejsn:
> I'm wondering if we can "easily" match the real time information to the existing feed which has a bit larger timeframe.

Maybe, though with the existing feed we lose the BG:VOZ suburban trains (route_short_name=BG xxxx in the new one). Maybe a combination of both feeds would be ideal (adding BG:VOZ and adding the "range" information from the other feed)? Scraping the data for BG:VOZ only requires ~10 requests for each day of operating.

jbruechert:
> > I'm wondering if we can "easily" match the real time information to the existing feed which has a bit larger timeframe.
> 
> Maybe, though with the existing feed we lose the BG:VOZ suburban trains (BGxxxx in the new one). Maybe a combination of both feeds would be ideal (adding BG:VOZ and adding the "range" information from the other feed)? Scraping the data for BG:VOZ only requires ~10 requests for each day of operating.

Ah true, that sounds good! If MOTIS deduplication was a bit more advanced already we could just add both, right now we'd duplicate the trips. If you can generate one just for BG:VOZ that would be great!

I want to wait for @traines-source opinion on how to best do the realtime matching, because he has done lots of work on that already.

hoermalmeister:
@vekejsn you can switch the Liburnija Zadar rt source to yours as well, because I noticed mine doesn't match all of the trips correctly, even though we are using the same api. It matches for some trips, but not all of them

felixguendling:
> I'm a bit concerned that the Srbijavoz feed only goes very few days into the future.
> 
> I wonder if `https://w3.srbvoz.rs/redvoznje//api/vozdetalji1?idvoza=104&brojvoza=8325%20&datum=31-07-2025&stanicaod=16102&stanicado=16103` already contains enough information that we could convert it to SIRI and rely on MOTIS's fuzzy matching.
> 
> CC @traines-source

For SIRI fuzzy matching we need:

- matching stop IDs or at least geo coordinates for both IDs (schedule vs rt)
- scheduled time for departure/arrival
- realtime for departure/arrival
- partial trips work

As far as I can tell, the feed doesn't have the scheduled times. We could try to use the real-time times as scheduled times. However, if a trip is so late that it's closer to the next frequency than to it's own schedule time, it will produce mismatches.

jbruechert:
> As far as I can tell, the feed doesn't have the scheduled times. We could try to use the real-time times as scheduled times. However, if a trip is so late that it's closer to the next frequency than to it's own schedule time, it will produce mismatches.

As fas as I understand, the feed only has scheduled times and a delay in minutes in a separate field. Coordinates would still have to be added based on ids / names. But I think we have a solution that works for now now.

vekejsn:
That is correct, though you also get the information which stop reported the delay (though this doesn't have to line up with the actual stops the train makes).

jbruechert:
The headsigns in the new feed look a bit wrong, leading to things like this:

image

You can just leave these empty, MOTIS will fall back to the last stop.

PR #1438, opened 2025-08-12
Update au.json

g1nger-bred:
added all modes of Melbourne-based transit and renamed previous edit to reflect true source of data

### DIFF
--- a/feeds/au.json
+++ b/feeds/au.json
@@ -161,9 +161,42 @@
             "fix": true
         },
         {
-            "name": "Public-Transit-Victoria-(PTV)",
+            "name": "V-Line",
             "type": "transitland-atlas",
             "transitland-atlas-id": "f-r1-ptv"
+        },
+        {
+            "name": "PTV-Metro",
+            "type": "transitland-atlas",
+            "transitland-atlas-id": "f-r1r-ptv"
+        },
+        {
+            "name": "PTV-Metro-rt",
+            "spec": "gtfs-rt",
+            "type": "http",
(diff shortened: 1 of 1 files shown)


feeds/au.json
```diff
@@ -161,9 +161,42 @@
...
+        {
+            "name": "PTV-Metro",
+            "type": "transitland-atlas",
+            "transitland-atlas-id": "f-r1r-ptv"
+        },
+        {
+            "name": "PTV-Metro-rt",
+            "spec": "gtfs-rt",
+            "type": "http",
+            "url": "https://api.opendata.transport.vic.gov/opendata/public-transport/gtfs/realtime/v1/metro"
```
jbruechert: This domain doesn't resolve for me unfortunately

applecuckoo: @g1nger-bred @jbruechert it's missing the `.au` at the end, the full domain should be `api.opendata.transport.vic.gov.au`

mlundblad: I tried downloading from the URLs now that it's updated with .au.

I'm getting a response, but it's a SOAP fault. Is this expected? (I guess it should be a POST call with some params to get realtime data.

@jbruechert do you think this is mergable now?

applecuckoo: I had a look into this and it looks like the individual endpoints (`/trip-updates`, `/vehicle-positions`, and `/service-alerts` for trams and trains, and just `/trip-updates` for buses) need to be added as the single link isn't a valid endpoint. You also need to get an API key from [here](https://opendata.transport.vic.gov.au/) and supply it using the `KeyID` header.

jbruechert:
Looks like this has been done in a different way in the meantime.
Thanks for your efforts nevertheless!

PR #1440, opened 2025-08-13
Fi:HSL switch to trip updates

hoermalmeister:


jbruechert:
Why only one of the two?

hoermalmeister:
Can we have two rt sources for one static?

jbruechert:
> Can we have two rt sources for one static?

Yes, as many as neccessary. See de-DELFI for example

hoermalmeister:
In that case, done

jbruechert:
Vehicle positions are currently not used by MOTIS, but I guess it doesn't hurt to collect them already.

Issue #1015, opened 2025-03-18
Fix train display names in DELFI Feed


jbruechert:
for ICEs, the correct numbers are in trip_short_name, but they are not prefixed with ICE. I don't think the current gtfsclean features are enough to deal with this.

felixguendling:
> For ICEs in Germany, the correct numbers are in trip_short_name, but they are not prefixed with ICE.

> per route_type
> regex to decide which route names to keep

I think the way we had it in nigiri before could've labeled also TGV as ICE. I didn't check. But just based on `route_type` they might be indistinguishable.

felixguendling:
For DELFI it could also make sense to propose improvements here: https://github.com/mfdz/GTFS-Issues

jbruechert:
> > For ICEs in Germany, the correct numbers are in trip_short_name, but they are not prefixed with ICE.
> 
> > per route_type
> > regex to decide which route names to keep
> 
> I think the way we had it in nigiri before could've labeled also TGV as ICE. I didn't check. But just based on `route_type` they might be indistinguishable.

Yes, that is a separate problem. For now this is still about displaying the correct number :)

miklcct:
I think the route here is "ICE", which is something the rider understands.

jbruechert:
> I think the route here is "ICE", which is something the rider understands.

Which example are you referring to?

miklcct:
> > I think the route here is "ICE", which is something the rider understands.
> 
> Which example are you referring to?

ICE in Germany.

felixguendling:
> * put route name into `route_short_name` **if** commonly understood
> * put trip number into `trip_short_name` **if** commonly understood
> * always display both
> * remove everything from the feeds that is not understood by riders.

100% agree.

I think this strategy is the right thing to do.
Everything else doesn't make sense.

felixguendling:
> pass out route_short_name and trip_short_name raw in the API
display both in the web interface

Feel free to open a PR. Maybe the `string_idx_t` "compression" like I introduced it in fares could be applied here too to de-duplicate the strings.

jbruechert:
Result of a quick discussion with @vkrause today:
- Since either route_short_name or route_long_name is required, we are missing a way to represent trains which are identified only by the trip name
- Many railways fall back to only specifying route_long_name and trip_short_name, which leads to Google Maps not displaying route_long_name (lack of space?)
[...]

PR #1448, opened 2025-08-16
Add support for MOTIS v2.2 import scripts

vkrause:
The idea here is to have a top-level folder "scripts" containing scripts named after the feeds they are for.

The upside of this is that we don't need to maintain a separate configuration, and common code needed for multiple feeds can be factored out into a shared file.

The downside is that any deviation in the naming will make this silently ignore the script.

### DIFF
--- a/src/generate-motis-config.py
+++ b/src/generate-motis-config.py
@@ -33,6 +33,7 @@ if __name__ == "__main__":
     arguments = parser.parse_args()
 
     feed_dir = Path("feeds/")
+    script_dir = Path("scripts/")
 
     atlas = transitland.Atlas.load(Path("transitland-atlas/"))
     mdb = mobilitydatabase.Database.load()
@@ -118,6 +119,8 @@ if __name__ == "__main__":
                                 }
                             if source.default_timezone is not None:
                                 config["timetable"]["datasets"][name]["default_timezone"] = source.default_timezone
+                            if os.path.exists(os.path.join(script_dir, f"{name}.lua")):
+                                config["timetable"]["datasets"][name]["script"] = f"{script_dir}/{name}.lua"
 
                         case "gtfs-rt" if isinstance(source, metadata.UrlSource):
                             name = f"{region_name}-{source.name}"
@@ -153,3 +156,7 @@ if __name__ == "__main__":
 
         with open("out/config.yml", "w") as fo:
             yaml.dump(config, fo)
+
+    # copy scripts
+    shutil.rmtree("out/scripts", ignore_errors=True)
+    shutil.copytree(script_dir, "out/scripts")


mlundblad:
Would it make sense to support "generic" scripts? Like for example if we add scripts to do more generic augmenting (like e.g. looking up route colors from Wikidata, or things like that).

Or maybe those could just be imported in the region-specific scripts when applicable…

vkrause:
> Would it make sense to support "generic" scripts? Like for example if we add scripts to do more generic augmenting (like e.g. looking up route colors from Wikidata, or things like that).
> 
> Or maybe those could just be imported in the region-specific scripts when applicable…

Right, was wondering that as well, hard to tell without knowing what we'll come up with yet. Even for generic things I could imagine some limited amount of parameterization would often be necessary, in that case importing generic code would probably make sense.

An alternative approach would be to have a `script` entry in the JSON data, which wouldn't be too bad either. Then sharing scripts would be trivial, and we would get import errors if the script is wrongly named. If the 800 French feeds ever all need the same fix applied this is probably the only viable way.

We could also allow both ways at the same time, but not sure whether that would give us the best or the worst of both worlds.

jbruechert:
Adding a new json property for this would also have the advantage of making renames of feeds or scripts not unexpectedly break running the right script without noticing.

felixguendling:
Scripts are currently implemented in a way that if they fail, the import will just ignore it (so partial results from the script might be there until it crashed).

https://github.com/motis-project/nigiri/blob/97223a23c492bf4ec20f9d6e80f7c572f5980f47/src/loader/register.cc#L345-L351

So monitoring the logs in `data/logs/tt.txt` is important. Maybe a check that `grep` doesn't find any script failures on the first import could make sense in the CI.

We could also make the import crash if a script doesn't work.

I haven't tried if/how `require` works in those lua scripts or if we have to enable something for this to work.

vkrause:
Interesting, I was thinking about adding a Lua syntax check to the CI until I actually got import failures on script errors here, so I actually assumed this was the intended behavior.

vkrause:
The new version now uses an explicit `script` key in the metadata and produces a fatal error if the referenced script isn't found.

vkrause:
> I haven't tried if/how `require` works in those lua scripts or if we have to enable something for this to work.

`require` works fine with https://github.com/motis-project/nigiri/pull/249, although I'm not quite sure yet how to properly control search paths for that.

felixguendling:
Thanks for the PR!

The search path seems to be easy to override in case the default behavior doesn't fit for us:
https://github.com/ThePhD/sol2/issues/692


> I actually got import failures on script errors 

That's probably not what we want? Because in this case, it can lead to one script killing the whole nightly import. Maybe we need more try/catch or error checking somewhere? My intention was to make it robust and not depend on scripts being 100% correct. Especially with a scripting language, the script might work all the time until the data changes, which sends the script in a branch with errors.

Checking the script statically is probably not possible except the static checker knows which functions on the extra objects like trip, location, route, agency exist.

vkrause:
Having worked with this a bit more, I've also seen the runtime failure behavior now. It looks like import failures happen when there's already an error when loading the script (syntax error, or runtime error in the global part, which is why I was hitting that frequently while experimenting with `require`), errors in the processing functions are indeed non-fatal as intended. That's probably fine, as global script errors should be independent of the input data and thus not occur randomly on data changes.

felixguendling:
Feel free to add the script I wrote for testing if you think it helps: https://github.com/public-transport/transitous/issues/1015#issuecomment-3186147851

Since the new version is deployed now, this PR can also be merged now.

jbruechert:
I think we are ready to merge this now, all instances are upgraded

vkrause:
Resolved the merge conflict.

vkrause:
CI error was an unrelated issue with the BlaBlaCar bus feed, pushed a fix for that as well.

jbruechert:
Looks good, feel free to merge when ready.

traines-source:
Sorry, didn't realize that we need to change something on our end too, i.e. import didn't run today, i.e. no scripts yet. Should be easy to fix though.

traines-source:
Ok, import ran now, and I found a FlixTrain with route_type 102 https://api.transitous.org/?tripId=20250925_12%3A50_eu-flixbus_FLX10-1249-1250092025-BL%23TS-00. Tomorrow with the DELFI fixes it should be more easily visible.

vkrause:
Thank you! Sorry, I didn't realize this needed manual intervention on your side.

PR #1457, opened 2025-08-22
Update Kalisz area feeds and add licenses

Lach-anonym:
I'm not sure if it will work, the download for GTFS feeds from KLA and local bus operators does not want to terminate on my side, however the incomplete download datas are actually complete GTFS files
I hope the Słupsk link search strategy will work here

### DIFF
--- a/feeds/pl.json
+++ b/feeds/pl.json
@@ -270,10 +270,35 @@
         },
         {
             "name": "PKS-Kalisz",
-            "type": "mobility-database",
-            "mdb-id": "mdb-2461",
-            "skip": true,
-            "skip-reason": "Outdated feed",
+            "type": "http",
+            "spec": "gtfs",
+            "url": "https://pks.kalisz.pl/GTFS/",
+            "function": "data_slupsk_latest_resource",
+            "license": {
+                "spdx-identifier": "CC0-1.0"
+            },
+            "fix": true
+        },
+        {
+            "name": "KLA-Kalisz",
+            "type": "http",
+            "spec": "gtfs",
+            "url": "rj.sdip.kalisz.pl/?download=gtfs_kla",
+            "license": {
+                "spdx-identifier": "CC0-1.0"
+            }
+        },
+        {
+            "name": "Kalisz-LocalBuses",
+            "type": "http",
+            "spec": "gtfs",
+            "url": "rj.sdip.kalisz.pl/?download=gtfs_przewoznicy",
+            "license": {
+                "spdx-identifier": "CC0-1.0"
+            },
+            "drop-agency-names": [
+                "PKS w Kaliszu Sp. z o.o."
+            ],
             "fix": true
         },
         {
@@ -604,6 +629,9 @@
             "spec": "gtfs",
             "url": "https://zimslupsk.pl/otwarte-dane.html",
             "function": "data_slupsk_latest_resource",
+            "license": {
+                "spdx-identifier": "CC0-1.0"
+            },
             "fix": true
         },
         {


jbruechert:
Unfortunately it looks like the length that the server sends is wrong, and the entire body is needed. Most (all I tested) download tools seem to not do that.
I think this is unfortunately another case where reporting this to the server administrator makes the most sense.

Lach-anonym:
I have good news - the downloads now have the file size, so they work at least in my browser - can anyone else test that aswell?

jbruechert:
Looks like they added a new problem

Lach-anonym:
Right, I'll get to solve this at some point

kasmar00:
@Lach-anonym Have you tried contacting Kalisz authorities/operator? I tried downloading the gtfs from `http://rj.sdip.kalisz.pl/?download=gtfs_kla` (the link listed in transitous as well as national access point) and the result is _intriguing_ at best. The endpoint seems to respond with file download headers(`Content-Disposition: attachment; filename="kalisz_gtfs_kla.zip"`) but instead sends a HTML with embeded(?) zip (judging from random strings like `fare_rules.txt` inside other unredeable garbage):
image

Lach-anonym:
I've just tried to download it again and the result was... a corrupt zip file, different from what was before
After opening as text I get the exact same result as posted above
I can call them tomorrow about that

kasmar00:
Thanks, In the mean time: I tried googling "fixing corrupt zip files", and it seems that `zip -FF  kalisz_gtfs_kla.zip -out new.zip` works and produces a valid zip (not yet sure about valid GTFS) 🤯 I can try to automate this and proxy it on my domain.

Lach-anonym:
Indeed, that also will do! I was afraid the (wannabe) HTML formatting will not make it possible
Thanks a lot for the quick fix!

kasmar00:
I repacked the feeds at:
- https://gtfs.kasznia.net/static/kalisz.zip
- https://gtfs.kasznia.net/static/kalisz-others.zip
For PKS I think we can temporailirly use the latest file from the website, as the link doesn't seem to change too often. I can also try proxying on my side.

Lach-anonym:
Excellent, thank you very much!

kasmar00:
Quick check: all the failures are unrelated to the current PR (couple outdated feeds, I'll find time to look into them 😅)

Funny thing with the PKS feed: I can't download it from my VPS (403 error and html error page), but can from my home server. If we run into any issues with that in the daily import, I can try to proxy it somehow.

jbruechert:
Thanks!

Lach-anonym:
The KLA city buses do not show in Transitous, any ideas for the reason?

jbruechert:
Have a look at the processed files.

Lach-anonym:
I did and the format seems to be good for me

Lach-anonym:
Okay, so I think the issue could be with missing feed validity end date

Lach-anonym:
Screenshot_20251028_205502_Vivaldi

That is also concerning

jbruechert:
That looks like potentially broken CSV quoting.

kasmar00:
I replaced the agency.txt file in my feed. It should work correctly on next reload. Unfortunately, it's not future proofed for new operators 😢. I plan to add some monitoring (maybe automatically running https://gtfs-validator.mobilitydata.org/) to my feeds, that should help spot the issue of missing operators.

Lach-anonym:
I can confirm that some new local operators have appeared (and that fixed the issue with checking-in in Träwelling lmao), however the KLA city bus trips are still not fetched for some reason

kasmar00:
> however the KLA city bus trips are still not fetched for some reason

I made the fix in feed info, to include an end date (in retrospect, I can probably remove both start and end dates completely), but forgot to pull it on VPS 😅😅😅😅😅😅 It should regenerate and import tonight.

kasmar00:
The Kalisz feed is fetched properly, but first service day is 10th November: https://stats.motis-project.org/d/3f6bfbe5-1b43-4806-a242-c9e781346d3b/transitous-timetables?orgId=1&from=now-2d&to=now&timezone=browser&var-feed=pl-KLA-Kalisz&var-agency=$__all&var-instance=transitous-motis&viewPanel=panel-22
I don't understand why, the calendar.txt file looks alright:
```
service_id,monday,tuesday,wednesday,thursday,friday,saturday,sunday,start_date,end_date
SOB,0,0,0,0,0,1,0,20251103,29990101
SZK,1,1,1,1,1,0,0,20251103,29990101
WAK,0,0,0,0,0,0,0,20251103,29990101
NIE,0,0,0,0,0,0,1,20251103,29990101
```
and calendar_dates.txt contains exceptions only for dates higher than 2025-11-11:
```
service_id,date,exception_type
NIE,20251111,1
SZK,20251111,2
WAK,20251222,1
SZK,20251222,2
WAK,20251223,1
SZK,20251223,2
NIE,20251224,1
SZK,20251224,2
NIE,20251225,1
SZK,20251225,2
(abbriv)
```

Lach-anonym:
so the KLA feed worked on 11.11, but PKS didn't
now PKS works and KLA is gone
local buses work all the time
this is so cursed

kasmar00:
I think, with KLA the issue may be high `end_date`. For some reason transitous displays "last active 01.01.1975" - seems like some kind of integer overflow. I'll try to replace the end date with something more sensible, as I doubt they really plan to operate this timetable in a thousand years 😅

kasmar00:
KLA works on my machine! So it should be loaded by transitous correctly tonight.
image

PR #1458, opened 2025-08-22
gb: Switch to aubin gtfs feed

jbruechert:
This one has more details, and should be updated quicker than the one we currently use

**CC**: @MrKrisKrisu, @HerrLevin, this might need changes in your license allowlist, and would probably cause complaints if it breaks.


jbruechert:
Looks like the GitHub Actions runner might have OOMed. It worked locally, so should be fine.

PR #1459, opened 2025-08-22
Add Intercity RT and more zbiorkom.live feeds

kasmar00:
Adds:
- Realtime feed for PKP Intercity - generated based on zbiorkom.live data (trip updates and alerts). I had my own instance running with it for a few days and it looks alright.
- SKM Warszawa Feed (_Warsaw Sbahn_) - it was previously in `Warszawa` Feed but was moved to own feed by zbiorkom.live maintainer
- Grodziskie Przewozy Autobusowe - new feed from zbiorkom.live

### DIFF
--- a/feeds/pl.json
+++ b/feeds/pl.json
@@ -20,6 +20,15 @@
             "transitland-atlas-id": "f-pkp~intercity~pl"
         },
         {
+            "name": "PKP-Intercity",
+            "type": "url",
+            "spec": "gtfs-rt",
+            "url": "https://gtfs.kasznia.net/rt/pkpic.pb",
+            "license": {
+                "spdx-identifier": "CC-BY-4.0"
+            }
+        },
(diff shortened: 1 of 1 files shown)


jbruechert:
Just to clarify, why is it necessary to generate a feed based on zbiorkom.live? We already have some of their gtfs-rt feeds included directly.

Lach-anonym:
Zbiorkom doesn't use GTFS-RT format for railway, so this is a conversion

kasmar00:
Yep, they don't provide a RT feed for railways and it's not on their roadmap (and from the discussion with the maintainer it wouldn't be easy on their side). The feed I created is a conversion of their processed open data into a RT feed.
If this one works smoothly, I plan to make a similar conversion for other operators.

jbruechert:
Thanks!

PR #1471, opened 2025-08-27
nz: add Flamingo GBFS feeds and replace Metlink key

applecuckoo:
courtesy ping to @Hoverth as your Metlink key isn't working

as the title states, this PR adds the GBFS feeds for New Zealand's [Flamingo](https://www.flamingo.co.nz/) and replaces the Metlink key with my own


Hoverth (approved):
The API key thing is strange, as I haven't been notified about any issues, but thanks for fixing it!

This all looks good to me, apart from maybe the submodule being updated, but one of the repo maintainers will have to check that out.

jbruechert:
The submodule update doesn't really make a difference in our setup, but it isn't making anything worse either.

jbruechert:
Thanks!

PR #1473, opened 2025-08-29
nz: fix Flamingo naming

applecuckoo:
self explanatory, fixes the naming of the Flamingo feeds so that all of them actually show up

### DIFF
--- a/feeds/nz.json
+++ b/feeds/nz.json
@@ -22,37 +22,37 @@
             "url": "https://jbb.ghsq.de/gtfs/at/legacy"
         },
                 {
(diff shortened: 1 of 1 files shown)


feeds/nz.json
```diff
@@ -22,37 +22,37 @@
             "url": "https://jbb.ghsq.de/gtfs/at/legacy"
         },
                 {
-            "name": "Flamingo",
+            "name": "Flamingo Auckland",
```
mlundblad: The feed names can't include spaces.

So, something like "Flamingo-Auckland" and so on…

PR #1478, opened 2025-09-01
au: add Tasmania and Transperth feeds

applecuckoo:
self explanatory, also add myself as a co-maintainer of the feed file

Transperth is now being added through the Mobility Database instead of being added directly.


mlundblad:
Thanks!

PR #1484, opened 2025-09-02
Update proxy to new address

Fale:
Old address can stay online for a while if needed, but the new deployment is easier for me to manage, so I'd like to move to that one :)


jbruechert:
Thanks! I don't think we require the old address to keep working.

Fale:
It seems like the last calls in were yesterday at 11 AM UTC. I'll wait for a ~72hrs without requests, just to be sure, and then put it offline, then.
Thanks :)

PR #1498, opened 2025-09-11
Add GTFS for RegioJet - trains in Poland

kasmar00:
This feed contains RegioJet routes in Poland. Routes in Czechia are already in CZ national feed.

### DIFF
--- a/feeds/pl.json
+++ b/feeds/pl.json
@@ -704,6 +704,18 @@
             }
         },
         {
+            "name": "RegioJet",
+            "type": "http",
+            "spec": "gtfs",
+            "url": "https://gtfs.kasznia.net/static/regio-jet.zip",
+            "license": {
+                "spdx-identifier": "CC-BY-4.0"
+            },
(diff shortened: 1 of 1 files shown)


jbruechert:
Thanks!

Lach-anonym:
Is it possible to change vehicle type IDs, so for example the Przemyśl-Praha train will be a night train? It's to make the transport filtering in Transitous useful

mlundblad:
> Is it possible to change vehicle type IDs, so for example the Przemyśl-Praha train will be a night train? It's to make the transport filtering in Transitous useful

This might be possible to do when the support for the import scripts lands in Transitous, I think?

kasmar00:
Hi, unfortunately the library I'm using (impuls.readthedocs.io) doesn't support extended route types. I'm considering making a contribution to support it, but it's not on my priorities list yet.
In the meantime the lua scripts could help. I can write such a script once the support lands.

PR #1516, opened 2025-09-19
Add Novi Sad (RS) with realtime

vekejsn:


vekejsn:
As soon as the GTFS rebuilds it shouldn't error any more.

mlundblad:
Thanks!

PR #1519, opened 2025-09-19
Add license for JDF merged and remove Bean Shuttle

Lach-anonym:
I've added the CC0 license to JDF merged feed (same reason and license as for CZPTT) and removed Bean Shuttle, as it does not hold a license for regular lines and also destinations and origin points are in fact not fixed... it's just a shuttle minibus!

### DIFF
--- a/feeds/cz.json
+++ b/feeds/cz.json
@@ -65,11 +65,6 @@
             "fix": true
         },
         {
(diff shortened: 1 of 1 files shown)


mlundblad:
Thanks!

PR #1526, opened 2025-09-21
Add unofficial intercity feeds for Serbia

vekejsn:
Adds feeds for Litas, Autoprevoz Čačak, Kavim Jedinstvo and Pantransport.

### DIFF
--- a/feeds/rs.json
+++ b/feeds/rs.json
@@ -78,12 +78,12 @@
             "url": "https://gitlab.com/vekejsn/gtfs-generators/-/jobs/artifacts/main/raw/pancevo_gtfs.zip?job=pancevo"
         },
         {
-            "name": "Autoprevoz_Cacak",
+            "name": "Autoprevoz-Cacak",
             "type": "http",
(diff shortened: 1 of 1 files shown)


mlundblad:
Thanks!

jbruechert:
The limits for the fast trip filter are fairly generous already, are these trips really correct / is that option really needed?

vekejsn:
In the case of the Pantransport feed, it would drop around 10% of all trips, which I'm not sure if it's great, for AP Čačak it would drop half of all trips, in the case of Litas and Jedinstvo only one trip gets dropped, though I'm not sure if that's great as well - since the only international route Litas operates has one leg drop out, similarly for Jedinstvo - the only trip on the route to Kragujevac gets dropped .

jbruechert:
The limit for busses is 200km/h, so if that is the case something must be wrong with stop times or stop coordinates.

I'm worried that right now these trips could create shortcuts in the routing which would always be chosen over other trips, but aren't realistic.

PR #1532, opened 2025-09-22
Add new Polish feeds

Lach-anonym:
Sadly, for now the line 500 in PKS Nova will be duplicated due to lack of real time for the rest of the main feed


jbruechert:
Thanks!

PR #1535, opened 2025-09-22
Create scripts/de-DELFI.lua

felixguendling:


scripts/de-DELFI.lua
```diff
@@ -0,0 +1,39 @@
...
+function remove_leading_zeros(str)
+  return string.format("%d", tonumber(str))
+end
+
+function process_location(location)
+  if location:get_name() then
+    local name = location:get_name()
+    local suffix = " (Berlin)"
+    if #name > #suffix and name:sub(-#suffix) == suffix then
+      -- Remove suffic " (Berlin)" for VBB stops
```
vkrause: The DELFI feed here has VBB filtered out, is this still necessary then? Or does this need to (also) be applied to the VBB feed?

felixguendling: You're right. Probably not needed. Anyway, it can be added back later if needed. Trip name formatting is way more important to get in.

PR #1536, opened 2025-09-23
Update cl.json: Add Red Movilidad feed

Hattshire:


mlundblad:
If you don't mind, you could also add yourself to the maintainers section in the feed file.

Also, I noticed that the URL of the feed added seems to contain a date, would this mean that this would need to be updated later?

mlundblad:
This is now covered by #1590

PR #1542, opened 2025-09-24
Add realtime feeds for Spain, Portugal and Romania

vekejsn:


jbruechert:
Thanks!

dancojocaru2000:
Out of curiosity, where are those feeds from? The ones for Romania and Portugal seem to be a website without any details about where the data is from or such.

Not to seem confrontational, I'm just curious because the data for Romania seems to currently not be completely accurate, and I'm not sure where to go to try fixing that.

vekejsn:
Portugal got blocked quite soon after the generator was set up; Romania I think takes the data from one of the maps from CFR (proxied). The repository for the generator is here: https://gitlab.com/vekejsn/gtfs-rt-exporters

dancojocaru2000:
Thanks for letting me know!

I'll look into how the Romania one can be improved, because the map is missing some details, such as cancelled trains or replacement buses. But great work parsing the data from the map!

PR #1544, opened 2025-09-24
Add script for Trafiklab

mlundblad:
- Re-write route_type for tourist railways
- Add route_short_name on some routes only having long names, which actually have short names (numbers) in real life
- Map route colors for a few agencies/routes

### DIFF
--- a/feeds/se.json
+++ b/feeds/se.json
@@ -3,6 +3,10 @@
         {
             "name": "Mathis Brüchert",
             "github": "mbruchert"
+        },
+        {
+            "name": "Marcus Lundblad",
+            "github": "mlundblad"
         }
     ],
     "sources": [
@@ -12,7 +16,8 @@
             "transitland-atlas-id": "f-u-se",
             "drop-agency-names": [
                 "Flixbus"
-            ]
+            ],
+            "script": "se-Trafiklab.lua"
         }
     ]
 }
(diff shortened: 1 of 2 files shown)


jbruechert:
How sure can we be that these route ids are stable?

mlundblad:
> How sure can we be that these route ids are stable?

Yeah, I was thinking a bit about that as well.

But I think these IDs are maybe quite stable after all.
It looks like the after the first "1" there is the agency ID (as far as I recall these have been the same for probably 10 years in Trafiklab's data (74 for SJ, 275 for SL …). I think those are Rikstrafikens internal codes.
Then it looks to be the line number (display name) zero padded for most of these (except the commuter trains, which _do_ have line numbers (40, 41, 42X, 43, and 48), but I didn't see an obvious pattern, and there's only four of these routes in routes.txt. So I skipped trying to map route_short_name there. And the last part I take it is "serial number" to give uniqueness (in some agencies the same display name is used on different lines, like for Västtrafik where "1" can be a tram line in Göteborg, a bus in Borås and so on.
In SL and UL the line numbers are unique (although in SL there can be variants with a letter, I guess those get a different numeric "serial number" suffix).

So, not sure. Maybe we could try and in worst case revert this part?

The first mapping (agency and route_type → route_type) I think should be fairly stable.

PR #1546, opened 2025-09-25
Add RT feeds for PL train operators

kasmar00:
Add RT feeds for polish train operators. Data is sourced from zbiorkom.live as in #1459.
I've been running those for the past few days on my local instance with no major issues and pretty good RT availability (above 97% for most of the time, with the exception of "Łódzka Kolej Aglomeracyjna" which contains shuttle busses for which no RT data is available at the moment).

### DIFF
--- a/feeds/pl.json
+++ b/feeds/pl.json
@@ -43,6 +43,15 @@
             "fix": true
         },
         {
+            "name": "PolRegio",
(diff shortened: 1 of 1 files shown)


jbruechert:
Nice, thanks!

PR #1557, opened 2025-09-28
Treat files.girlc.at as main urls

kasmar00:
Zbiorkom.live acts as a cache, not as publisher (something similar to how https://api.transitous.org/gtfs/ works), and may not have the most current feed versions.
(Also, this unifies the approach to feeds available via both zbiorkm.live and girlc.at, as some feeds had this double url, and some didn't).

### DIFF
--- a/feeds/pl.json
+++ b/feeds/pl.json
@@ -442,8 +442,7 @@
             "name": "Wschód-Express",
             "type": "http",
             "spec": "gtfs",
-            "url": "https://cdn.zbiorkom.live/gtfs/bialystok-wschod.zip",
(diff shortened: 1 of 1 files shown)


jbruechert:
Looks good to me. It might be worth asking the maintainer of girlc.at what they prefer.

Lach-anonym:
Afair the maintainer recommends using the girlc.at urls

kasmar00:
> It might be worth asking the maintainer of girlc.at what they prefer.

It was the maintainer that asked me for this change, so I can say that's what they prefer 😃

jbruechert:
Thanks!

PR #1563, opened 2025-09-29
Add Italy feeds

Lach-anonym:
Also added license to one feed
Riviera Trasporti has the calendar extension function, I wonder if it will work

### DIFF
--- a/feeds/it.json
+++ b/feeds/it.json
@@ -112,7 +112,7 @@
             "license": {
                 "spdx-identifier": "CC-BY-4.0"
             },
-            "url": "htps://usbshare.agenziatpl.it/index.php/s/7PrFyAHRrqnxKFj/download/2024-02-14_PMT.zip",
+            "url": "https://usbshare.agenziatpl.it/index.php/s/7PrFyAHRrqnxKFj/download/2024-02-14_PMT.zip",
             "skip": true,
             "skip-reason": "Expired feed"
         },
--- a/feeds/it.json
+++ b/feeds/it.json
@@ -12,7 +12,8 @@
             "mdb-id": "mdb-2668",
             "fix": true,
             "license": {
-                "url":"https://www.dati.gov.it/content/italian-open-data-license-v20"
+                "url":"https://www.dati.gov.it/content/italian-open-data-license-v20",
+                "spdx-identifier": "IODL-2.0"
             }
         },
         {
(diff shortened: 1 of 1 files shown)


feeds/it.json
```diff
@@ -66,6 +67,14 @@
...
             "fix": true
         },
+        {
+            "name": "Liguria-Riviera",
+            "type": "http",
+            "url": "https://srvcarto.regione.liguria.it/dtuff/download_statico/opendata/trasporti/GTFS/GTFS-IT-ITC3-RT-20241001-20241231-pf-fares.zip",
+            "license": {
+                "spdx-identifier": "CC-BY-4.0"
+            },
+            "extend-calendar": true
```
jbruechert: This won't work, the feed doesn't use calendar.txt, so there are no patterns to extend.

Lach-anonym: Oh, I see

feeds/it.json
```diff
@@ -12,7 +12,8 @@
             "mdb-id": "mdb-2668",
             "fix": true,
             "license": {
-                "url":"https://www.dati.gov.it/content/italian-open-data-license-v20"
+                "url":"https://www.dati.gov.it/content/italian-open-data-license-v20",
+                "spdx-identifier": "IODL-2.0"
```
jbruechert:
```suggestion
                "spdx-identifier": "LicenseRef-IODL-2.0"
```
There is no standardised identifier for this license, this should work I think

feeds/it.json
```diff
@@ -12,7 +12,8 @@
             "mdb-id": "mdb-2668",
             "fix": true,
             "license": {
-                "url":"https://www.dati.gov.it/content/italian-open-data-license-v20"
+                "url":"https://www.dati.gov.it/content/italian-open-data-license-v20",
+                "spdx-identifier": "LicenseRef-IODL-2.0"
             }
```
jbruechert: seems like id does not work, in this case please just remove the spdx-identifier because there is none.
```suggestion
                "url":"https://www.dati.gov.it/content/italian-open-data-license-v20"
            }
```

feeds/it.json
```diff
@@ -78,6 +79,15 @@
...
             "fix": true
         },
+        {
+            "name": "Lombardia-Autolinee-Varesine",
+            "type": "mobility-database",
+            "mdb-id": "mdb-1059",
+            "license": {
+                "url":"https://www.dati.gov.it/content/italian-open-data-license-v20",
+                "spdx-identifier": "IODL-2.0"
+            }
```
jbruechert, line 87:
```suggestion
            "license": {
                "url":"https://www.dati.gov.it/content/italian-open-data-license-v20"
            }
```

feeds/it.json
Lach-anonym: That's weird, because I've added the IODL-2.0 identifier to Träwelling

jbruechert:
Thanks!

feeds/it.json
jbruechert: They probably don't validate the identifiers. I can't find this one on spdx.org

Lach-anonym: Yeah, it's not on SPDX, as it's a custom identifier made by Italians, as seen on the license website

jbruechert: SPDX Identifiers are only the standardized set of licenses. Which is fine really, it can be identified by the url as well.

PR #1566, opened 2025-09-29
Skip GTFS-RT for Czech railways

Lach-anonym:
The only results I've seen for Czech railway were trains on time (often falsely) or without data, so I'suggest to skip the feed until it's fixed


mlundblad:
Thanks!

PR #1567, opened 2025-09-30
Add Metro for Porto

SteilDev:
Hey,
The Metro for Porto is somehow not available as static link, I created a repo which pulls the latest GTFS file from the website, and runs gtfstidy over it. The latest gtfs file is then uploaded to the repo.

This PR adds the (static) link to the latest gtfs file.

### DIFF
--- a/feeds/pt.json
+++ b/feeds/pt.json
@@ -147,11 +147,11 @@
         {
             "name": "Metro-Porto",
             "type": "http",
-            "url": "https://github.com/SteilDev/metro-porto-gtfs-latest/raw/refs/heads/main/gtfs_metroporto_latest.zip",
+            "url": "https://www.metrodoporto.pt/pages/337",
+            "function": "data_metroporto_latest_resource",
             "license": {
-            "spdx-identifier": "CC0-1.0",
-                "url": "https://creativecommons.org/publicdomain/zero/1.0/"
-              }
+                "spdx-identifier": "CC0-1.0"
+            }
         }
     ]
 }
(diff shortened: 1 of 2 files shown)


feeds/pt.json
```diff
@@ -139,6 +143,15 @@
...
             "type": "transitland-atlas",
             "transitland-atlas-id": "f-ey9-aonda"
+        },
+        {
+            "name": "Metro-Porto",
+            "type": "http",
+            "url": "https://github.com/SteilDev/metro-porto-gtfs-latest/raw/refs/heads/main/gtfs_metroporto_latest.zip",
+            "license": {
+            "spdx-identifier": "CC0-1.0",
+                "url": "https://creativecommons.org/publicdomain/zero/1.0/"
```
jbruechert:
```suggestion
```
Unless the link gives some context for the license, like a source that it applies to this feed, the spdx-identifier is enough to automatically link to the license text.

jbruechert:
> Hey, The Metro for Porto is somehow not available as static link, I created a repo which pulls the latest GTFS file from the website, and runs gtfstidy over it. The latest gtfs file is then uploaded to the repo.

Hi, thanks for looking into this.
I think this problem can already be solved with the limited python scripting feature in the Transitous repository.

For example https://github.com/public-transport/transitous/blob/main/feeds/lu.json#L20 calls https://github.com/public-transport/transitous/blob/main/src/region_helpers.py#L30 to insert the link to the latest gtfs file.
You can probably use almost exactly the same code you are already using with this, but without the additional delay of the mirroring and without depending on an external repository.

SteilDev:
Hey, thanks for the review and hint with the region helpers file. I changed it, it now calls the method to pull the link to the latest GTFS file.

jbruechert:
Thanks!

PR #1570, opened 2025-10-01
Drop Eurostar services from the DELFI feed

vkrause:
We have a dedicated feed for Eurostar that has shapes, colors and at least partial realtime data.


Altonss:
Somehow, some THALYS (EUROSTAR) services from DELFI still seem visible :thinking:

vkrause:
> Somehow, some THALYS (EUROSTAR) services from DELFI still seem visible 🤔

Right. I guess that's because the filter will only take effect when the next DELFI feed update is downloaded (unlike script changes that apply on every import).

PR #1572, opened 2025-10-02
au: NSW feed refactor

applecuckoo:
cc @daviewales, you might be interested in this

This PR adds the Flex and RT data available from Transport for New South Wales (TfNSW). There are a few things that are suboptimal right now, so for now this is a draft.

The realtime feeds are paired with their *own* static feed which isn't compatible with the 'complete' feed that we currently have, so we pull a lookup table [they provide](https://opendata.transport.nsw.gov.au/data/dataset/reference-tables-tfnsw-gtfs-feeds) and add all the affected agency names into the feed file.

The other problem is that some of the timetables have a finer scope than the RT endpoints, which means Transitous would be redundantly pulling the same endpoints multiple times and thus using up more requests on my API key than we actually need. For example, there is an alerts endpoint that covers all modes of transport, while the feeds for buses are split up by operator. (note: There is a combined buses endpoint mentioned in the OpenAPI schema, but it is broken and unsupported.)

The NSW Trains and Sydney Trains also have overlapping data, and the NSW Trains feed should take priority over the Sydney Trains feed in these cases.

checklist of feeds/tasks:

- [x] 'complete' feed
    - [x] dedupe using table
- [x] flex data
- [x] Sydney Metro
- [x] Sydney Trains
- [x] Ferries
- [x] NSW Trains
- [x] Light rail
- [x] Buses
    - [x] Regional buses
- [x] generator script?

note: all commits will be squashed when ready for merging

### DIFF
--- a/src/generate-australia-new-south-wales.py
+++ b/src/generate-australia-new-south-wales.py
@@ -1,0 +1,37 @@
+#!/usr/bin/env python3
+# SPDX-FileCopyrightText: applecuckoo <nufjoysb@duck.com>
+#
+# SPDX-License-Identifier: AGPL-3.0-or-later
+#
+# generate-australia-new-south-wales.py - blocklist generator for TfNSW 'complete' feed
+
+import csv
+from io import StringIO
+import json
+import requests
+
+def clean_text(text) -> str:
+    return text.replace('\u0092', '').replace('\u00a0', '')
+
+# pull resource ID from TfNSW CKAN instance and get latest URL
+ckan_url = 'https://opendata.transport.nsw.gov.au/api/3/action/resource_show?id=30b850b7-f439-4e30-8072-e07ef62a2a36'
+ckan = requests.get(ckan_url).json()
+lookup_table_url = ckan['result']['url']
+lookup_table = requests.get(lookup_table_url)
+lookup_table.encoding = 'iso-8859-1'
+lookup_table = clean_text(lookup_table.text)
+lookup = csv.DictReader(StringIO(lookup_table.split('\n', 1)[1]))
+
+blocklist = []
+for rows in lookup:
+    rt_agency = rows['For Realtime GTFS agency_name'].strip()
+    if rt_agency != '' and rt_agency not in blocklist:
+        blocklist.append(rows['Complete GTFS agency_name'].strip())
+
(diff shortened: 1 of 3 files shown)


daviewales:
Thanks @applecuckoo, I am interested! That's an impressive list of feeds.
I previously tried to get some of the real-time feeds [added to the Mobility database](https://mobilitydatabase.org/feeds?q=sydney&gtfs_rt=true), but I'm pretty sure I got the URLs wrong now that I've seen your list.

As a quick sense-check I've noticed that in this PR there are a bunch of feeds listed with different names, but the same URLs, for the following URLs:

- https://api.transport.nsw.gov.au/v1/gtfs/realtime/buses
- https://api.transport.nsw.gov.au/v2/gtfs/alerts/all

Is that expected?

applecuckoo:
> Is that expected?

yep, all the non-regional buses share the same trip update endpoint. Using `alerts/all` was intentional as well, but after some thought I've switched back to using the 'per-mode' endpoints since that would consume less data than just calling `alerts/all`. Ultimately, I'd like to use that endpoint if MOTIS knows to 'share' it for all the feeds though.

src/generate-australia-new-south-wales.py
```diff
@@ -0,0 +1,37 @@
...
+import requests
+
+# pull resource ID from TfNSW CKAN instance and get latest URL
+ckan_url = 'https://opendata.transport.nsw.gov.au/api/3/action/resource_show?id=30b850b7-f439-4e30-8072-e07ef62a2a36'
+ckan = requests.get(ckan_url).json()
+lookup_table_url = ckan['result']['url']
+lookup_table = requests.get(lookup_table_url)
+lookup_table.encoding = 'iso-8859-1'
+# remove random u0092 characters
+lookup_table = lookup_table.text.replace('', '')
```
mlundblad: Would it be possible/make sense to explictly use a \u sequence to express u0092 here? (My Python-foo isn't that big…)

Maybe that would make this a bit easier to understand (and avoid some accidental change of this replace string in the future).

applecuckoo: @mlundblad done now along with a few other formatting fixes, should be good to merge

mlundblad:
Thanks!

PR #1573, opened 2025-10-02
Egypt - Cairo

SteilDev:
Hey,
I found this website [[1](https://data.transportforcairo.com/#/)] which publishes several GTFS datasets, some of which (the two for Cairo that I included) are updated frequently [[2](https://data.transportforcairo.com/#/)].
From what I read ([[3](https://data.transportforcairo.com/about/)] "Each dataset in the system can be shared publicly or restricted to allow access to only specific users."), there should not be a problem of using this dataset here.

The other datasets are outdated, hence i did not include them.

### DIFF
--- a/feeds/eg.json
+++ b/feeds/eg.json
@@ -1,0 +1,28 @@
+{
+    "maintainers": [
+        {
+            "name": "Patrick Steil",
+            "github": "SteilDev"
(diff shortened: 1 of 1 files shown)


jbruechert:
Thanks!

PR #1574, opened 2025-10-02
Fix PKP Intercity script

kasmar00:
Script was added to wrong entity (RT instead of static)


mlundblad:
I think the CI failure is not related to this change, so I think it can be merged.

mlundblad:
Thanks!

PR #1578, opened 2025-10-03
Add Lua script to fix Austrian route types

vkrause:


jbruechert:
at.json is generated by a script, so that needs to be adapted as well.

vkrause:
> at.json is generated by a script, so that needs to be adapted as well.

Ah, I had missed that, done now.

jbruechert:
Thanks!

PR #1584, opened 2025-10-04
au-vic/nz: misc. formatting scripts

applecuckoo:


scripts/nz-AT.lua
```diff
@@ -0,0 +1,11 @@
+-- SPDX-FileCopyrightText: applecuckoo <nufjoysb@duck.com>
+-- SPDX-License-Identifier: AGPL-3.0-or-later
+
+require "scripts.motis"
+
+function process_route(route)
+    -- set all routes with the RB (Rail Bus) prefix as rail replacement buses
+    if string.find(route:get_short_name(), "RB") then
```
mlundblad, line 8: I guess this could use some substring function to get the prefix to make it a bit more robust? Right now it would match anything containing "RB", but maybe that is unlikely to give "false positives"?

applecuckoo: this should be fine, the string 'RB' doesn't appear in any of the other routes, only the rail replacement buses

mlundblad: ok!

mlundblad:
Thanks!

PR #1585, opened 2025-10-04
Add Austrian GTFS Flex feeds

vkrause:
Those were previously explicitly skipped in the generation script, from a time when we couldn't handle Flex yet.

### DIFF
--- a/feeds/at.json
+++ b/feeds/at.json
@@ -15,6 +15,84 @@
     ],
     "sources": [
         {
+            "name": "PTA-Eastern-Region-Flex-2025",
+            "type": "http",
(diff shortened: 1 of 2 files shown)


mlundblad:
Thanks!

jbruechert:
Hmm, as far as I remember the flex variants are a superset of the non-flex feeds. We should probably ignore these now.

vkrause:
> Hmm, as far as I remember the flex variants are a superset of the non-flex feeds. We should probably ignore these now.

Indeed, after having a closer look that seems to be a full subset (I had originally been confused by the files being ordered and formatted very differently and came to the opposite conclusion). I'll fix that.

vkrause:
#1588

PR #1587, opened 2025-10-04
Fix display names on most of the international trips in the DELFI feed

vkrause:
What this doesn't cover yet are trips that are modeled as split trips with two different operators.


mlundblad:
Thanks!

PR #1590, opened 2025-10-05
Add Chile

SteilDev:
I stumbeld across [[1]](https://hub.tumidata.org/en/dataset/gtfs-santiago-de-chile), which listed this (official) DTP GTFS provider [[2]](https://www.dtpm.cl/index.php/noticias/gtfs-vigente) as original source.
Again, no stable link to the latest GTFS file available, hence i wrote a method which downloads the latest feed based on the filename (based on the same starting prefix GTFS_YYYYMMDD). If this fails, it will return the link to the lexicografically largest feed.

### DIFF
--- a/feeds/cl.json
+++ b/feeds/cl.json
@@ -3,6 +3,10 @@
         {
             "name": "Nedžad Beus",
             "github": "vekejsn"
+        },
+        {
+            "name": "Patrick Steil",
+            "github": "SteilDev"
         }
     ],
     "sources": [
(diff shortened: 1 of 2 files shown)


jbruechert:
Thanks!

mlundblad:
Btw, would this cover #1536 as well?

That one specified a static URL to one of these dated URLs.

SteilDev:
Yes, that is from the same source. But (as you already mentioned), the other url points to a GTFS file with a specific date (https://www.dtpm.cl/descargas/gtfs/GTFS_20250906.zip). This PR loads the latest GTFS file from the endpoint, based on the GTFS_YYYYMMDD* format

PR #1591, opened 2025-10-05
Fix RegioJet script

kasmar00:
Fixing the route selector, as `"display-name-options": "copy-trip-names-matching"` creates extra route objects (one per each trip)


jbruechert:
Thanks! If that makes it easier to reason about this, it might make sense to move the display name setting to the script as well.

PR #1593, opened 2025-10-07
au-vic: add new V/Line feeds

applecuckoo:
just got an email about new vehicle position and trip update feeds for the Melbourne V/Line trains (set up as `au-vic-Transport-Victoria-regional-trains`), so this is a quick PR to add this in.


mlundblad:
Thanks!

PR #1596, opened 2025-10-07
Fix typo in Austrian Lua script

vkrause:
The constant for night trains was wrong, resulting in this defaulting to 0 and thus turning night trains into flights.


jbruechert:
Thanks!

PR #1606, opened 2025-10-11
update es.json

jonaes:
add a bunch of licenses as well as some new feeds and rt-feeds if found on the way (hoping I didn't duplicate any...).


jbruechert:
Thanks!

PR #1609, opened 2025-10-12
Regenerate FR

Altonss:
Regenerate FR (+ cleanup generation script)

Edit:
- Bad Zip structure error for feed `83019` : related issue https://github.com/public-transport/gtfsclean/issues/17

### DIFF
--- a/feeds/fr.json
+++ b/feeds/fr.json
@@ -523,8 +523,7 @@
             "license": {
                 "url": "https://transport.data.gouv.fr/datasets/reseau-de-transport-interurbain-mobigo-en-bourgogne-franche-comte",
                 "spdx-identifier": "etalab-2.0"
-            },
-            "skip": true
+            }
         },
         {
             "name": "reseau-de-transport-interurbain-mobigo-en-bourgogne-franche-comte--11082",
@@ -535,8 +534,7 @@
             "license": {
                 "url": "https://transport.data.gouv.fr/datasets/reseau-de-transport-interurbain-mobigo-en-bourgogne-franche-comte",
                 "spdx-identifier": "etalab-2.0"
-            },
-            "skip": true
+            }
         },
         {
             "name": "reseau-de-transport-interurbain-mobigo-en-bourgogne-franche-comte--11083",
@@ -547,8 +545,7 @@
             "license": {
                 "url": "https://transport.data.gouv.fr/datasets/reseau-de-transport-interurbain-mobigo-en-bourgogne-franche-comte",
                 "spdx-identifier": "etalab-2.0"
-            },
-            "skip": true
+            }
         },
         {
             "name": "reseau-de-transport-interurbain-mobigo-en-bourgogne-franche-comte--11084",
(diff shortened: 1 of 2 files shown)


Lach-anonym:
I've checked the Horizon feed recently and I found something interesting...
Screenshot_20251012_124223_MiXplorer

Also I've noticed there are data duplicates for example in Nantes and Bretagne, but surely there is more to discover

Altonss:
> I've checked the Horizon feed recently and I found something interesting...

Indeed thanks for noticing! It would be really nice to have a `fix-agency-timezone` option (for missing or wrong data).

Altonss:
> Also I've noticed there are data duplicates for example in Nantes and Bretagne, but surely there is more to discover

For the duplicates we could remove some agencies from this feed https://transport.data.gouv.fr/datasets/arrets-horaires-et-circuits-des-lignes-de-transports-en-commun-en-pays-de-la-loire-gtfs-destineo-reseaux-aom-aleop-1 (like Naolib) : the "original" Naolib feed seems to have better shapes .

Lach-anonym:
Indeed!

I guess the mismatched shapes are because the regional feed is republishing the network feeds, there was indeed an issue with the coords at Naolib some days ago and now it's fixed, however region still has the old data.

How about the real time data matching after filtering the agencies? I see there is a SIRI feed for all the networks in PdL, didn't look into it, but it might help with networks that do not publish their RT feeds or their data is wrong

Lach-anonym:
Can this be added to the script, as we support SIRI-Lite? Haven't tested it
https://www.data.gouv.fr/datasets/blablacar-bus-horaires-theoriques-et-temps-reel-du-reseau-europeen/?resource_id=66569792-afb9-4464-a52a-33e3fbb0df3e

Lach-anonym:
PdL region started to aggregate the data from Nantes airport and they really cooked...

https://api.transitous.org/?stopArriveBy=false&stopId=fr-arrets-horaires-et-circuits-des-lignes-de-transports-en-commun-en-pays-de-la-loire-gtfs-destineo-reseaux-aom-aleop-1_AEROPORT_NANTES%3AQuay%3A1400000&time=2025-10-23T18%3A15%3A00.000Z

Screenshot_20251022_235944_Vivaldi

Those flights actually exist, but it's impossible to buy tickets for them, so I doubt they are open to public
https://www.flightradar24.com/airport/snr/arrivals

Screenshot_20251022_235957_Vivaldi
Screenshot_20251023_000047_Vivaldi

felixguendling:
> Can this be added to the script, as we support SIRI-Lite? Haven't tested it https://www.data.gouv.fr/datasets/blablacar-bus-horaires-theoriques-et-temps-reel-du-reseau-europeen/?resource_id=66569792-afb9-4464-a52a-33e3fbb0df3e

It's JSON. We currently only support XML. Otherwise it looks quite similar.

felixguendling:
I added SIRI light support. It requires MOTIS v2.7 and a way to generate `protocol: siri_json` from Transitous' Python scripts.

PR #1612, opened 2025-10-12
Fix MVV Tram 12

### LINKED ISSUE #1489: Line 12 in München is marked as Bus in API
When I search for departures at [Leonrodplatz, Munchen ](https://api.transitous.org/api/v1/stoptimes?stopId=de-DELFI_de:09162:11&n=30) I see line 12 marked as Bus, but its a tram. I don't have a date in the URL because this is on every day.

<details>
<summary> One of the results in the API from line 12: </summary>
<pre>
{
      "place": {
        "name": "Leonrodplatz",
        "stopId": "de-DELFI_de:09162:11:2:2",
        "lat": 48.159184,
        "lon": 11.546972,
        "level": 0,
        "arrival": "2025-09-04T19:59:00Z",
        "departure": "2025-09-04T19:59:00Z",
        "scheduledArrival": "2025-09-04T19:59:00Z",
        "scheduledDeparture": "2025-09-04T19:59:00Z",
        "scheduledTrack": "2",
        "track": "2",
        "description": "Steig 2+5",
        "vertexType": "TRANSIT",
        "pickupType": "NORMAL",
        "dropoffType": "NORMAL",
        "cancelled": false
      },
      "mode": "BUS",
      "realTime": true,
      "headsign": "Schwabing Nord",
      "agencyId": "14747",
      "agencyName": "Straßenbahn München",
      "agencyUrl": "https://www.delfi.de",
      "tripId": "20250904_21:48_de-DELFI_2912717643",
      "routeShortName": "12",
      "pickupDropoffType": "NORMAL",
      "cancelled": false,
      "tripCancelled": false,
      "source": "de_DELFI.gtfs.zip/stop_times.txt:27308907:27308930"
    }
</pre>
</details>

As you can see, the agencyName is Straßenbahn München (Tram) and mode is set on BUS. Is this a mistake in the API or is this something that the MVV needs to change?

Altonss:
Should temporarily fix #1489

### DIFF
--- a/scripts/de-DELFI.lua
+++ b/scripts/de-DELFI.lua
@@ -44,4 +44,9 @@ function process_trip(trip)
   if trip:get_route():get_route_type() == 106 and is_number(trip:get_short_name()) then
     trip:set_display_name(trip:get_route():get_short_name() .. ' (' .. remove_leading_zeros(trip:get_short_name()) .. ')')
   end
+  
+  -- Fix MVV Tram line 12 (was set to bus)
+  if trip:get_route():get_agency():get_name() == 'Straßenbahn München' and trip:get_short_name() == '12' and trip:get_route_type() == 3 then
+    trip:set_route_type(0)
+  end
 end


jbruechert:
This is a bit of a fundamental discussion, but do we really want to fix these kinds of small data errors in scripps? I'm worried that is more likely to create accidental unintended changes than to improve things long-term.

Altonss:
> This is a bit of a fundamental discussion, but do we really want to fix these kinds of small data errors in scripps? I'm worried that is more likely to create accidental unintended changes than to improve things long-term.

Yes I agree with you, those small errors might not be worth fixing in scripts (those would be rather hard to maintain) :+1:

felixguendling:
Maybe it would help if `process_*` could also be an array. Then we could have a statistic in each `set_*` method to see if some functions in the array don't trigger. If they don't trigger they should probably be reviewed if they're still necessary.

scripts/de-DELFI.lua
```diff
@@ -44,4 +44,9 @@ function process_trip(trip)
   if trip:get_route():get_route_type() == 106 and is_number(trip:get_short_name()) then
     trip:set_display_name(trip:get_route():get_short_name() .. ' (' .. remove_leading_zeros(trip:get_short_name()) .. ')')
   end
+  
+  -- Fix MVV Tram line 12 (was set to bus)
+  if trip:get_route():get_agency():get_name() == 'Straßenbahn München' and trip:get_short_name() == '12' and trip:get_route_type() == 3 then
+    trip:set_route_type(0)
```
felixguendling, line 50: you can't set route type on `trip` - this has to happen on `route`

1Maxnet1:
Issue has been resolved see https://github.com/public-transport/transitous/issues/1489

PR #1620, opened 2025-10-17
au: split off feed files by states and territories

applecuckoo:
cc @Steve-Tech @mexon @daviewales, leave a comment if you'd like to help maintain the other states/territories besides the ones you added

just a simple split off for all the states and territories, since currently we don't have any feeds that have an Australia-wide focus


mexon (approved):

mlundblad:
Thanks!

PR #1628, opened 2025-10-19
hu-kaposvar: update source

gy-mate:
https://www.kaposbusz.hu/static/files/Menetrend%20t%C3%A9rk%C3%A9pek/kaposvar(1).zip is https://www.kaposbusz.hu/static/files/Menetrend%20t%C3%A9rk%C3%A9pek/kaposvar(3).zip now.  It seems like they are bumping the version number so it's better to scrape their website to get the current URL.

Also, reformat line breaks.

### DIFF
--- a/feeds/hu.json
+++ b/feeds/hu.json
@@ -76,7 +76,8 @@
         {
             "name": "kaposvár",
             "type": "http",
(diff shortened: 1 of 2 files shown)


gy-mate:
The issue seems to be about _Hódmezővásárhely_, not _Kaposvár_. I'll try to fix that in a separate PR.

jbruechert:
Thanks!

Looks like the CI failure is caused by an unrelated expired feed.

PR #1644, opened 2025-10-26
Feed proxy config

traines-source:
A proposal on how to setup a (GTFSRT/GBFS) feed proxy.
Basic idea: 
* decide for which feeds to use the proxy
* encrypt secrets using a special Ansible vault using e.g. `ansible-vault encrypt_string --vault-id feed-proxy@/etc/feed-proxy.vault 'foobar' --name 'Authorization'` and put them into `ansible/roles/feed-proxy/vars/feed-secrets.yml` to override fields in the same dictionary structure (`url` and `headers`) as in the feed jsons – multiple GTFS-RT feeds are numbered consecutively from 0
* on the proxy machine, setup a cronjob that periodically runs `git pull --ff-only && rm mobilitydatabase.csv && git submodule update --remote && cd ansible && ansible-playbook --vault-id feed-proxy@/etc/feed-proxy.vault -e "feed_proxy_host=proxy.example.com" -e "feed_proxy_conf_dir=/tmp" feed-proxy.yml`.

feed-proxy vault pass can be obtained from `ansible/roles/feed-proxy/vars/feed-secrets-vault-pass.yml` using the main vault password or via PM.

TODO
- [x] should the proxy be ~~enabled~~ or disabled by default for each feed? – for easy changing of the default later on, `use-feed-proxy` can and should be set to either false or true where it is relevant in the feed jsons
- [ ] better way to handle multiple RT feeds with same ID? – put vault-encrypted secrets directly into the respective feed jsons instead? (needs a bit more tinkering and makes the jsons less readable/useful)
- [x] complete nginx setup (currently commented out) if needed
- [ ] maybe instead of running periodically independently, run as part of the import pipeline with a hook?
- [ ] do we need a backwards compat feature? (Currently, as soon as the script is run, the proxy will obviously forget about previously configured feeds/feed ids - might lead to RT outages)
- [x] end-to-end testing
- [x] set actual FEED_PROXY url
- [ ] other propositions?

### DIFF
--- a/feeds/pl.json
+++ b/feeds/pl.json
@@ -367,7 +367,8 @@
             "url": "https://next.zbiorkom.live/api6-open/bydgoszcz/gtfsRealtime/default/tripUpdates",
             "license": {
                 "spdx-identifier": "MIT"
-            }
+            },
+            "use-feed-proxy": true
         },
         {
             "name": "PKP-SKM",
@@ -609,7 +610,8 @@
             "url": "https://next.zbiorkom.live/api6-open/bialystok/gtfsRealtime/WSCHODEXPRESS/tripUpdates",
             "license": {
                 "spdx-identifier": "MIT"
-            }
+            },
+            "use-feed-proxy": true
         },
         {
             "name": "PKS-Nova-Supraśl",
@@ -698,7 +700,8 @@
             "url": "https://next.zbiorkom.live/api6-open/opole/gtfsRealtime/default/tripUpdates",
             "license": {
                 "spdx-identifier": "MIT"
-            }
+            },
+            "use-feed-proxy": true
         },
         {
             "name": "Arriva",


src/generate-motis-config.py
```diff
@@ -152,23 +161,35 @@ def check_file_exist_in_out_folder(file_name: str):
                                     config["timetable"]["datasets"][name]["rt"] = []
 
                                 rt_feed: dict[str, Any] = {
-                                    "url": source.url
+                                    "url": source.url if use_original_url else FEED_PROXY+name+"-"+str(len(config["timetable"]["datasets"][name]["rt"]))
```
jbruechert:
```suggestion
                                    "url": source.url if use_original_url else FEED_PROXY + name + "-" + str(len(config["timetable"]["datasets"][name]["rt"]))
```

src/generate-motis-config.py
```diff
@@ -16,7 +16,9 @@
 from typing import Any
 from pathlib import Path
 from utils import eprint
+from urllib.parse import quote
 
+FEED_PROXY="https://rt.triptix.tech"
```
jbruechert, line 21: Not that it matters super much, but why is this not under the transitous.org domain?

traines-source: The idea was that at some point, as opposed to the *.transitous.org stuff, commercial users can use this as well, for a fee. Not sure how viable this actually is.

PR #1645, opened 2025-10-28
Reactivate Mińsk Mazowiecki

kasmar00:
Feed was extended (turns out the timetable didn't change when it expired 😅)


jbruechert:
Thanks!

Issue #1648, opened 2025-10-29
VRN / RNN GTFS-RT Feed


MichaelsJP:
Hi. The Verkehrsverbund Rhein-Neckar (VRN) released an endpoint for their GTFS-RT data. 

URL: http://www.vrn.de/service/entwickler/gtfs-realtime/
Data Page (German): https://opendata.vrn.de/datasets/vrn-und-rnn-gtfs-realtime-echtzeitfahrplandaten 

Is this feed of interest?

jbruechert:
It is definitely interesting and looks well done, but before we can add it, we need to look into how well we can cut out the overlap from the DELFI feed, and how the quality of the realtime coverage compares between the two.

PR #1649, opened 2025-10-29
Add VRN GTFS-RT Feed

MichaelsJP:
Related PR for #1648.

I guess what is missing is the proper licence file.

Should I just copy the licence text from https://www.govdata.de/dl-de/by-2-0 to a new Licence file?

### DIFF
--- a/feeds/de.json
+++ b/feeds/de.json
@@ -97,6 +97,15 @@
             }
         },
         {
+            "name": "VRN",
+            "type": "url",
+            "spec": "gtfs-rt",
+            "url": "http://www.vrn.de/service/entwickler/gtfs-realtime",
+            "license": {
+                "spdx-identifier": "DL-BY-DE-2.0"
+            }   
+        },
+        {
             "name": "AVV-Aachen",
             "type": "http",
             "url": "https://opendata.avv.de/current_GTFS/AVV_GTFS_Masten_ohne_SPNV_GlobaleID_oFremd.zip",


jbruechert:
> Should I just copy the licence text from https://www.govdata.de/dl-de/by-2-0 to a new Licence file?

There is no need to copy the license file into this repository. We currently only do that for code the licenses of code in this repository, not for the links to data.

feeds/de.json
```diff
@@ -96,6 +96,15 @@
                 "spdx-identifier": "CC-BY-SA-4.0"
             }
         },
+        {
+            "name": "VRN",
+            "type": "url",
+            "spec": "gtfs-rt",
```
jbruechert, line 102: Realtime feeds are always a diff to a static timetable. The fitting one needs to be added as well, and duplicates need to be removed from DELFI using drop-agency-names.

MichaelsJP:
@jbruechert Thank you for the input. Regarding your points:
- I found the related static timetable: 
  - https://opendata.vrn.de/datasets/vrn-und-rnn-gtfs-sollfahrplandaten-aktuell
  - https://geoportal.vrn.de/services/sharing/rest/content/items/4ec4b1d131eb46a6bb8e216ce9b90eff/data
  - I'll add it as well in the PR.
  
Would you mind giving me some tips on how to approach the `drop-agency-names` thing?
I would naively just look for similar names in the agency.txt. But maybe you already have a well established way of doing this.

jbruechert:
> I would naively just look for similar names in the agency.txt. But maybe you already have a well established way of doing this.

That is exactly the way I would approach this as well. But it is important to only remove the local agencies and not some which might operate outside of the VRN area as well.
The DELFI / data directly from the associations combination is always a bit messy.

PR #1652, opened 2025-10-31
Update IDS-JMK RT feed URL & add Centrotrans RT

vekejsn:


feeds/cz.json
```diff
@@ -51,7 +55,7 @@
             "name": "IDS-JMK",
             "type": "url",
             "spec": "gtfs-rt",
-            "url": "https://content.idsjmk.cz/kestazeni/gtfsReal.dat"
+            "url": "https://rt.gtfs.baguette.pirnet.si/gtfs-rt/idsjmk/trip_updates.pb"
```
jbruechert, line 70: What is the difference between the feeds?

vekejsn: The "official" one provides only vehicle positions, the new one uses the websocket service (https://data.brno.cz/datasets/379d2e9a7907460c8ca7fda1f3e84328/about) to get delays with 1min granularity.

jbruechert: In that case, it might be worth keeping both.
MOTIS currently can't display vehicle positions, but it would be nice to already have the feeds once it gains support.

feeds/cz.json
```diff
@@ -40,7 +44,13 @@
             "name": "PID",
             "type": "url",
             "spec": "gtfs-rt",
-            "url": "https://api.golemio.cz/v2/vehiclepositions/gtfsrt/trip_updates.pb"
+            "url": "https://api.golemio.cz/v2/vehiclepositions/gtfsrt/pid_feed.pb"
```
jbruechert, line 47: What is the difference in this case?

Iirc gtfs-rt requires a different feed per message type, which one is this?
Realistically MOTIS can handle mixed ones as well nowadays of course.

vekejsn: This one has combined vehicle positions and trip updates.

PR #1654, opened 2025-10-31
Add Łomianki and Radzymin

kasmar00:


Lach-anonym:
GZM, Kraków Mobilis and Tczew feeds need the fix added
Also why no trips are imported for the KLA feed?

jbruechert:
Thanks!

PR #1662, opened 2025-11-03
readme: add light/dark mode switch for logo

applecuckoo:
self-explanatory, adds the right tags to load the light or dark versions of the Transitous logo depending on your current theme


Altonss (approved):
Looks good to me, works fine on mobile browser :)
Thanks for this improvement!

PR #1665, opened 2025-11-06
Add feeds from girlc.at and fix links

kasmar00:
- adds 2 new feeds from girlc: Płońsk and Żywiec
- for feeds where RT feeds come from zbiorkom.live replace static feeds with zbiorkom.live version (to keep it in sync)
- add Zielona Góra
- Add script to mark Koleje Wielkopolskie trains as Regional Rail

### DIFF
--- a/feeds/pl.json
+++ b/feeds/pl.json
@@ -124,7 +124,8 @@
             "url": "https://cdn.zbiorkom.live/gtfs/kielce.zip",
             "license":{
                 "spdx-identifier": "MIT"
-            }
+            },
+            "fix": true
         },
         {
             "name": "Kielce",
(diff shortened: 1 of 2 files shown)


src/region_helpers.py
```diff
@@ -47,17 +47,14 @@ def delhi_gov_in_csrf(source: HttpSource) -> HttpSource:
     return source
 
 
-def data_mzk_gorzow_latest_resource(source: HttpSource) -> HttpSource:
+def data_zielona_gora_latest_resource(source: HttpSource) -> HttpSource:
```
kasmar00, line 50: No longer used by Gorzów Wielkopolski

feeds/pl.json
```diff
@@ -120,9 +120,11 @@
         },
         {
             "name": "Kielce",
-            "type": "transitland-atlas",
-            "transitland-atlas-id": "f-u3n9-kielce",
-            "fix": true
+            "type": "http",
+            "url": "https://cdn.zbiorkom.live/gtfs/kielce.zip",
+            "license":{
+                "spdx-identifier": "MIT"
```
jbruechert, line 126: The test looks like this one has some broken references as well, like the previous one.

Please keep fix: true

jbruechert (changes requested):

feeds/pl.json
kasmar00: Added.

PR #1666, opened 2025-11-06
ansible: configure nginx log exporter

### LINKED ISSUE #838: Metrics
It would be very interesting to have metrics for the Transitous instance of Motis:
* how many API requests per day / per month / until now
* when are peeks in the amount of requests
* maximum number of parallel requests
* which service is the request for (routing, geocoding, ...)
* what is the usage of the system resources (cpu, ram, disk, bandwidth)

For example [Grafana](https://grafana.com/oss/grafana/) could be used for building a dashboard and [Prometheus](https://prometheus.io/) for collecting the data.

The request based matrics could be derived from the server logs:
* https://github.com/martin-helmich/prometheus-nginxlog-exporter
* https://github.com/nginx/nginx-prometheus-exporter
* https://medium.com/@mykprok/monitor-nginx-metrics-with-grafanadr-a-step-by-step-guide-6a3014c1783f
* https://github.com/nginx/nginx-prometheus-exporter/blob/main/grafana/README.md

For the system resources there are other exporters:
* https://github.com/prometheus/node_exporter

Maybe I could try to implement this myself.

What do you think about this?

traines-source:
for user agent, HTTP status code, response time and endpoint metrics
(already deployed on prod, dashboard is private for the time being)

closes #838

### DIFF
--- a/ansible/roles/nginx/files/prometheus-nginxlog-exporter.hcl
+++ b/ansible/roles/nginx/files/prometheus-nginxlog-exporter.hcl
@@ -1,3 +1,7 @@
+# SPDX-FileCopyrightText: 2025 Traines <git@traines.eu>
+#
+# SPDX-License-Identifier: AGPL-3.0-or-later
+
 listen {
   port    = 4040
   address = "10.11.0.1"


jbruechert:
Thanks :)

PR #1673, opened 2025-11-10
nz: add goBay (Hawke's Bay)

applecuckoo:
this PR should bring our NZ static feed coverage on par with Transit :-)

RT feeds are a different story though... some emailing and/or FOI-ing might be in order for the future


mlundblad:
Thanks!

PR #1674, opened 2025-11-10
Extend Lua script to add line colors to DELFI dataset

HerrLevin:
This pull request adds support for assigning colors to the delfi gtfs source based on the CC0 data sourced from https://github.com/Traewelling/line-colors

Screenshot 2025-11-10 at 22-34-30 MOTIS
Screenshot 2025-11-10 at 22-34-17 MOTIS
Screenshot 2025-11-10 at 22-33-57 MOTIS

I've added a python script to generate the lua table but I didn't want to call it in any ansible or ci script yet.

**Full disclosure:**
I couldn't get the original delfi lua script working without removing the following lines:
https://github.com/public-transport/transitous/blob/8d6f0d23fb232f0828129dffd0ed7eb71ecbdafb/scripts/de-DELFI.lua#L34-L41

Otherwise I would always run into a segmentation fault.

I'm not sure wether this is an error on my part or if the latest motis executable has some kind of problem here.


felixguendling:
LGTM. I had this on my TODO list too but you were faster. Great work!

Are you on MacOS ARM64? I had issues with segfaults in LuaJIT there, too but I didn't have the time to dig deeper.

HerrLevin:
Nice, thanks!

And yes. ARM64. There was some very weird behaviour when using lua scripts. Good to know. :D

Quick question due to the failing test: Can `SPDX-FileCopyrightText` for de-delfi-colors.lua just contain a link to the line colors repository?

mlundblad:
I noticed that the U-Bahn lines in Hamburg are encoded as "U 1", "U 2", "U 3", and "U 4" in the Träveling data.
As far as I could see the route_short_names from Delfi seems to be "U1", "U2", "U3", "U4"

Not sure if maybe some normalization would be needed here? (possibly comparing with ignoring whitespace?)

felixguendling:
> Not sure if maybe some normalization would be needed here? (possibly comparing with ignoring whitespace?)

Either making comparison more robust or just updating the string by removing space between "U" + "some number(s)"

HerrLevin:
Thanks for noticing! I've normalized the routes by removing the spaces. Now Hamburg is mapped correctly.

jbruechert:
> Quick question due to the failing test: Can `SPDX-FileCopyrightText` for de-delfi-colors.lua just contain a link to the line colors repository?

The test seems happy now, and the copyright text looks fine.

mlundblad:
Thanks!

felixguendling:
btw for me it doesn't crash anymore on MacOS ARM64 after this patch:
https://github.com/motis-project/motis/commit/f8e35bba8f59de330067d207155c2c748e3ee6bf

The original workaround worked for me, the official bugfix didn't solve it:
https://github.com/TeamREPENTOGON/LuaJIT/commit/deac6e400b73ee07d4d573bd39f11f0649811ed2

So now we have both (cherry-picked bugfix + previous workaround).

PR #1676, opened 2025-11-11
Add IPA pronunciation link for Transitous

HerrLevin:
At the Open Transport Community Conference there was talk about how unsure some people are about how exactly the name is pronounced.
I've added a phonetic transcription for the pronunciation, as I remember it from the conference as being "correct".


felixguendling:
That's also how I remember it :)

PR #1680, opened 2025-11-15
Re-add upcoming events section to the website

vkrause:
We don't have confirmed talks at those events (yet), but at least people of the team attending.

Also, reverse the sorting of the talks section to have the newest ones first.


felixguendling:
Maybe also worth a social media post? :)

vkrause:
> Maybe also worth a social media post? :)

I'd do one each (at least) and closer to the corresponding event, once we know more about what we'll actually do there.

PR #1682, opened 2025-11-17
Add new RT feeds and ŁKA

kasmar00:
- ŁKA - switch to two feeds:
   - one for trains: sanitized, with colors and properly marked rail replacement buses
   - and one for buses: sanitized, with RT from zbiorkom
- New RT feeds for cities around Warsaw
- New feed for Świnoujście

### DIFF
--- a/feeds/pl.json
+++ b/feeds/pl.json
@@ -527,7 +527,8 @@
             "url": "https://cdn.zbiorkom.live/gtfs-rt/lodz.pb",
             "license": {
                 "spdx-identifier": "MIT"
-            }
+            },
+            "fix": true
         },
         {
             "name": "Białystok",


Lach-anonym:
Can you check the location of Pleszew station in ŁKA feed? Normally it is wrongly placed where Pleszew Miasto is located, so I hope it can be solved somehow

kasmar00:
OMG nice catch! And Pleszew Miasto isn't even a regular track station 🤯
I'll correct the location in my sanitized feed.

kasmar00:
Fixed the feeds list, import passes now.

And some more 🤯: Pleszew Miasto is on a train line that's both 750mm and 1435mm but there are regular services on 1435mm gauge (missing on transitous, we'll see to have that fixed 😉) 🤯

jbruechert:
Thanks!

PR #1685, opened 2025-11-20
BlaBlaCar Bus trip display update

Lach-anonym:
I've set the BBC B trips display to also show the route number
Also I've modified the headsign to show the longer route name, it might make the headsigns much larger, but I think it's more comprehensive than what is already included in the feed, the last stop option for the script would be helpful here...


Lach-anonym:
I did not expect to find out that OE feed is expired lol

jbruechert:
The current headsigns look mostly fine. IMO this should always be the destination. What do you think is better about using route_long_name as the headsign?

Lach-anonym:
The three letters in headsign are quite mysterious for an average user, the full route name at least shows some places served by a trip

Lach-anonym:
Closing because BlaBlaCar will disappear anyways

PR #1686, opened 2025-11-21
Fix timezone in CVDL region feeds

Lach-anonym:
I've noticed that France recently got too good night service... so I've reminded myself about Montréal timezone set for Châteauroux - that has damaged the entire feed (even if other agencies had Paris)
This PR fixes the issue

### DIFF
--- a/src/generate-france.py
+++ b/src/generate-france.py
@@ -37,7 +37,8 @@ if __name__ == "__main__":
 
     # Map of datasets to MOTIS Lua scripts to apply to them
     scripts = {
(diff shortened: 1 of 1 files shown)


jbruechert:
Have you tested this? I don't think the feed will survive gtfsclean in the current state.

Lach-anonym:
Not really, but if it passes then still the time zone issue has to be fixed, there are still those 3AM trips...

Lach-anonym:
seems to be fixed for the time being

PR #1689, opened 2025-11-22
Add a Lua script for SNCF GTFS feed

Lach-anonym:
At no point this is finished, I need someone to check the script, as I'm not that good at coding - there are comments to help in understanding what has to be done. Feel free to put comments, so issues can be fixed.
SNCF official feed is really hard to deal with - some categories have to be entered manually, as there is not enough data

### DIFF
--- a/scripts/fr-sncf.lua
+++ b/scripts/fr-sncf.lua
@@ -36,28 +36,30 @@ end
 
 function process_trip(trip)
   if trip:get_route():get_route_type() == "106" then -- regional train with 5 or 6 digits with number starting from 4 is actually a bus
-    if string.len(trip:get_headway) == 5 or 6 and
-      if string.byte(trip:get_headway(), 1) == "4" then
+    if string.len(trip:get_headsign) == 5 or 6 and
+      if string.byte(trip:get_headsign(), 1) == "4" then
         trip:get_route():set_route_type(714) -- rail replacement bus
         trip:get_route():set_clasz(BUS)
-        trip:set_display_name("Autocar TER " .. trip:get_headway())
+        trip:set_display_name("Autocar TER " .. trip:get_headsign())
       end
     else
-    trip:set_display_name("TER " .. trip:get_headway())
+    trip:set_display_name("TER " .. trip:get_headsign())
   elseif trip:get_route():get_route_type() == "102" then
     if trip:get_route():get_clasz() == "NIGHT_RAIL" then
-      if string.len(trip:get_headway) == 5 then
+      if string.len(trip:get_headsign) == 5 then
         trip:get_route():set_route_type(714)
         trip:get_route():set_clasz(BUS)
-        trip:set_display_name("Autocar ICN " .. trip:get_headway())
+        trip:set_display_name("Autocar ICN " .. trip:get_headsign())
       else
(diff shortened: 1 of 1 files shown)


felixguendling:
Is there a NeTEx feed with better data?

Lach-anonym:
I have no idea if NeTEx feed is better, but it's available from the same database as the GTFS feed: https://transport.data.gouv.fr/datasets/horaires-sncf

jbruechert:
I'm sorry, but with these many things to correct a script is really not the right medium to do this.
This needs either a new feed conversion or NeTEx support if that is better.

Lach-anonym:
I've seen the comments under the schedule datas and there is one that complains about the NeTEx feed having the same low quality, although it at least features the branding info

felixguendling:
> I'm sorry, but with these many things to correct a script is really not the right medium to do this. This needs either a new feed conversion or NeTEx support if that is better.

I would not say that the amount of things a script changes is relevant for "is a script the right tool?". The question is more: are those things stable - i.e. is the script maintainable?

All those cryptic IDs like "OCESN-87391003-87471003" make me doubt that they are stable. A stable ID would probably look more like a readable train number.

jbruechert:
The NeTEx feed looks a bit better (branding like TER and proper numbers are in there), but the converter log contains a worrying amount of
```
Error  : UIC Code not found from station resarail <BEBHM>
Error  : UIC Code not found from station resarail <NLABK>
Error  : UIC Code not found from station resarail <NLAAC>
```

It also shows that this is converted from what looks like hrd, something that MOTIS might be able to read directly.

> I would not say that the amount of things a script changes is relevant for "is a script the right tool?". The question is more: are those things stable - i.e. is the script maintainable?

> All those cryptic IDs like "OCESN-87391003-87471003" make me doubt that they are stable. A stable ID would probably look more like a readable train number.

I'd argue that even manually collecting all numbers that belong to a specific route type is not sustainable. You are of course absolutely right about the IDs.

Lach-anonym:
The feeds seem to be generated from some internal SNCF system, for GTFS the train number is only present in the headsign of trips, which makes the work much harder. The data is stable at least for the period of validity of the feed and my guess is that neither the IDs nor the short names will not change that frequently, so it's enough to check that during the timetable changes. Though I must admit that without tools to check that it's hard.
About OCESN line codes, they represent the first and last station of the train route, for the Nancy-Lyon ICs I've also encountered some routes prepared for future line closures, when the trains will be shortened

jbruechert:
image

The information in the NeTEx feed looks workable.

Lach-anonym:
well then it's a gamechanger

felixguendling:
Since we'll add NeTEx support to MOTIS in the coming months, this is probably a better option then. Especially if the IDs the script here is based on are not stable, which would require updating it all the time with high manual effort.

felixguendling:
> The information in the NeTEx feed looks workable.

It's funny that this is one of the places where the NeTEx is invalid because there's a `<BrandingRef ref="OUI" />` to a non-existing `<Branding id="OUI">` with `Name` and `URL`. But I think we can work around this and create a dummy brand with only an ID.

https://enturas.atlassian.net/wiki/spaces/PUBLIC/pages/728727682/Branding

PR #1692, opened 2025-11-23
Update de-delfi-colors.lua

SpielenmitLili:
Updated de-delfi-colors.lua with help of script to recent version of line-colors from Träwelling repo


jbruechert:
Thanks :)

PR #1699, opened 2025-11-25
Add IDFM GTFS-RT feed

Jouca:
Add the possibility to have GTFS-RT feeds for TripUpdates & Alerts on IDFM (Paris, Île-de-France, France).

These feeds are publicly hosted and is created by my GitHub project : https://github.com/Jouca/IDFM_GTFS-RT

### DIFF
--- a/feeds/fr.json
+++ b/feeds/fr.json
@@ -153,6 +153,24 @@
             }
         },
         {
+            "name": "reseau-urbain-et-interurbain-dile-de-france-mobilites",
+            "type": "url",
+            "url": "http://gtfsidfm.clarifygdps.com/gtfs-rt-trips-idfm",
+            "license": {
+                "url": "https://github.com/Jouca/IDFM_GTFS-RT/blob/main/LICENSE"
+            },
+            "spec": "gtfs-rt"
+        },
+        {
+            "name": "reseau-urbain-et-interurbain-dile-de-france-mobilites",
+            "type": "url",
+            "url": "http://gtfsidfm.clarifygdps.com/gtfs-rt-alerts-idfm",
+            "license": {
+                "url": "https://github.com/Jouca/IDFM_GTFS-RT/blob/main/LICENSE"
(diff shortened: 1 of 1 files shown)


traines-source:
Thanks! I hadn't thought about the fact that the French config is automatically generated by https://github.com/public-transport/transitous/blob/main/src/generate-france.py and has a pending regeneration https://github.com/public-transport/transitous/pull/1609. We should probably adapt that script like for Austria https://github.com/public-transport/transitous/pull/1651 to only touch the fr.json entries it knows about.

In the meanwhile, we could theoretically just merge your changes like that and hopefully remember them when regenerating with the script.

I suppose your feed matches the trip ids from the IDFM GTFS? Do you have any insight in when they and you switch over to new feed versions with potentially new trip ids (not that this is very relevant for the time being, since we're losing lots of RT especially in France due to non-coordinated switchovers, but there are ideas how to improve that...)

Jouca:
No problem!

If you're looking for someone to follow up the PR #1609 relating to the generation of the french dataset with the new system for taking only entries he knows about (since I'm not sure if it's still in work), I can manage to do it for another PR.

Relating to the trips, it should definitely match from the IDFM GTFS since on the project I ensure to take the GTFS everyday from their server and refresh the local database, the links also for the real-time feeds normally shouldn't change (unless I have to do something relating to server side, but I'll keep you updated with a PR with the modifications if I have to do some)!

jbruechert:
We should be able to add the new entries to the script, like we currently do for Lyon. But the script definitely needs to be changed to only editing the entries it knows about in general.
That might also allow to get rid of the manual realtime matching in the script, when we can simply rename the two sources to have the same name in the json.

traines-source:
After @jbruechert had done the necessary work regarding the script, someone else now added your feeds in #1796, so they should become available from tomorrow morning :tada: Thank you for your work, and sorry for the wait!

PR #1701, opened 2025-11-25
generate-france: Add id to generated sources

jbruechert:
First step to editing instead of regenerating the entire file.


Lach-anonym:
Do you plan to add agency skipping in the script?

jbruechert:
No, I want to remove as much as possible from the script and allow to edit it in the json directly.

PR #1722, opened 2025-12-01
website: Allow to specify the time

jbruechert:


jbruechert:
image

Won't win any design awards but works

website/src/TransitousWidget.purs
```diff
@@ -0,0 +1,96 @@
...
+
+view :: State -> Dispatch Message -> ReactElement
+view state dispatch = H.div "text-end card-body p-4"
+  [ SearchBox.view state.departureBox (dispatch <<< DepartureMsg)
+  , SearchBox.view state.arrivalBox (dispatch <<< ArrivalMsg)
+  , H.input_ "form-control"
+      { type: "datetime-local"
+      , value: fromMaybe "" $ do
+          dt <- state.dateTime
+          s <- hush (formatDateTime "YYYY-MM-DDTHH:mm" dt)
```
luna-cant-code: Just out of curiosity: This is going to use the 24h clock?

jbruechert: It seems to work, I didn't question it too much tbf :D

But looks correct: https://pursuit.purescript.org/packages/purescript-formatters/7.0.0

If you were referring to the user-visible string: This is just bringing it into the format needed for the html input element, I don't think it influences visuals.

PR #1724, opened 2025-12-02
usage policy: Improve browser-related rules

jbruechert:
The header we currently recommend can not be sent in practice, because
our OPTIONS response does not allow it.

### DIFF
--- a/website/content/api.html
+++ b/website/content/api.html
@@ -19,7 +19,7 @@ We provide the service on a best-effort basis and reserve the right to stop prov
 <h3>Resource Usage</h3>
 
 <p>Our hosting resources are limited.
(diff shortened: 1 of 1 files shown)


derhuerst:
> The header we currently recommend can not be sent in practice, because our [CORS pre-flight] `OPTIONS` response does not allow it.

Can't we just allow `User-Agent`?

jbruechert:
> > The header we currently recommend can not be sent in practice, because our [CORS pre-flight] `OPTIONS` response does not allow it.
> 
> Can't we just allow `User-Agent`?

As fas as I know, the User-Agent header can't be modified in the browser. Of course we can allow another header, but personally I'm happy with the referrer.

vkrause (approved):

derhuerst:
> As fas as I know, the `User-Agent` header can't be modified in the browser.

Apparently [it depends on the browser](https://stackoverflow.com/a/42815264). 😕

PR #1725, opened 2025-12-02
updated delfi-colors

luna-cant-code:


mtwente:
A question continuing this conversation: https://github.com/Traewelling/line-colors/pull/305 

This PR now merged the colour definition file for DELFI data, containing also data for non-Delfi routes (i.e. the Swiss ones). But since this file is only used in combination with DELFI data (and most of Swiss public transport is not part of DELFI data), this means that the information in Swiss routes does absolutely nothing at the moment, right? I would need a separate separate colour processing script like `ch-opentransportdata-colours.lua` for the Swiss opentransportdata.ch data so the colours are added to the Swiss output, wouldn't I?

luna-cant-code:
> this means that the information in Swiss routes does absolutely nothing at the moment, right? 

That is absolutely correct.

>  would need a separate separate colour processing script like ch-opentransportdata-colours.lua for the Swiss opentransportdata.ch data so the colours are added to the Swiss output, wouldn't I?

Yesn't

We're currently trying to find a consensus on the issue, not only because of Swiss data but also because of VBB and VBN data in Germany.
At the moment, we just collect the data - Duplicate agency IDs between Swiss, DELFI and VBN data don't exist, so they don't really hurt

https://github.com/Traewelling/line-colors/issues/335

In short: A separate .lua for Switzerland and other feeds might be helpful.
I wanted to open an issue anyway, just couldn't find the time to do that properly 😅

luna-cant-code:
Should I just open an issue and PR for that, or do you want to do that?

I need some training anyways tbh, don't want to be the girl that's just pushing updated colours 😅
Wanted to contribute a bit more than that 😅

mtwente:
Go ahead! :)

Issue #1540, opened 2025-09-23
Duplicate departures in the VBB area


jeriox:
(first, thanks so much for the project and your effort!)

I noticed that a lot of departures are duplicated in the VBB area. While you explicitly drop BVG and S-Bahn Berlin from the DELFI feed (https://github.com/public-transport/transitous/blob/main/feeds/de.json#L25), there are a lot of other operators in the VBB. I thought about submitting a PR to just drop them as well. While this would work for operators that only operate inside the VBB area (e.g. Verkehrsbetrieb Potsdam, which suffers from duplicate departures in the dataset), my understanding of the data is that this would lead to problems when removing operators that both operate inside and outside VBB. An example for that is Ostdeutsche Eisenbahn (ODEG), which has duplicate departures of the RE1 service through Berlin. However, dropping them from the DELFI feed would mean also losing departures that VBB does not know about (e.g ODEG RB13 Parchim - Schwerin), which operates in Mecklenburg-Vorpommern.

Another issue with this is that (at least for Verkehrsbetrieb Potsdam), the DELFI departures have proper real time data attached to them, while the VBB departures oftentimes just show them as punctual.

I'm looking forward to hearing your opinion on this and happy to help out!

felixguendling:
The reason to use VBB data directly is that the GTFS-RT data created by @derhuerst is better than what we had at this time. It's hard to say if that changed now that we use SIRI data (e.g. because we don't know if it contains also REF-AUS or only AUS).

With the disadvantages you list, we should definitely reconsider this decision.

Another option would be to check why the deduplication in MOTIS does not catch those cases.

jeriox:
Thanks for the explanation! Is there anything I can do to help?

jbruechert:
I think switching back to just using DELFI would be okay here.
Of course it would be nice if at some point automatic deduplication would catch most of this, as we could then use the more detailed feeds to get additional information like route colors, that DELFI still does not have.

jeriox:
> Of course it would be nice if at some point automatic deduplication would catch most of this

I tried to find out why this is the case, but I am a bit lost ^^ i set up a local MOTIS instance and imported both VBB and DELFI feeds. There I can confirm the duplicate departures. Looking at the raw GTFS data, I noticed that the same trip (with identical times and stop id) carries two different trip_ids between the VBB feed and the GTFS feed. Is this expected? 

[...]

PR #1727, opened 2025-12-03
Dropped BSAG Bremen and VOS Osnabrück from DELFI to use local GTFS feed

luna-cant-code:
BSAG Bremen offers higher data quality in VBN's GTFS dataset, in my experience, that's why I'd like to kick it out from DELFI.

The higher quality is noticeable with special trips which aren't always in DELFI's dataset.

### DIFF
--- a/feeds/de.json
+++ b/feeds/de.json
@@ -33,7 +33,8 @@
                 "SWO Mobil GmbH",
                 "FlixBus-de",
                 "EUROSTAR",
-                "Bremer Straßenbahn AG"
+                "Bremer Straßenbahn AG",
+                "Verkehrsgemeinschaft Osnabrück"
             ],
             "http-options": {
                 "fetch-interval-days": 2


felixguendling:
Does the VBB problem exist here as well?

> there are a lot of other operators in the VBB. I thought about submitting a PR to just drop them as well. While this would work for operators that only operate inside the VBB area (e.g. Verkehrsbetrieb Potsdam, which suffers from duplicate departures in the dataset), my understanding of the data is that this would lead to problems when removing operators that both operate inside and outside VBB

https://github.com/public-transport/transitous/issues/1540

luna-cant-code:
I'm not that often in the VBB area, just seen it myself with the occasional duplicate data in Leipzig from services originating/terminating there, like the S-Bahn line S4.

You could drop (almost) all VBB operators from DELFI, especially those only operating in Berlin and Brandenburg.
The S4 of S-Bahn Mitteldeutschland would be one exception due to (sometimes) missing real-time data.

luna-cant-code:
(Quick addition to the list: VOS Osnabrück)

jbruechert:
For these clearly very local operators it's fortunately easy. For everything else, I suppose we need better fuzzy merging.

luna-cant-code:
> For these clearly very local operators it's fortunately easy. For everything else, I suppose we need better fuzzy merging.

Fuzzy merging makes sense but I don't know how to start on that 😅

jbruechert:
> > For these clearly very local operators it's fortunately easy. For everything else, I suppose we need better fuzzy merging.
> 
> Fuzzy merging makes sense but I don't know how to start on that 😅

This needs to happen in MOTIS, so it is a little more involved.

Issue #1208, opened 2025-06-12
Time zones?


pbsurf:
On transitous.org, it seems times are converted from the schedule time zone to the user's current time zone, which is a bit confusing - is this intended behavior?

Specific example: on a computer set to UTC-7, TGV 5537 is listed as departing from Strasbourg (UTC+2) at 04:37 instead of [13:37](https://www.sncf-connect.com/en-en/train/timetables/strasbourg/Marseille), both in "Departures" and for routes found in "Connections".

Also, it would be nice to have an option to pause the live animation on the map since it consumes significant CPU and GPU.

derhuerst:
I think it would be best to show users both the transport-local and the user-local time.

For example, the UI could show one of the two times, with a tooltip (or sth. similar, not ideal but at least there's *something*) showing the other.

jbruechert:
As far as I remember without checking, the API returns the date and time including a time zone so apps can handle this however they like.
The web interface is directly what comes with MOTIS, but I'm pretty sure they would be open to accepting a PR to display the local time as well.
For me personally the API is the primary focus though.

traines-source:
AFAIK it's just always returned as UTC from the API, so in that sense clients can display it however they like yes :D MOTIS does have the info about the agency (/stop?) timezone internally though, we would just need to expose it. How reliable that data from the GTFS feeds is, is another question.

felixguendling:
Since in one GTFS feed, all agencies have to have the same timezone (which is used for translating times from `stop_times.txt` to UTC), the agency timezone can't be reliable for cases like Amtrak or airlines. Therefore, only stop timezones should be used for this purpose if available. However, `stop.timezone` is an optional field.

So we need to decide if agency timezones should be used for this purpose as fallback (which is probably good for a most feeds but highly misleading for others) or we need some library to determine timezones on our own.

Code for printing local times is already available. So bringing this information (local time in stop timezone, fallback agency timezone) is probably not a big deal.

https://github.com/motis-project/nigiri/blob/ea1b88e7a3da8d03af8b4de8c5275332c5e386fa/src/rt/frun.cc#L556-L573

mlundblad:
In GNOME Maps we use libgweather currently to determine the timezones.

But it has its drawbacks. The calculation is somewhat resource-heavy, it does triangulation based on the wather station data (METAR station) positions.
So, I did as a shortcut:

* Calculate the timezone at the start and end of an itinerary, if both are identical, assume the same timezone applies for every stop
[...]

PR #1733, opened 2025-12-04
Web UI: Timezone-aware display for stop-local vs user-local times

Jujux167:
Addressed Issue

This reengineering task is based on the issue reported in the Transitous repository: https://github.com/public-transport/transitous/issues/1208
Times on transitous.org were shown only in the browser’s local timezone, causing confusion for cross–time‑zone trips (e.g., remote departures appearing offset and obscuring DST/day changes at the stop).

What Was Reengineered

Added a small, reusable timezone formatting utility for the web UI:
Time.purs : PureScript FFI interface exposing formatTimes.
Time.js : JS implementation using Intl.DateTimeFormat with timeZone and timeZoneName.
Utility behavior:
Treats API schedule times as UTC inputs.
Converts to the stop’s local timezone using the stop’s IANA tz (e.g., "Europe/Paris").
Also formats the user’s (browser) local time.
Returns { stopTime, userTime, showBoth } where showBoth is true if stop tz ≠ browser tz.
Reengineering Strategy / Approach

Kept scope to the web UI; no backend/MOTIS changes and no map animation changes.
Implemented a minimal FFI utility to centralize timezone handling:
Input: UTC datetime string and optional stop timezone.
Output: two formatted strings plus a boolean flag to drive conditional UI rendering.
Used native Intl.DateTimeFormat for correctness, DST handling, and performance.
Preserved existing structure and code style; introduced only the utility so components can adopt it with minimal refactors.
Note on scope: This repository contains the search UI and forwarder; the departures/connections rendering appears to live in MOTIS Web. This PR provides the utility here to standardize formatting and enable integration where those components are implemented.
Impact of Changes

Improves clarity and correctness for cross‑time‑zone trips:
When stop tz equals browser tz: renders a single, compact time.
When they differ: enables rendering both, e.g., “13:37 CET (Europe/Paris) / 04:37 EST (your time)”.
Handles DST transitions and date rollovers correctly at the stop.
Minimal performance impact (native Intl); possible to cache formatters per tz if needed.
Backwards compatible: if tz is missing/invalid, the utility falls back to browser tz and showBoth = false.
Files Changed

Added: Time.purs
Added: Time.js
Follow-up (for integration in departures/connections UI)

Replace direct time rendering with formatTimes(utc, Just stopTz):
If showBoth is false: show stopTime only.
If showBoth is true: show stopTime plus userTime with concise labels/tooltips.

Reviewer: @suhadaudd11

### DIFF
--- a/website/src/Time.js
+++ b/website/src/Time.js
@@ -1,0 +1,35 @@
+// SPDX-FileCopyrightText: 2025 Transitous contributors
+// SPDX-License-Identifier: AGPL-3.0-or-later
+
+// PureScript FFI for timezone-aware time formatting
+// Uses Intl.DateTimeFormat with timeZone and timeZoneName.
+
+/**
+ * @param {string} utcString - UTC datetime string (ISO-8601 or Date-parsable)
+ * @param {?string} maybeStopTz - Optional IANA timezone name for the stop
+ * @returns {function(): { stopTime: string, userTime: string, showBoth: boolean }}
+ */
+export const formatTimes = (utcString) => (maybeStopTz) => () => {
+  // Parse input as UTC. new Date(utcString) interprets ISO-8601 with Z correctly as UTC.
+  // If utcString is epoch milliseconds as string/number, Date will handle that as well.
+  const date = new Date(utcString);
+
+  // Browser/user timezone
+  const userTz = Intl.DateTimeFormat().resolvedOptions().timeZone;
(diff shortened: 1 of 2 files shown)


suhadaudd11:
Dear Julien,

Please cancel this Pull Request as it is currently being merged into the original/main repository.
Kindly ensure any future PR is directed to your forked repository, as we do not want to mix with the real production project. This is only for the assignment purpose and for your learning. Thanks.

PR #1751, opened 2025-12-12
ua: Add more guessed urls

jbruechert:
There are a number of skipped feeds that could be saved with a bit of work.

Drop gbfs feed that 404s

### DIFF
--- a/feeds/ua.json
+++ b/feeds/ua.json
@@ -85,10 +85,115 @@
             }
         },
         {
-            "name": "nextbike",
+            "name": "ivano-frankivsk",
+            "type": "http",
+            "url": "https://city.dozor.tech/iv-frankivsk/gtfs/static.zip",
+            "fix": true,
+            "extend-calendar": true
+        },
+        {
+            "name": "ivano-frankivsk",
             "type": "url",
-            "spec": "gbfs",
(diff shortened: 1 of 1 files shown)


felixguendling:
> agency.txt entry missing

MOTIS should just generate a dummy agency entry.
Only the timezone must be set on the dataset.

jbruechert:
> > agency.txt entry missing
> 
> MOTIS should just generate a dummy agency entry.
> Only the timezone must be set on the dataset.

This is currently blocked on gtfsclean, which is quite unfortunate. I will try if MOTIS is able to parse the rest of the feed by itself, then we might be able to skip gtfsclean here.
There are some other issues in these feeds though, like unescaped stop names containing commas, leading to a different number of columns. gtfsclean only throws these away, but is at least able to salvage the correct entries.

jbruechert:
Works in principle, but introduces null-island trips on the map.

felixguendling:
> null-island

https://github.com/public-transport/transitous/issues/1752

I wrote my opinion here. I think this can be hidden by the UI. Deleting the whole stop because one information is (obviously) wrong is like throwing the baby out with the bathwater IMO.

PR #1756, opened 2025-12-14
jp: GTFS-Realtime for Tokyo railways

MKuranowski:
This PR adds two GTFS-Realtime feeds for Tokyo railways - one with alerts for almost all routes, and one with trip updates for a couple of the biggest agencies. Both are available under CC-BY-4.0.

I'm guessing cc @altonss as they're the maintainer mentioned in jp.json.


mlundblad:
Thanks!

PR #1761, opened 2025-12-16
fix route_type of planes from Nantes Airport (NTE)

jendrikw:
I'm marking this as a draft because these changes are untested. If someone knows how to test it so that things don't blow up in production please tell me.

This is how it currently looks:
image

### DIFF
--- a/feeds/fr.json
+++ b/feeds/fr.json
@@ -331,7 +331,8 @@
             "license": {
                 "url": "https://transport.data.gouv.fr/datasets/arrets-horaires-et-circuits-des-lignes-de-transports-en-commun-en-pays-de-la-loire-gtfs-destineo-reseaux-aom-aleop-1",
                 "spdx-identifier": "ODbL-1.0"
-            }
+            },
+            "script": "fr-aeroport-nantes.lua"
         },
         {
             "name": "arrets-horaires-et-circuits-des-lignes-de-transports-aleop-1",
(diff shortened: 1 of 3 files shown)


jbruechert:
I think it might make sense to instead remove the entire agency. The times do not look realistic to me.

Lach-anonym:
There is an option it has been fixed in the newer feeds, but that requires regenerating the French feeds
And I doubt it will happen

jbruechert:
> There is an option it has been fixed in the newer feeds, but that requires regenerating the French feeds And I doubt it will happen

There is nothing blocking that, someone just needs to do it.
Unless this is one of the feeds for which the url changes for every version, we would get updates of it in any case.

Lach-anonym:
Oh I thought it has something to do with implementing NeTEx and that we could get better sources

felixguendling:
- It could be that for NeTEx the script looks different (scripting also works for NeTEx but the data might contain different names, etc. what you use for matching)
- if the times are wrong in GTFS, they are probably also wrong in NeTEx
- NeTEx != NeTEx - until now I had to make changes to MOTIS for each new NeTEx dataset. So support for one profile doesn't mean anything for another profile. It's not very likely that MOTIS will work with a random NeTEx dataset that I haven't adjusted the parser for.

Lach-anonym:
Also may I remind that the departure and arrival times are swapped, so it would be better to skip this one agency, I don't know if it has even improved at all
But this function is still missing for the France generating script

applecuckoo:
I assume 923d5f7 succeeds this PR?

jendrikw:
@applecuckoo Yes, superseded by https://github.com/public-transport/transitous/commit/923d5f76331383595c754dbc5b5e27284fa53e7c

PR #1772, opened 2025-12-19
Update URLs for gtfs.cc

gtfscc:
This pull request updates the URLs for feeds from gtfs.cc, while also adding coverage for Czosnów County Public Transit


jbruechert:
Thanks!

PR #1776, opened 2025-12-21
change the display name of Eurostar trains from nl feeds

jendrikw:


felixguendling:
As far as I understood it, our strategy regarding EuroStar was to remove them from all feeds and only use those from the official Eurostar feed. But maybe @jbruechert knows more? You can filter routes, trips, and agencies with `return false` in Lua.

jendrikw:
I'm sorry, I didn't realize this has been discussed before. I can that change the script to filter them out if you prefer.

jbruechert:
> As far as I understood it, our strategy regarding EuroStar was to remove them from all feeds and only use those from the official Eurostar feed. But maybe @jbruechert knows more? You can filter routes, trips, and agencies with `return false` in Lua.

I guess that makes most sense, yeah

PR #1780, opened 2025-12-22
Christmas is coming: A lot of new colours for the Rhine Neckar region

luna-cant-code:
Thanks to @wittighausen we got a huuuuuuge package of new and shiny colours for this year's holidays! ^^


wittighausen:
Just a quick question, as I’m not entirely sure how this works: is it possible to adjust the line names?

For example, the lines operated by Stadtwerke Pirmasens have their own abbreviation before the line number (e.g. “SWPS 201”). This causes an issue here because the space between the abbreviation and the line number gets removed, resulting in an incorrect line name. As a consequence, the line color cannot be displayed correctly, since it depends on the correct line name.

Edit: Nevermind, it works even though the names don't match perfectly.

mlundblad:
Thanks!

PR #1783, opened 2025-12-24
Add licensing information and other options to Namtang feed

ExoSkye:
I should've probably added these in my first PR - but I didn't think it was necessary. However upon further inspection I found that MobilityDatabase sources do need licensing information to be attached. So I've added that here.

I also noticed that the feeds being too fast checking was causing some trips to be removed when they shouldn't be, so I've also told it to not remove trips that are too fast.

### DIFF
--- a/feeds/th.json
+++ b/feeds/th.json
@@ -10,7 +10,6 @@
             "name": "th-namtang",
             "type": "mobility-database",
             "mdb-id": "mdb-1831",
-            "drop-too-fast-trips": false,
             "license": {
                 "spdx-identifier": "CC-BY-4.0"
             }


jbruechert:
The Mobility Database only has license urls, so setting the identifier ourselvs makes sense.

Have you looked into why these trips are too fast? This could point to wrong coordinates. If that is the case, we should not simply allow these, as it can create "wormholes" that break routing in other places.

ExoSkye:
> Have you looked into why these trips are too fast? This could point to wrong coordinates. If that is the case, we should not simply allow these, as it can create "wormholes" that break routing in other places.

I'm not sure - I'll take a look when I'm back at home, but it seemed honestly kinda random which ones were being removed (at least as per the warnings on MobilityDatabase)

jbruechert:
> > Have you looked into why these trips are too fast? This could point to wrong coordinates. If that is the case, we should not simply allow these, as it can create "wormholes" that break routing in other places.
> 
> I'm not sure - I'll take a look when I'm back at home, but it seemed honestly kinda random which ones were being removed (at least as per the warnings on MobilityDatabase)
> 
The ones in the validation report look extremely justified though, 20km im one minute...

ExoSkye:
As in, that there didn't seem to be a pattern in which were being filtered out, other than speed

There wasn't a specific stop that was very common, they just seemed very.. well yeah, random

not in the reasoning, to be clear

ExoSkye:
I'm happy to remove the config option when it comes to that though, the licensing is worth having in the feed info

ExoSkye:
Ok, so on further inspection - there definitely are some issues I didn't notice earlier with this feed.
For one, it doesn't contain all of the SRT trains (for example I couldn't find Special Commuter 1115 from Thon Buri to Nakhom Pathom, which you can see on SRT's ttsview [here](https://ttsview.railway.co.th/v3/search/?qType=22&qParam=171&auth=c30a940d05bdeb979eaeff8e83feadd10742fa6e02355db0731de68715528e151766572377))

And on top of that, the ones that it does have sometimes have no departure/arrival times, replaced with what appears to be filler data (which is causing some of the trips to be too fast).

This is.. rather annoying, to say the least - and would explain why I've noticed in the past that Google Maps has inaccurate data at times.

Given ttsview exists, it may be possible to make an automated conversion tool to make gtfs from it, but I'm not sure if I personally have the skillset to do so. (plus what license that would be under would be iffy, given the `Copyright © 2025 SRT All rights reserved.` text on the ttsview website)

It may be worth excluding SRT from the feed, but other operators have this issue too, so I believe it's best just to remove the config option I put in to skip removing these, since evidently it would be bad for them to be included.

I'm just a humble train photographer with a tiny bit of software dev experience lol

If you agree with removing the option, based on this, then I'll commit that asap!

ExoSkye:
It would also appear that the Namtang app, which this data is also used for, is also missing the S. Commuter 1115, and for the Special Express 7 (which was one of the ones with filler data), it knows of it's existence but can't route using it (due to the missing timings)

jbruechert:
Removing the option sounds good. Once the data improves, it will automatically improve over time on our side :)

ExoSkye:
Alright!

ExoSkye:
Sorted!

PR #1791, opened 2025-12-27
th: add fixup script for namtang

ExoSkye:
This includes a fixup script for namtang, using various pieces of information to rebuild the correct route types.

Sadly I haven't worked out how to satisfactorily sort out the weird `th_name;en_name` format that namtang uses in various places (because they can't do it in the GTFS standard way, of course...)

But this, at least, is definitely a step in the right direction - prior, the BTS lines and red line were all considered trams? when they're very definitely not that lol


jbruechert:
Thanks!

PR #1800, opened 2025-12-29
Add Mobility Database feeds for Japan

jbruechert:
- **Add mobility database feeds for Japan**
- **fix url-override for gtfs-rt feeds from mobility database**

### DIFF
--- a/feeds/jp.json
+++ b/feeds/jp.json
@@ -17,7 +17,8 @@
         {
             "name": "hiroden",
             "type": "transitland-atlas",
-            "transitland-atlas-id": "f-hiroden~jp"
+            "transitland-atlas-id": "f-hiroden~jp",
+            "fix": true
         },
         {
             "name": "tokyo-rail",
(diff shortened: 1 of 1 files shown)


mlundblad:
Looks like there's some ID mismatch, or something?

"reseau-orizo-grand-avignon--83330" sounds like it tries to match some RT feed with a French static feed, maybe?

mlundblad:
ありがっとう! 😄

emmambd:
Amazing to see you use these feeds in Transitous already! 🚀

@mlundblad - for the realtime mismatch issue, would you mind creating an issue in [our repository here](https://github.com/MobilityData/mobility-feed-api/issues) so we can fix the bug at the root?

jbruechert:
> Amazing to see you use these feeds in Transitous already! 🚀
> 
> @mlundblad - for the realtime mismatch issue, would you mind creating an issue in [our repository here](https://github.com/MobilityData/mobility-feed-api/issues) so we can fix the bug at the root?

Hi, nice to see you over here :)
That issue was pre-existing on the Transitous main branch at the time, and is unrelated to Japan.
For France, we use our own script to import the feeds from the open data portal, because there are multiple cases of one item containing multiple gtfs resources, and as far as I can see the Mobility Database does not currently handle this case.

emmambd:
@jbruechert Could you share an example of 1 item containing multiple GTFS resources? Then our team can investigate.

jbruechert:
> @jbruechert Could you share an example of 1 item containing multiple GTFS resources? Then our team can investigate.

I noticed I can't actually reproduce the problem I originally had any more, which as far as I remember always gave me the NewTripId version for sncf without other options. At least now there are two separate entries in the Mobility Database, which seems correct. Maybe I just missed something when checking initially. (SNCF item for reference: https://transport.data.gouv.fr/datasets/horaires-sncf)

The second problem I noticed is at least one example of a static feed being listed as gtfs-rt. I don't really know what is causing that though. You can see it by trying to download https://mobilitydatabase.org/feeds/tdg-83330. The result is technically not valid GTFS because the files are in a subdirectory, but it is definitely not gtfs-rt.

PR #1807, opened 2025-12-31
Add colors Tram+Bus Augsburg

Schweigi491:
adds colors for tram+bus lines at Augsburg and fixes Dingo-lines at Dingolfing

### DIFF
--- a/scripts/de-delfi-colors.lua
+++ b/scripts/de-delfi-colors.lua
@@ -36,6 +36,15 @@ local colors = {
         ["SE"] = { color = 0xffffff, text_color = 0x000000 },
         ["SFEX"] = { color = 0x00a76d, text_color = 0xffffff },
     },
+    ["Anrufsammeltaxi1"] = {
+        ["70"] = { color = 0xa49892, text_color = 0xffffff },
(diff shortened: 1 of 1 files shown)


mlundblad:
Thanks!

mlundblad:
Thanks!

Sheigutn:
The Dingo line colors currently show up fine, but this change will break them on the next deploy since [de-DELFI.lua#L61](https://github.com/public-transport/transitous/blob/main/scripts/de-DELFI.lua#L61) expects the route names to have no spaces, which [generate-colors.py#L83](https://github.com/public-transport/transitous/blob/main/src/generate_colors.py#L83) intentionally removes.

Schweigi491:
Oh thanks, didn't know that. I will change it back

coffeeandtrains:
Hi there,
just to have some sort of reference, in case https://github.com/mfdz/GTFS-Issues/issues/211 eventually resolves: The PR added line colours for the S-Bahn Munich S8 line. This line (along with the S7 line above it) would need to be moved to the S-Bahn München agency.

PR #1809, opened 2025-12-31
revert DINGO-lines

Schweigi491:
revert last changes on DINGO-lines to the correct formatting

### DIFF
--- a/scripts/de-delfi-colors.lua
+++ b/scripts/de-delfi-colors.lua
@@ -697,12 +697,12 @@ local colors = {
         ["164"] = { color = 0xb17f4a, text_color = 0xffffff },
     },
     ["Bus8"] = {
-        ["Blauer DINGO"] = { color = 0x203b7c, text_color = 0xffffff },
-        ["Gelber DINGO"] = { color = 0xf3cb0f, text_color = 0x000000 },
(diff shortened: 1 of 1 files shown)


jbruechert:
Has this been done in the line-colors repository first? We are only a downstream consumer of that

Schweigi491:
In the line-colors the Dingos where added with the space (but not by me), then added here without the space.
I accidentially added the space because I didn't knew we need it without.
So should I remove the spaces in line-color as well?

jbruechert:
I can't and don't want to review the content of line colors changes, all commits here should just be regenerating the file from the upstream repository.

So I usually just blind-merge commits that don't touch anything apart from line colors, assuming that it was already reviewed upstream.

wittighausen:
As far as I understand, the problem here is the space between, for example, `Blauer` and `DINGO`, right?
If so, then it’s basically irrelevant, because in Träwelling the colors can be displayed either way, with or without the space, in the `de-delfi-colors.lua` file.

The same issue occurs, for example, in Kaiserslautern, where the city bus lines have an operator prefix with a space between it and the line number, and it works flawlessly, even though here it is written without the space.

And I think that with the next push of new line colors, this space will be removed by the script anyway.

luna-cant-code:
I've commented out one line of code and updated those colours - In theory, this should work and not break anything.

PR #1816, opened 2026-01-03
commented out space removing in DELFI data due to mapping issues

luna-cant-code:
Due to issue #1809 and my own experiences, it'd be best to disable the removal of spaces in DELFI line colours and pass them on as they're set in the DELFI dataset.

This should remedy mapping issues for services using Transitous

Screenshot 2026-01-03 at 12 11 22
Here's an example on how it looks currently in Träwelling (possibly) due to the removed space in the line name


mlundblad:
Thanks!

wittighausen:
I mean, in Pirmasens the city bus services are also written with a space, but there, the colors work like they should (And as far as I remember, the colors once worked for the DINGO lines too).
Screenshot 2026-01-03 at 15-04-22 RIS - Träwelling

luna-cant-code:
Funnily enough, Dingo didn't work for me ^^

But I think this code change won't be changing anything negatively

Sheigutn:
This PR will break a lot of line colors with spaces since de-DELFI.lua expects the lines without spaces, see [here](https://github.com/public-transport/transitous/blob/main/scripts/de-DELFI.lua#L61). 

The DINGO colors worked right after #1805 was merged, but stopped working once #1807 was merged and deployed, since both generate_colors.py and de-DELFI.lua replaced the spaces in the line names to match, which was the correct way to deal with spacing inconsistencies in line-colors.csv, but which will not be the case anymore after this PR. 

It seems to me that #1807 was done by manually editing de-delfi-colors.lua without regenerating it by using generate_colors.py, otherwise the "fix" would not have been needed. 

My suggestion would be to revert this PR and just to regenerate the de-delfi-colors.lua file. And maybe to make it more obvious that de-delfi-colors.lua should *not* be edited manually (though it does already have the *autogenerated by generate_colors.py* comment).

luna-cant-code:
> This PR will break a lot of line colors with spaces since de-DELFI.lua expects the lines without spaces, see here.

Ahhhh shit...Didn't notice that!
I'm going to open a revert PR

I'm very sorry for the inconveniences!!!

luna-cant-code:
@Sheigutn, just opened the PR - After that, everything should be back to normal

PR #1822, opened 2026-01-03
Remove Stadtwerke Landshut agency from DELFI feed

Sheigutn:
Trips in the city of Landshut are currently served by two agencies in DELFI for some reason, which results in some duplicate trips:

See also: https://github.com/mfdz/GTFS-Issues/issues/233

I decided to remove the agency with incomplete data that was added just recently until DELFI figures out what's going on. 

Image

### DIFF
--- a/feeds/de.json
+++ b/feeds/de.json
@@ -34,7 +34,8 @@
                 "FlixBus-de",
                 "EUROSTAR",
                 "Bremer Straßenbahn AG",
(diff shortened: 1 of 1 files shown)


jbruechert:
Thanks!
Please try to monitor this so we won't accidentally lose this agency completely once DELFI fixes this.

Sheigutn:
Will do!

PR #1823, opened 2026-01-03
Exclude foreign agencies from the swiss feed

wittighausen:
The reason for this are duplicated trips from the swiss feed and the other (correct) feeds.


jbruechert:
Thanks, makes sense!

PR #1825, opened 2026-01-05
th: fix typo of silom as silon

ExoSkye:
Accidentally made this typo earlier. Also - there appears to be an updated namtang feed now, should I remove the "extend-calender-dates" option from the feed config?


mlundblad:
Thanks!

mlundblad:
> Accidentally made this typo earlier. Also - there appears to be an updated namtang feed now, should I remove the "extend-calender-dates" option from the feed config?

Yeah, I guess we shouldn't use extend-calendar-dates unless it's actually needed.

PR #1826, opened 2026-01-05
th: remove extend-calender-dates from namtang feed

ExoSkye:
as mentioned in #1825. may be worth waiting until tomorrow (UTC) to merge this as mobilitydatabase hasn't updated their version of the feed yet.

### DIFF
--- a/feeds/th.json
+++ b/feeds/th.json
@@ -13,8 +13,7 @@
             "license": {
                 "spdx-identifier": "CC-BY-4.0"
             },
-            "script": "th-namtang.lua",
-            "extend-calendar": true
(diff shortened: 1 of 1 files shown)


jbruechert:
> as mentioned in #1825. may be worth waiting until tomorrow (UTC) to merge this as mobilitydatabase hasn't updated their version of the feed yet. 

We use the upstream url directly unless it is unavailable, so there should be no need to wait.
Our update pipeline also happens slightly later then the one of the Mobility Database, so we should be fine either way.

jbruechert:
Thanks!

PR #1829, opened 2026-01-07
PL: Fix Łódź and Żywiec, add Sulejówek

kasmar00:
- Fixes (`"fix":true`) Łódź and Żywiec (both had been failing since ~new years)
- Adds Sulejówek

And bonus: I added myself as maintainer to not miss those issues next time 😅

### DIFF
--- a/feeds/pl.json
+++ b/feeds/pl.json
@@ -15,6 +15,10 @@
         {
             "name": "Alexander Dubis",
             "github": "gtfscc"
+        },
+        {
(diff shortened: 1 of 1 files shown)


kasmar00:
Looking at the failures in CI: most of them come from expired feeds from girlc.at, I let the maintainer know and they will fix it (sooner or later 😅)

jbruechert:
Thanks!

PR #1835, opened 2026-01-09
Simplify Matrix channel references

vkrause:
Only list the Transitous channel, to reduce the risk of Transitous issues ending up in the general open-transport channel.

This is mostly a leftover from the early days, before we had our own channel.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,9 +30,7 @@ routing service that:
 
 ## Contact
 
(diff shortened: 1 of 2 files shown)


jbruechert:
Makes sense

vkrause:
We might also want to add the Matrix logo to the header, next to Mastodon/Github, but I failed to figure out how to do that.

PR #1837, opened 2026-01-10
nz: reinstate Flamingo Christchurch feed

applecuckoo:
they have relaunched in Christchurch since I removed their feed


mlundblad:
Thanks!

PR #1839, opened 2026-01-10
Initial support for configuring NeTEx feeds

vkrause:
None of the tested feeds are currently ready for production, but we are getting there for a few of them, and this allows us to test in the proper setup already.

### DIFF
--- a/docs/docs/index.md
+++ b/docs/docs/index.md
@@ -19,7 +19,7 @@ Free and open public transport routing.
 
 A community-run provider-neutral international public transport routing service.
 
-Using openly available GTFS/GTFS-RT/GBFS/etc. feeds and FOSS routing engine we want to operate a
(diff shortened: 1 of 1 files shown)


jbruechert (approved):

vkrause:
The CI error also happens locally without those changes, how should we proceed with that?

jbruechert:
I think this CI check should not even run after rebasing, as then the file has not changed compared to main.

vkrause:
> I think this CI check should not even run after rebasing, as then the file has not changed compared to main.

The file has change though, due to porting away from `gtfs-flex`. Should I disable the expired feed instead?

jbruechert:
> The file has change though, due to porting away from `gtfs-flex`. Should I disable the expired feed instead?

Oh right. Since the error is not caused by your changes, I don't think it needs to be fixed in this PR. You can go ahead and merge anyway.

PR #1840, opened 2026-01-11
Add Contributor Covenant Code of Conduct

felixguendling:


CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,86 @@
...
+
+Community leaders have the right and responsibility to remove, edit, or reject comments, commits, code, wiki edits, issues, and other contributions that are not aligned to this Code of Conduct, and will communicate reasons for moderation decisions when appropriate.
+
+## Scope
+
+This Code of Conduct applies within all community spaces, and also applies when an individual is officially representing the community in public spaces. Examples of representing our community include using an official e-mail address, posting via an official social media account, or acting as an appointed representative at an online or offline event.
+
+## Enforcement
+
+Instances of abusive, harassing, or otherwise unacceptable behavior may be reported to the community leaders responsible for enforcement at info@transitous.org. All complaints will be reviewed and investigated promptly and fairly.
```
jbruechert: Does this exist yet, and who do the e-mails go to?

felixguendling: at the moment it goes to @vkrause which should be sufficient for now
we can add forwarding to multiple people in the future

jbruechert: I wonder if it would not be better to list individual email addresses here instead, so it is transparent who is going to receive the mail.

felixguendling: also fine for me, should I add @jbruechert, @vkrause and me for now? I (or anyone else with write permission) can also add any other contact address :)
I think we would anyhow discuss things before taking action if problems would be reported.

felixguendling: Updated

PR #1848, opened 2026-01-15
rate limits with less delays

traines-source:
As dicussed in the matrix channel.

The MOTIS UI is quite wasteful with plan requests, I realized that if you repeatedly change options, the subsequent requests actually get slower due to the current nginx config (without nodelay/delay, nginx spaces out requests according to the rate limit, which is one request every three seconds).

burst means how many requests are accepted at once without rejecting, delay specificies the number of requests to not space out (nodelay would not space out anything, which is also bad if the burst number is close to the number of cores we have available). https://blog.nginx.org/blog/rate-limiting-nginx

Exact limits to be discussed.

Maybe we should also increase worker_connections to be able to handle more delayed requests at once. 

Not yet tested in production.

### DIFF
--- a/ansible/roles/nginx/tasks/main.yml
+++ b/ansible/roles/nginx/tasks/main.yml
@@ -9,7 +9,7 @@
 
 - name: Unpack MOTIS (frontend)
   unarchive:
-    src: https://github.com/motis-project/motis/releases/download/v2.6.2/motis-linux-amd64.tar.bz2
+    src: https://github.com/motis-project/motis/releases/download/v2.7.0/motis-linux-amd64.tar.bz2
     dest: /opt/motis/
     remote_src: yes
 


jbruechert:
Makes sense. Feel free to test with the nginx config for the staging instance.

traines-source:
Probably overthinking this, but
* probably it's better to return 429 than making the client wait for up to a minute -> changed some limits to that the maximum time to space out is ~30 seconds. For plan we could maybe also go for burst=20 and delay=10 as long as we have the capacity.
* tiles now also rate limited
* changed the default API limit from 42r/s to r/m. Was that intentional? The endpoints falling in this category (now) are map/initial, map/trips, one-to-many, rentals, which are all somewhat expensive, at least not less expensive than the harmless group.

jbruechert:
> * changed the default API limit from 42r/s to r/m. Was that intentional? The endpoints falling in this category (now) are map/initial, map/trips, one-to-many, rentals, which are all somewhat expensive, at least not less expensive than the harmless group.

I don't really remember, but 42r/s doesn't make much sense. 42r/m might be a bit low though.

traines-source:
With the staging instance, everything seems alright. Will roll it out to prod later today if nothing pops up.

felixguendling:
> The MOTIS UI is quite wasteful with plan requests

We could also add a "search" button to not waste requests the user didn't want.

traines-source:
As long as we have enough capacity, that's maybe not that high prio. But maybe one should also cancel requests and cancel actual calculation (IIRC that doesn't happen).
But at least now/soon MOTIS will not appear slower than it is due to that :)

PR #1859, opened 2026-01-17
Fix SNCB route types and display names

vkrause:


jbruechert:
Some Fosdem preparation? :D

PR #1861, opened 2026-01-18
Restore European Sleeper and fix its route type

vkrause:
The agency name in the nl-ovapi feed changed, also filtering them out. While at it, also correctly mark it as night rail.


jbruechert:
Thanks!

PR #1864, opened 2026-01-18
Add import script for fixing route types and display names in NL

vkrause:
There's still some room for improvement on regional services, but that requires a bit more research first.

### DIFF
--- a/scripts/nl-OpenOV.lua
+++ b/scripts/nl-OpenOV.lua
@@ -1,0 +1,37 @@
+-- SPDX-FileCopyrightText: Volker Krause <vkrause@kde.org>
+-- SPDX-License-Identifier: AGPL-3.0-or-later
+
+require "scripts.motis"
+
+-- source route type, route type name prefix, MOTIS class, target route type, display name prefix
+local route_type_map = {
+    { 2, "Intercity direct", LONG_DISTANCE, 102, "ICD" },
+    { 2, "Intercity", REGIONAL_FAST_RAIL, 103, "IC" },
+    { 2, "Sprinter", REGIONAL_FAST_RAIL, 106, nil },
+    { 2, "Sneltrein", REGIONAL_RAIL, 106, nil },
+    { 2, "Stoptrein", REGIONAL_RAIL, 106, nil },
+    { 2, "Nightjet", NIGHT_RAIL, 105, "NJ" },
+    { 2, "EuroCity direct", LONG_DISTANCE, 102, "ECD" },
+    { 2, "EuroCity", LONG_DISTANCE, 102, "EC" },
+    { 2, "ICE", HIGHSPEED_RAIL, 101, "ICE" },
+    { 2, "Eurostar", HIGHSPEED_RAIL, 101, "EST" },
+}
+
+function process_route(route)
+    for _,m in ipairs(route_type_map) do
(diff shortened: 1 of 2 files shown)


jbruechert:
Are the values taken from NS / physical departure boards?

vkrause:
The local departure boards actually only seem to show what's currently in the route name ("Intercity", "Sprinter", etc, without a number). That's not ideal for the cross-border routes though, we should IMHO have the same name for those everywhere.

The current implementation does this for long-distance services, but there's also local cross-border services. Those seem to be labeled as "Sneltrein [RE|RB]NN" for e.g. DB services (unlike the domestic ones which are just "Sneltrein"), and IMHO also should be renamed for consistency. I yet have to figure out whether that actually holds for all of them though.

scripts/nl-OpenOV.lua
```diff
@@ -0,0 +1,37 @@
+-- SPDX-FileCopyrightText: Volker Krause <vkrause@kde.org>
+-- SPDX-License-Identifier: AGPL-3.0-or-later
+
+require "scripts.motis"
+
+-- source route type, route type name prefix, MOTIS class, target route type, display name prefix
+local route_type_map = {
+    { 2, "Intercity direct", LONG_DISTANCE, 102, "ICD" },
+    { 2, "Intercity", REGIONAL_FAST_RAIL, 103, "IC" },
```
dancojocaru2000, line 9: I think the way things are displayed on the local departure boards should be what Transitous shows. I disagree that trains should be shown as "IC 3938" when someone at the train station looking at the departure board will see "Intercity" (or, when in the short form in the "next train" part, only "IC").

Photo of NS departure board for tracks 3b and 4b. On track 3b, ICE 152 towards Amsterdam Centraal leaves in less than 1 minute. On track 4b, the Intercity to Enkhuizen leaving at 18:53 has a 5 minute delay

scripts/nl-OpenOV.lua
```diff
@@ -0,0 +1,37 @@
...
+    { 2, "Sprinter", REGIONAL_FAST_RAIL, 106, nil },
+    { 2, "Sneltrein", REGIONAL_RAIL, 106, nil },
+    { 2, "Stoptrein", REGIONAL_RAIL, 106, nil },
+    { 2, "Nightjet", NIGHT_RAIL, 105, "NJ" },
+    { 2, "EuroCity direct", LONG_DISTANCE, 102, "ECD" },
+    { 2, "EuroCity", LONG_DISTANCE, 102, "EC" },
```
dancojocaru2000, line 15: On NS departure boards, EuroCity trains show up as "ECC 1234": 

Photo of an NS departure board showing the EuroCity ECC 9260 train leaving in 3 minutes towards Brussel Zuid

dancojocaru2000 (changes requested):

dancojocaru2000:
What could be done to get this PR merged (ideally with the changes I proposed as well)? It is rather irritating that I still can't use Transitous for regional train routing in the Köln-Frankfurt area because ICEs from the Dutch dataset show up.

felixguendling:
Your review comments (IC vs Intercity, ECC vs EC) were not addressed yet.
Otherwise, I think it should work and improve the situation.

jbruechert:
I would also prefer staying close to what local departure boards show.
IMO we should not copy DB Navigator, it is very Germany-centric after all.

PR #1867, opened 2026-01-20
fix incorrect 'spdx_identifier' to 'spdx-identifier' in some feeds

e-kotov:
Incorrect 'spdx_identifier' json key in some feeds resulted in `"spdx_license_identifier": null,` in https://api.transitous.org/gtfs/license.json


jbruechert:
Didn't realize this was so widespread, thanks!

PR #1872, opened 2026-01-21
Remove obsolete GTFS entries from pl.json and unskip Oleśnica

Lach-anonym:
Removed entries for Orneta and Suchowola, along with skip properties for a previous entry. Oleśnica seems to be up to date

I don't know why those feeds have disappeared, maybe the agency has changed there or these networks are now integrated into other feeds, not sure

### DIFF
--- a/feeds/pl.json
+++ b/feeds/pl.json
@@ -766,8 +766,6 @@
             "spec": "gtfs",
             "url": "https://files.girlc.at/gtfs/olesnica.zip",
             "fix": true,
-            "skip": true,
-            "skip-reason": "Expired",
             "license": {
                 "spdx-identifier": "CC0-1.0"
             }
@@ -934,15 +932,6 @@
             }
         },
         {
-            "name": "Orneta",
-            "type": "http",
-            "spec": "gtfs",
-            "url": "https://files.girlc.at/gtfs/orneta.zip",
-            "license": {
-                "spdx-identifier": "CC0-1.0"
-            }
-        },
-        {
             "name": "Legnica",
             "type": "http",
             "spec": "gtfs",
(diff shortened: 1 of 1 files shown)


jbruechert:
Thanks!

kasmar00:
Hi, the maintainer says that:
- Suchowola disapeared, cause its winter school break in the region, and unfortunately their code removes a feed on empty timetable, I will re-add when it's back 😅 
- Orneta - feed is now available under Tribus, as well as some other areas

Lach-anonym:
That's an awful situation for Suchowola - that means during summer the feed will neither be available...

jbruechert:
> Suchowola disapeared, cause its winter school break in the region, and unfortunately their code removes a feed on empty timetable, I will re-add when it's back 😅

We can add an option to ignore the feed failing to fetch on our side, so it can at least stay in the file.

Alternatively, if the feed maintainer agrees, the feed could have an up to date feed_end_date, while still not having a currently active service. I'd argue that is the correct way to model this case.
I realised the check on our side does not accept that, but fixing that is just replacing one `and` with an `or`.

kasmar00:
> That's an awful situation for Suchowola - that means during summer the feed will neither be available...

We had a similar case for Warsaw tourist ferries - feed was removed from transitous, and I'll readd it when the new timetable is announced.

> We can add an option to ignore the feed failing to fetch on our side, so it can at least stay in the file.
> Alternatively, if the feed maintainer agrees, the feed could have an up to date feed_end_date, while still not having a currently active service. I'd argue that is the correct way to model this case.
> I realised the check on our side does not accept that, but fixing that is just replacing one and with an or.

The feed for suchowola was reinstated, with operating window starting in a few days. I'll add it in a minute.

PR #1878, opened 2026-01-23
gl: Add Nuuk

mlundblad:


Lach-anonym:
Shouldn't it be dk_gl.json?

mlundblad:
> Shouldn't it be dk_gl.json?

Maybe, though, there's an ISO code for Greenland.

To cite Wikipedia: "The autonomous territories of [Faroe Islands](https://en.wikipedia.org/wiki/Faroe_Islands) and [Greenland](https://en.wikipedia.org/wiki/Greenland) have their own [ISO 3166-1](https://en.wikipedia.org/wiki/ISO_3166-1) codes and are not included in Denmark's entry in ISO 3166-2."

Lach-anonym:
Ah, fair enough

jbruechert:
This feed endpoint seems to be no longer correct. I could not easily find a replacement. Looks like an e-mail to the agency will be needed.

Lach-anonym:
Also I found out that Nuuk buses are already included in the Rejseplanen (general Danish) feed
Which is very funny, as in Rejseplanen HAFAS search, it's impossible to see the route on the map because of view boundaries ending in the middle of Faroe Islands

mlundblad:
> Also I found out that Nuuk buses are already included in the Rejseplanen (general Danish) feed Which is very funny, as in Rejseplanen HAFAS search, it's impossible to see the route on the map because of view boundaries ending in the middle of Faroe Islands

Ahh, I hadn't realized that!
Yeah, indeed they seem to appear.

I'll close this for now.

PR #1883, opened 2026-01-25
Add Mobi app to website

extroversiv:
Let me know if you are OK with me using the transitous logo in the mobi logo and or where else you would like more acknowledgement.


jbruechert:
> Let me know if you are OK with me using the transitous logo in the mobi logo and or where else you would like more acknowledgement.

I see no problem with that :)
Looks like a cool app!

PR #1886, opened 2026-01-26
Add mobi app to website

extroversiv:
The svg image was not displayed correctly, e.g., on my phone, probably due to missing svg reference to the font. Changed it to png.


jbruechert:
I'm fine with PNG as well, but Inkscape (and probably most svg editors) has the option to convert the text to a path, after which the font is no longer needed.

extroversiv:
> I'm fine with PNG as well, but Inkscape (and probably most svg editors) has the option to convert the text to a path, after which the font is no longer needed.

Yeah, I did that too, but then I thought, a png is fine.

jbruechert:
Hmm, maybe the new PNG is a bit low-res looking at it on the website.

PR #1888, opened 2026-01-26
Change Mobi icon to higher resolution

extroversiv:
OK, one more try with 500 pixels. Sorry for the extra loop.


jbruechert:
Thanks :)

PR #1896, opened 2026-01-31
fix: es: Fix Metro Bilbao GTFS

josebama:
Change the Metro Bilbao feed to the one coming from the regional transport consortium (CTB).
It's the source nap.transportes.gob.es was getting their data from, but somehow they stopped getting updates, as the latest update is from December 2026, and they only have data until 15th of January 2026. While the feed from CTB keeps getting daily updates. So this change cuts the middleman and gets the up to date data directly from the source.

### DIFF
--- a/feeds/es.json
+++ b/feeds/es.json
@@ -1153,10 +1153,6 @@
             "name": "Metro-Bilbao",
             "type": "http",
             "url": "https://ctb-gtfs.s3.eu-south-2.amazonaws.com/metrobilbao.zip",
-            "fix": true,
-            "http-options": {
-                "fetch-interval-days": 1
-            },
             "license": {
                 "spdx-identifier": "CC-BY-4.0",
                 "url": "https://data.ctb.eus/en/pages/legal-notice"


josebama:
I tested it locally and I got this

```
Fetching es-Metro-Bilbao…
Postprocessing es-Metro-Bilbao with gtfsclean…
Parsing GTFS feed in '/transitous/out/.tmp-es_Metro-Bilbao.gtfs.zip' ... done. (0 trips [0.00%], 0 stop times [0.00%], 0 stops [0.00%], 0 shapes [0.00%], 0 services [0.00%], 0 routes [0.00%], 0 agencies [0.00%], 0 transfers [0.00%], 0 pathways [0.00%], 0 levels [0.00%], 0 fare attributes [0.00%], 0 translations [0.00%] dropped due to errors.)
Removing trips travelling too fast...done. (-0 trips [-0.00%])
Removing service duplicates... done. (-0 services [-0.00%])
Outputting GTFS feed to '/transitous/out/.tmp-es_Metro-Bilbao.gtfs.zip'... done.
```

I assume that's that it's working?

Does the line saying `(0 trips [0.00%], 0 stop times [0.00%], 0 stops [0.00%], 0 shapes [0.00%], 0 services [0.00%], 0 routes [0.00%], 0 agencies [0.00%], 0 transfers [0.00%], 0 pathways [0.00%], 0 levels [0.00%], 0 fare attributes [0.00%], 0 translations [0.00%] dropped due to errors.)` mean that this feed no longer needs the `fix: true` option?

jbruechert:
> Does the line saying `(0 trips [0.00%], 0 stop times [0.00%], 0 stops [0.00%], 0 shapes [0.00%], 0 services [0.00%], 0 routes [0.00%], 0 agencies [0.00%], 0 transfers [0.00%], 0 pathways [0.00%], 0 levels [0.00%], 0 fare attributes [0.00%], 0 translations [0.00%] dropped due to errors.)` mean that this feed no longer needs the `fix: true` option?

Yes, looks like it. Best way to confirm is running it once without fix: true. You might need to remove out/feed.zip in order to force it to reprocess it.

feeds/es.json
```diff
@@ -1152,17 +1152,14 @@
...
             "type": "http",
-            "url": "https://nap.transportes.gob.es/api/Fichero/download/1267",
+            "url": "https://ctb-gtfs.s3.eu-south-2.amazonaws.com/metrobilbao.zip",
             "fix": true,
             "http-options": {
-                "fetch-interval-days": 1,
-                "headers": {
-                    "ApiKey": "e8143a1e-d12e-4a01-9b08-caf3071e4853"
-                }
+                "fetch-interval-days": 1
```
jbruechert: This is only needed if required by terms and conditions, or if the server does not provide a last-modified header.

josebama: I just checked and the response came with the last-modified header, and the provider doesn't specify any poll/download limit in their terms of use, so I've removed it

josebama:
> > Does the line saying `(0 trips [0.00%], 0 stop times [0.00%], 0 stops [0.00%], 0 shapes [0.00%], 0 services [0.00%], 0 routes [0.00%], 0 agencies [0.00%], 0 transfers [0.00%], 0 pathways [0.00%], 0 levels [0.00%], 0 fare attributes [0.00%], 0 translations [0.00%] dropped due to errors.)` mean that this feed no longer needs the `fix: true` option?
> 
> Yes, looks like it. Best way to confirm is running it once without fix: true. You might need to remove out/feed.zip in order to force it to reprocess it.

I removed the fix option, cleared my local out folder and it seemed to work fine

jbruechert:
Thank you!

PR #1898, opened 2026-02-01
Website: Add HA Departures as using app

jfowl:
See https://github.com/alex-jung/ha-departures/blob/b22797785f60d3de6ec1cdf4c8f4e868a5c10a26/README.md?plain=1#L15

image


jbruechert:
Thanks!

PR #1901, opened 2026-02-02
cz:IDS-JMK new rt feed

hoermalmeister:


hoermalmeister:
also added new IDZK-rt

feeds/cz.json
```diff
@@ -61,7 +61,7 @@
             "name": "IDS-JMK",
             "type": "url",
             "spec": "gtfs-rt",
-            "url": "https://content.idsjmk.cz/kestazeni/gtfsReal.dat"
+            "url": "https://owncloud.cesnet.cz/index.php/s/DkvuaInMj1JsfB2/download"
```
jbruechert, line 64: Can you please explain why the new RT is only available through your Nextcloud?
Is this a feed you generate yourself?

hoermalmeister: Yes, I generate it from their api, which provides also delays, not only locations

jbruechert:
Thanks :)

feeds/cz.json
jbruechert: Okay!
Just wanted to make sure we correctly document the origin of the file, in case someone else wants to investigate if it breaks.

PR #1902, opened 2026-02-02
Add it:Viterbo-Cotral feed

hoermalmeister:
https://cotralspa.it/open-data/

### DIFF
--- a/feeds/it.json
+++ b/feeds/it.json
@@ -610,6 +610,11 @@
             "name": "Veneto-Venezia-Ferries",
             "type": "transitland-atlas",
             "transitland-atlas-id": "f-u20f-actvspa"
+        },
+        {
+            "name": "Viterbo-Cotral",
+            "type": "http",
+            "url": "https://travel.mob.cotralspa.it:4443/GTFS/GTFS_COTRAL.zip"
         }
     ]
 }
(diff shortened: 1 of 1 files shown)


hoermalmeister:
... I just noticed we have this feed actually included as https://www.transit.land/feeds/f-cotral~lazio~italia, can anyone see, why it doesn't show up in the map?

applecuckoo:
@hoermalmeister If you're asking why it isn't in Transitous, Transitland Atlas feeds don't automatically import here. If you're asking why it's not working on the Transitland side, it's because the TLS certificate being served from upstream isn't really set up correctly.

hoermalmeister:
Well I understand that, but we actually have it cofigured in it.json

 {
            "name": "Lazio-Cotral",
            "type": "transitland-atlas",
            "transitland-atlas-id": "f-cotral~lazio~italia",
            "fix": true
        },
        
And the static url from transit.land is working just fine, it only doesn't show up in api.transitous.org

jbruechert:
The url produces an ssl error currently:
```
Error: Could not fetch it-Lazio-Cotral: [('url', SSLError(MaxRetryError('HTTPSConnectionPool(host=\'travel.mob.cotralspa.it\', port=4443): Max retries exceeded with url: /GTFS/GTFS_COTRAL.zip (Caused by SSLError(SSLCertVerificationError(1, "[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: Hostname mismatch, certificate is not valid for \'travel.mob.cotralspa.it\'. (_ssl.c:1081)")))'))), ('cache-url', Exception('Could not fetch file. HTTP Status code: 404'))]
```

You can try ignoring tls errors.

PR #1910, opened 2026-02-04
Also make the community events list data-driven

vkrause:
Easier to maintain and automatically drops past events.


jbruechert:
Thanks!

PR #1912, opened 2026-02-05
website: Add Transportia as an app using Transitous

TomakataABC:
Source code to the app is right here: https://github.com/Wafler1/transportia


jbruechert:
Thanks!

vkrause:
I don't seem to be able to get to the source code from the play store site linked here, maybe we should link to the Github repo instead (which does have a link to the play store, so nothing lost there)?

TomakataABC:
Fair enough, i can create another PR with that change. I just linked directly to the play store since it's the only platform for the app and thus downloading it was easier this way.

felixguendling:
I agree with @vkrause. I tried to find the source code repository using Google and at least on my page 1, it didn't show up. So for transparency it would be better to link to the source code.

vkrause:
> Fair enough, i can create another PR with that change. I just linked directly to the play store since it's the only platform for the app and thus downloading it was easier this way.

Either that, or a link from the play store page to the source code in some way would also work I think.

felixguendling:
@TomakataABC could you please replace the link with the GitHub link in a new PR? Thank you :-)

TomakataABC:
@felixguendling I forgor, it's now #1981 submitted

PR #1913, opened 2026-02-05
Add Lua scripts for SKPL and Arriva trains

Lach-anonym:


jbruechert:
Why? The PKP Intercity feed stops at the border, so does not fully replace this.

Lach-anonym:
There are duplicates of domestic trains, that run to the stations near the border
Also crossborder trains are covered already by @kasmar00's feed

Lach-anonym:
It's really embarrassing, I cannot seem to find the edit button for the PR name 😳

kasmar00:
🤔 I was actually thinking of removing my feed, as poland-germany Intercity trains are now in delfi and delfi has realtime support. But that would require modyfing the delfi script to set correct names for those trains.

Btw. Your drop won't work fully, as the German part of pl-de trains have operator set to DB Fernverkher.

kasmar00:
> Why? The PKP Intercity feed stops at the border, so does not fully replace this.

There are two PKP Intercity feeds. The "PKP Intercity Germany" has full trains to Germany (as it's a filtered gtfs.de), and the PKP Intercity stops at the border.

Lach-anonym:
Screenshot_20260205_195229_Vivaldi

The screenshot shows departures from Rzepin station

I've looked at the trains that got duplicated and I found out that... these are the main trains for carriages that run between Przemyśl/Kraków and Leipzig and get detached/connected in Wrocław! DELFI for some reason decided to provide data for those trains even if they do not cross the DE-PL border, same case for IC Ursa to/from Świnoujście

So the problem is a bit more complex

Lach-anonym (commented):
Official timetables: https://rozklad.com/maps/index.php?IDKlienta=ILAWA_ZKM

Lach-anonym:
The ICs are still duplicated, any ideas on how to proceed?

jbruechert:
In my opinion we should not solve duplicated departures by hand unless there is an agency-wide feed that replaces data in some local aggregation.

I still think the most realistic solution for this Transitous-wide is to implement fuzzy merging, and I plan to work on that already. Particularly at borders, there is often not even a clear responsibility as to which operator should be the canonical source.

PR #1914, opened 2026-02-05
Revert "Remove Stadtwerke Landshut agency from DELFI feed"

Sheigutn:
Reverts public-transport/transitous#1822. Latest DELFI feed now only contains the later agency.


jbruechert:
Thanks for keeping track of this :)

PR #1916, opened 2026-02-06
Add Trenitalia NeTEx feed

vkrause:
This gives us high-speed/long-distance services in Italy, as well as a bunch of additional regional trains.

This does seem to cause a few duplicate regional trains in northern Italy, overall a big improvement still though.

Needs MOTIS > 2.8.1.

### DIFF
--- a/feeds/it.json
+++ b/feeds/it.json
@@ -7,6 +7,13 @@
     ],
     "sources": [
         {
+            "name": "trenitalia",
+            "type": "http",
+            "url": "https://public-transport.github.io/transitous-trenitalia-preprocessor/trenitalia-netex.zip",
+            "spec": "netex",
+            "script": "it-trenitalia.lua"
+        },
+        {
             "name": "Calabria-CORE",
             "type": "mobility-database",
             "mdb-id": "mdb-2668",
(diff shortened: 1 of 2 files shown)


mlundblad:
So, this needs to wait for a MOTIS upgrade, right? (I think currently Transitous is running on 2.7.6)

mlundblad:
> So, this needs to wait for a MOTIS upgrade, right? (I think currently Transitous is running on 2.7.6)

Ahh, ok, yeah this was already discussed in Matrix…

vkrause:
Right, and not only that, something newer than 2.8.1 isn't even released yet.

hoermalmeister:
So are we able to add netex feeds directly now? In the past days i was looking into the netex files on https://www.cciss.it/nap/mmtis/public/catalog/Dataset and for a way to convert them into gtfs files, with mixed results, because they have slightly different structure (coordinates, names of stations in different fields, etc)

vkrause:
> So are we able to add netex feeds directly now?

In theory, yes, the import pipeline is prepared for NeTEx feeds meanwhile. However, so far all NeTEx feeds either didn't work or regressed in some way or the other compared to GTFS, or required changes in MOTIS. See also the NeTEx section in the meeting notes of https://github.com/public-transport/transitous/wiki/Transitous-Hack-Weekend-Berlin,-January-2026. This would be the first one being added.

If the other Italian feeds use the same format/profile as the Trenitalia one we might be lucky though.

jbruechert:
Thanks, exciting to add the first NeTEx feed :)

PR #1928, opened 2026-02-08
Add RT for Bydgoszcz

kasmar00:
(Sidenote: Lufthansa did a tutorial on pronouncing the name of this city: https://www.facebook.com/watch/?v=1011482391196427 😛)

### DIFF
--- a/feeds/pl.json
+++ b/feeds/pl.json
@@ -151,6 +151,15 @@
             "transitland-atlas-id": "f-u3ky-bydgoszcz"
         },
         {
+            "name": "Bydgoszcz",
+            "type": "url",
+            "spec": "gtfs-rt",
+            "url": "https://next.zbiorkom.live/api6-open/bydgoszcz/gtfsRealtime/default/tripUpdates",
+            "license": {
+                "spdx-identifier": "MIT"
+            }
+        },
+        {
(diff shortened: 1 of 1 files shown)


jbruechert:
Thanks!

DomeQdev:
This GTFS-realtime feed is different from the ones available on cdn.zbiorkom.live.
The feeds on cdn.zbiorkom.live come from the v5 backend, whereas
https://next.zbiorkom.live/api6-open/bydgoszcz/gtfsRealtime/default/tripUpdates
is served by the v6 backend (the next version).

In v6, instead of uploading static files to the Nginx CDN, the feeds are generated directly by the backend on demand and then cached for 20 seconds on the Cloudflare side. This approach is better than the previous one, because my CDN is currently being heavily hammered by around 20 different MOTIS instances.

Additionally, I now provide only tripUpdates, since MOTIS does not support vehiclePositions anyway. I also introduced delay propagation within the same block, because MOTIS has issues handling this properly.

I’m using the next.zbiorkom.live domain because api.zbiorkom.live is currently occupied by the v5 backend. After the v5 backend is deprecated, next.zbiorkom.live/api* will continue to work without redirects for one year, and after that it will become a normal redirect.

PR #1929, opened 2026-02-08
Add feed for morocco

CarlSchwan:
In preparation for my trip to Morocco soon :)  This is includes the only feed available in GTFS for ONCF (Office National des Chemins de Fer)

I also found another feed for Tétouan by  @bouhormq but it's 4 years old and likely quite outdated (see https://www.trufi-association.org/how-trufi-code-launched-an-app-and-career/)

### DIFF
--- a/feeds/ma.json
+++ b/feeds/ma.json
@@ -1,0 +1,18 @@
+{
+    "maintainers": [
+        {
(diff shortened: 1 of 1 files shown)


mlundblad:
Thanks!

PR #1932, opened 2026-02-10
Add support for SIRI feeds

vkrause:


src/generate-motis-config.py
```diff
@@ -30,6 +30,12 @@ def find_motis_asset(asset_name: str):
 def check_file_exist_in_out_folder(file_name: str):
     return os.path.isfile(os.path.join("out", file_name))
 
+def to_motis_rt_spec(spec: str) -> str:
+    match spec:
+        case 'gtfs-rt': return 'gtfsrt'
+        case 'siri-json': return 'siri_json'
```
jbruechert, line 36: Is this missing a case for siri (xml)?

vkrause: that's `siri` on both sides and thus handled by the default path below, but we can also make that explicit I guess.

jbruechert: Ah sorry, alright then

PR #1936, opened 2026-02-12
Include Madrid buses again

pacastega:
Two feeds for the Madrid region buses were being skipped because they were broken / expired. Both are fixed now.

### DIFF
--- a/feeds/es.json
+++ b/feeds/es.json
@@ -134,7 +134,7 @@
                 "headers": {
                     "ApiKey": "b2be1741-ed99-41e7-99bf-924b32694633"
                 }
-            },
+            }
         },
         {
             "name": "Autobuses-Junta-de-Extremadura",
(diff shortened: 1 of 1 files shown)


feeds/es.json
```diff
@@ -128,11 +132,9 @@
             "http-options": {
                 "fetch-interval-days": 1,
                 "headers": {
-                    "ApiKey": "e8143a1e-d12e-4a01-9b08-caf3071e4853"
+                    "ApiKey": "b2be1741-ed99-41e7-99bf-924b32694633"
                 }
             },
```
mlundblad: Stray comma left here

feeds/es.json
```diff
@@ -522,11 +524,9 @@
...
```
mlundblad: Stray comma left here

mlundblad:
The expired feed is not part of this PR, so I think we can merge this for now.

mlundblad:
Thanks!

pacastega:
> The expired feed is not part of this PR, so I think we can merge this for now.

I wasn't sure if I was supposed to fix the entire file before doing a pull request so I didn't touch those other feeds. I can fix those as well as a separate PR.

PR #1944, opened 2026-02-14
Fix Blablacar Bus route type

vkrause:
MOTIS v2.8 doesn't have the coach detection heuristic anymore, so those were now only shown as regular busses.


jbruechert:
Thanks!

PR #1948, opened 2026-02-15
Fix coach route types for most Flixbus feeds

vkrause:
The US is still missing, as that's slightly more complex.

This also simplifies the import script to no longer touch clasz, as that's not only not necessary when setting an extended route type.


jbruechert:
Nice! The same script should also apply to `ee-kaugliinid-rahvusvaheline`

vkrause:
Yep, added for that as well now.

PR #1949, opened 2026-02-15
Fix coach route type for Flixbus and Greyhound in the US

vkrause:


feeds/us.json
```diff
@@ -32,13 +32,15 @@
             "name": "Brightline",
             "type": "transitland-atlas",
             "transitland-atlas-id": "f-brightline~trails~rt",
-            "use-feed-proxy": true
+            "use-feed-proxy": true,
+            "script": "generic-bus-to-coach.lua"
```
jbruechert: This is an RT feed, should probably belong to the feed above?

vkrause: Looking at while this had seemingly worked during testing, Brightline is actually a rail service, I had simply not expected that in the US :D

PR #1955, opened 2026-02-16
Consistently use set_route_type with set_clasz

kasmar00:
Due to https://github.com/motis-project/motis/issues/1294 only setting `clasz` does not work. This PR consistently sets route_type along with clasz in scripts for Poland. Also added comments which const to use when updating to a future motis relase (containing this commit: https://github.com/motis-project/motis/commit/d3aeade35f4b7172b25555cc9f8949a6ef317f36)


jbruechert:
Thanks!

PR #1960, opened 2026-02-17
hu-mav: use shapes

gy-mate:
They seem okay now.

_This reverts commit 18397b8._


mlundblad:
Thanks!

(the error seems to be unrelated to this change, AFAICS)

PR #1967, opened 2026-02-19
Liechtenstein feed now obsolete

hoermalmeister:
malbun bergbahnen is now included in swiss national data portal


jbruechert:
Nice, thanks :)

PR #1968, opened 2026-02-20
Mention that the map view now supports GBFS feeds

1Maxnet1:
Recently the map view gained support for showing sharing information. This change reflects this in the documentation accordingly.


vkrause:
Thank you!

PR #1970, opened 2026-02-20
Remove set_clasz() calls in Lua import scripts

vkrause:
Replace those with set_route_type() where that's not already done anyway.

Extended route types have a higher level of detail, and overwrite classes already anyway. Future MOTIS versions will therefore deprecate the class API entirely.


jbruechert:
Thanks!

PR #1974, opened 2026-02-20
gb: set train route types in lua script

schaerfo:
~~Setting route types does not work \\:~~ Turns out I used an older MOTIS revision

Also, I didn't test the script with transitous tooling

### DIFF
--- a/scripts/gb-great-britain.lua
+++ b/scripts/gb-great-britain.lua
@@ -1,3 +1,6 @@
+-- SPDX-FileCopyrightText: Christian Schärf <transitous-dYvaXYH2@christian.schaerf.email>
+-- SPDX-License-Identifier: MIT
+
 -- Unfortunately, the feed lumps routes together by franchise, making fine-grained type assignment impossible for the time being
 local route_type_map = {
     -- Caledonian Sleeper


vkrause:
Tested in the Transitous setup, it does actually seem to work, I see the long distance and night rail types being applied (didn't check all of the rest).

scripts/gb-great-britain.lua
```diff
@@ -0,0 +1,66 @@
...
+    ["AW"] = { route_type = 106 },
+    -- West Midlands Railway
+    ["WM"] = { route_type = 106 },
+}
+
+function process_route(route)
+    local route_id = route:get_id()
+    if route_type_map[route_id] then
+        local route_type = route_type_map[route_id].route_type
+        print("assigning route type", route_type)
```
felixguendling:
```suggestion
```

schaerfo: Right, I missed that one

felixguendling (approved):

scripts/gb-great-britain.lua
```diff
@@ -0,0 +1,65 @@
+-- Unfortunately, the feed lumps routes together by franchise, making fine-grained type assignment impossible for the time being
```
felixguendling, line 4:
```suggestion
-- SPDX-FileCopyrightText: FirstName LastName <email@example.com>
-- SPDX-License-Identifier: AGPL-3.0-or-later

-- Unfortunately, the feed lumps routes together by franchise, making fine-grained type assignment impossible for the time being
```

jbruechert:
Thanks!

PR #1976, opened 2026-02-21
Add infrastructure for generating a "meta station" GTFS feed

vkrause:
Meta stations allow using e.g. "London" or "Paris" as destination while referring to a set of (main) stations in the city. This affects sufficiently few cities that we can maintain this mapping manually, as this only really produces the result for multiple terminus stations with limited connectivity between each other on long-distance modes.


vkrause:
For working properly this needs the latest MOTIS version, in current production this works in theory but geocoding will down-rank the meta-stations beyond findability (unless searching for them in rare languages like Latin).

PR #1978, opened 2026-02-22
Replace DottBerlin with DottDE

### LINKED ISSUE #1977: Add Dott Germany feed
The `https://gbfs.api.ridedott.com/public/v2/countries/de/gbfs.json` URL is the Dott GBFS feed for the entirety of Germany. 

It is also possible to use a city name instead of `countries/de` to only get the feeds for that city. Currently, Transitous includes the Berlin-only feed (`https://gbfs.api.ridedott.com/public/v2/berlin/gbfs.json`). I think it should be replaced with the Germany-wide feed.

dancojocaru2000:
The file at `countries/de` contains the GBFS for all of Germany, as opposed to just Berlin. Fixes #1977

### DIFF
--- a/feeds/de.json
+++ b/feeds/de.json
@@ -227,9 +227,9 @@
             "spec": "gbfs"
         },
         {
-            "name": "DottBerlin",
+            "name": "DottDE",
             "type": "url",
-            "url": "https://gbfs.api.ridedott.com/public/v2/berlin/gbfs.json",
+            "url": "https://gbfs.api.ridedott.com/public/v2/countries/de/gbfs.json",
             "spec": "gbfs"
         },
         {


jbruechert:
Thanks!

felixguendling:
This is not a standard format and will therefore probably not really work like this, but maybe @pablohoch knows more?

jbruechert:
Should we revert for now?

pablohoch:
> This is not a standard format and will therefore probably not really work like this, but maybe @pablohoch knows more?

Haven't seen that format before, but it won't work since we expect the "name" keys to be unique, so it'll just load one for each feed (the last one, currently russelsheim).

felixguendling:
I think it would be best to have a little script (probably easier to do outside of MOTIS) that transforms it into a proper manifest feed.


> Should we revert for now?

It won't crash anything I think, so it's not urgent to remove it. But to leverage the data, we need to either add support in MOTIS or transform it into a supported format outside  of MOTIS (like what citybik.es does with GBFS v3 manifest).

pablohoch:
> But to leverage the data, we need to either add support in MOTIS or transform it into a supported format outside of MOTIS (like what citybik.es does with GBFS v3 manifest).

Or just add all the cities from that file individually, e.g. https://gbfs.api.ridedott.com/public/v2/hennef/gbfs.json etc.

traines-source:
We do have some of these cities already, e.g. de-DottFrankfurt or weirdly fr-tier-dott-gbfs-casgbs contains Dott Rüsselsheim...

PR #1982, opened 2026-02-23
Update DELFI colors

wittighausen:


wittighausen:
closed due to duplication with #1983

PR #1987, opened 2026-02-26
hu: add Budaörs GTFS

gy-mate:
And move BKK's GTFS-RT right next to its Schedule feed.


gy-mate:
_The failing import is for another feed._

mlundblad:
Thanks!

PR #1988, opened 2026-02-26
hu: add Kiskunhalas GTFS

gy-mate:


gy-mate:
_The failing import is for another feed._

mlundblad:
Thanks!

PR #2002, opened 2026-03-09
hu: add BAHART GTFS

gy-mate:
Add the GTFS of the Lake Balaton ferry company.

### DIFF
--- a/feeds/hu.json
+++ b/feeds/hu.json
@@ -93,6 +93,11 @@
             }
         },
         {
+            "name": "BAHART",
+            "type": "http",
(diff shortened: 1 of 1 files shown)


felixguendling:
They don't set `cars_allowed` in their feed so maybe we can overwrite this manually in case all ferries also transport cars.

gy-mate:
Oh, I see.

Only one of their lines (Szántód–Tihany) transports cars.

gy-mate:
...but—upon inspection—that route is missing from their GTFS (https://github.com/gy-mate/gtfs-feeds/issues/14) so the overwrite is not needed at the moment.

felixguendling:
In case we have this case again I can add scripting support for `cars_allowed`.

felixguendling (approved):

PR #2004, opened 2026-03-10
bg: Add Varna feed

BPilot253:


BPilot253:
Fixed the calendar, hopefully no more issues now

mlundblad:
It looks like there's still an issue with calendar.txt (missing start_date)

BPilot253:
I accidentally had some aggressive caching for the GTFS files so maybe that is why it isn’t updating

mlundblad:
Thanks!

PR #2005, opened 2026-03-11
Add Bolzano STA

vkrause:
Found accidentally while exploring their NeTEx feed.

This includes adding support for downloading feeds from FTP, which unfortunately seems to be the only way this is available.


mlundblad:
Danke!

As far as I can see, the other failures should be unrelated to this new feed.

PR #2009, opened 2026-03-13
Improve RENFE trips legibility

Lach-anonym:
Set text color for the route to default value.

### DIFF
--- a/scripts/es-renfe.lua
+++ b/scripts/es-renfe.lua
@@ -24,6 +24,7 @@ function process_route(route)
             break
         end
     end
+    route:set_text_color(0)
 end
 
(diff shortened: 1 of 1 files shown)


jbruechert:
This should already be fine since the contrast checking feature. Where is the problem still visible?

image

vkrause:
The web UI in dark mode doesn't seem to handle this correctly though, there it's white on white.

jbruechert:
> The web UI in dark mode doesn't seem to handle this correctly though, there it's white on white.

Ah, I think the issue here is that the text colour is unset.
It would probably make sense to extend the contrast check to fix that, but in the meantime the script is fine of course.

PR #2023, opened 2026-03-20
Add Metra to us-il

kasmar00:
The RT feed sadly requires registering for an API key, and as I'm not from the area, I don't feel like maintaining one 😅 https://metra.com/metra-gtfs-api


jbruechert:
Thanks!

PR #2038, opened 2026-03-21
Add fahrtle as website using transitous

bovine3dom:
Link to github rather than the deployed URL because it will probably change :)


bovine3dom:
Sorry, I misunderstood how REUSE worked. I've linted it locally now and it should work - although I get a warning about line 124 of src/generate_colors.py

jbruechert:
Thanks, looks fine now :)

PR #2039, opened 2026-03-21
reverse proxy: OJP endpoint

traines-source:
at least temporarily pass through OJP endpoint. moving to /api/ojp20 instead to be discussed.
already deployed.

### DIFF
--- a/ansible/roles/nginx/templates/transitous.conf.j2
+++ b/ansible/roles/nginx/templates/transitous.conf.j2
@@ -90,6 +90,18 @@ server {
         add_header Cache-Control "public, max-age=36000";
     }
 
+    # OJP endpoint
+    location /ojp20 {
+        proxy_pass http://{{ item.vhost }}-backend/ojp20;
+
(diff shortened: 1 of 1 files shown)


jbruechert:
Is there any way we can tell different request types apart with OJP?

traines-source:
Nope unfortunately not really AFAIK, or not in a straightforward way. That's why I set it to a rather restrictive rate limiting (but it is not as restrictive as the most expensive endpoints).

jbruechert:
> Nope unfortunately not really AFAIK, or not in a straightforward way. That's why I set it to a rather restrictive rate limiting (but it is not as restrictive as the most expensive endpoints).

okay, we can probably still build something more complicated if it becomes an issue.

felixguendling:
I don't expect any significant traffic to OJP.
The only way to use it right now is via this demo UI: https://motis-project.github.io/ojp-demo-app-src/search
The link is not shared in a lot of places right now.

PR #2041, opened 2026-03-22
website: add MMM-PublicTransportHub app card and icon to Transitous homepage

KristjanESPERANTO:
Adds [MMM-PublicTransportHub](https://github.com/KristjanESPERANTO/MMM-PublicTransportHub), a module for [MagicMirror²](https://magicmirror.builders/), to the Transitous app showcase.

PR #2042, opened 2026-03-23
website: update Ananke and streamline website build/deploy

KristjanESPERANTO:
I noticed a few low-risk website cleanups while working on PR #2041, so this PR ships those separately.

No intended behavior change for transitous.org.

Live preview: https://kristjanesperanto.github.io/transitous/

Tested locally with `npm run start` and `npm run release`.

Note: this PR originally included a PureScript -> TypeScript widget migration. That part was split out and is not included here.

### DIFF
--- a/website/hugo.toml
+++ b/website/hugo.toml
@@ -11,6 +11,7 @@ theme = "ananke"
   custom_css = ["bootstrap.min.css", "bootstrap-icons.min.css", "custom.css"]
   favicon = "images/logo.svg"
   text_color = "#292e35"
+  post_content_classes = ["avenir"]
 
 [params.ananke.social.follow]
   networks = ["mastodon", "matrix", "github"]


jbruechert:
Thank you for looking into updating the Hugo theme, that is definitely very useful.

I'd have preferred to discuss rewriting the widget in another language before you started working on it, because if I'm going to keep being the one who maintains the website code, I'd like it to not be a flavour of JavaScript. I simply do not enjoy working with it.

KristjanESPERANTO:
Sorry. I sometimes tend to go a bit to far ahead, but I don’t mind taking back some steps :slightly_smiling_face: 

My goal was to simplify the build/setup, not to force a stack you don’t want to maintain. I’ll take another look at it and either remove the migration from this PR or open a new one for the Hugo theme update.

KristjanESPERANTO:
@jbruechert I removed the TypeScript migration :slightly_smiling_face:

jbruechert:
Thank you!

I pushed one change to restore the previously used font

PR #2044, opened 2026-03-24
Add new transport data sources from files.girlc.at

Lach-anonym:
MKuran website is down, hope it gets back online at some point


Lach-anonym:
We need to address the auth issue for Koleje Dolnośląskie feed, last fetch on Transitland being a month ago makes me worried...

jbruechert:
@kasmar00 Would you be open to reviewing this one? I'm afraid my insight in Polish data sources is a bit limited :)

Lach-anonym:
Well I've reminded myself that we'll need to deprecate several Polish rail operator feeds in one week due to changes on MKuran's side and running a single PKP feed (more info at his website). This might help with the KD auth issue.

kasmar00:
Hi,

@jbruechert 
> @kasmar00 Would you be open to reviewing this one? I'm afraid my insight in Polish data sources is a bit limited :)
- I believe `piecki` and `morag` are fine to be removed, they were replaced by `tribus` 🟢 
- I asked the maintainer about `siechnice` - the feed on zbiorkom.live is produced by girlc.at, and just rehosted there
- I nudged about `plonsk` which has expired

@Lach-anonym 

> We need to address the auth issue for Koleje Dolnośląskie feed, last fetch on Transitland being a month ago makes me worried...

This is a strange case, they seem to have changed the setup recenlty: https://github.com/transitland/transitland-atlas/pull/1877. As I understand, transitous fetches from transitland, so It's best to ask them what's going on.
A quick fix could be to use https://cdn.zbiorkom.live/gtfs/pkp-kd.zip, as it could be more reliable or use the auth basic and url they provide on the website: https://kd.kiedyprzyjedzie.pl/gtfs/kd_sample/google_transit.zip + auth basic: `kd_sample:Hm5mEuPC?1h7`

> Well I've reminded myself that we'll need to deprecate several Polish rail operator feeds in one week due to changes on MKuran's side and running a single PKP feed (more info at his website). This might help with the KD auth issue.

The update from the maintainer of those feeds is, that they will try to keep up, until it requires any work on their side.

I planned to remix the unified train feed (https://github.com/kasmar00/gtfs-polish-trains), to have platform data and add shapes. The only thing missing on my side is deployment and shape data for PKP Intercity (as it's a long distance operator it's not always obvious for the routers which path to take, and current feed https://github.com/kasmar00/PKPIntercityGTFS uses data that has extra, non passenger "control points" for routing). I'm also waiting to get data from Koleje Mazowieckie (their `RL` bus is missing in the open train data) without reply for about a month 😢

I don't have a good solution right now, as using just the unified train feed will result in worse shapes for PKP Intercity. 
An acceptable solution could be to use https://mkuran.pl/gtfs/polish_trains.zip (and the corresponding RT feed) with PKP Intercity dropped out and keep PKP Intercity feed as is for now (which is hosted by me, and not going anywhere at the moment). And sometime in the near future switch to my version of unified train feed, once it's deployed.

Lach-anonym:
Another option is to drop the shapes in unified feed, so they will be generated by MOTIS
However I haven't checked how properly does it work for trains so far

jbruechert:
>  A quick fix could be to use https://cdn.zbiorkom.live/gtfs/pkp-kd.zip, as it could be more reliable or use the auth basic and url they provide on the website: https://kd.kiedyprzyjedzie.pl/gtfs/kd_sample/google_transit.zip + auth basic: kd_sample:Hm5mEuPC?1h7

Using the sample credentials seems to work for me, sounds like a good solution.

I think the MOTIS shapes should probably be okay for PKP Intercity in practise. I don't think anyone will complain if the train takes a slightly different path than on the map :)

derhuerst:
> MKuran website is down, hope it gets back online at some point

cc @MKuranowski

Lach-anonym:
That's not really needed, as it got back quite briefly after I wrote it
But at least the pipeline runs again

PR #2047, opened 2026-03-25
website: add dark mode via head-additions partial

KristjanESPERANTO:
I think dark mode for the website would be nice.

Functionality:

- detect OS color scheme preference
- a sun/moon toggle button is injected into the navbar via JS
- persist user choice in localStorage

Live preview: https://kristjanesperanto.github.io/transitous/

What do you think?

### DIFF
--- a/website/layouts/partials/head-additions.html
+++ b/website/layouts/partials/head-additions.html
@@ -1,0 +1,41 @@
+<!--
+SPDX-FileCopyrightText: 2026 Kristjan Esperanto
+
+SPDX-License-Identifier: AGPL-3.0-or-later
+-->
+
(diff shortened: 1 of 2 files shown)


website/layouts/partials/head-additions.html
```diff
@@ -0,0 +1,41 @@
+<!--
+SPDX-FileCopyrightText: 2026 Transitous contributors
```
jbruechert: Please rather add your name here, we don't have a file that actually keeps track of who that is

KristjanESPERANTO: Okay, done.

PR #2049, opened 2026-03-28
Add Amarillo ride sharing feed for DE-BW

vkrause:
Needs API key setup to actually work.


vkrause:
API key removal process for the static feed has been set up, just the proxy for the realtime part missing now.

PR #2051, opened 2026-03-28
us-fl: add Gainesville RTS

twassman:
GTFS file is not dynamically updated (valid for Spring 2026 only), so may need to be updated seasonally.


mlundblad:
Thanks!

Yeah, a quick check seemed to indicate that accessing just /uploads gave a 403, otherwise maybe it could have been possible to have a script scape the index for links (I think we have something along those lines, using some Python HTML parsing library).

Issue #2048, opened 2026-03-27
Question: Add GTFS color if the PTA doesn't submit it?


vikiawv:
Hey dear maintainers, Is this possible somehow?

felixguendling:
It's possible, see for example:
https://github.com/public-transport/transitous/blob/main/scripts/de-delfi-colors.lua

vikiawv:
So if I understand it correctly I just need to make a PR for for example this file?

_at-linz-ag-colors.lua_
```
local colors = {
    ["Linz Linien GmbH"] = {
        ["1"] = { color = 0xDC487F, text_color = 0xffffff },
        ["2"] = { color = 0xBD80B1, text_color = 0xffffff },
        ["3"] = { color = 0x952F8A, text_color = 0xffffff },
        ["4"] = { color = 0xB32653, text_color = 0xffffff },
        ["50"] = { color = 0x3E8D53, text_color = 0xffffff },
    },
}
return colors
```
Is there something else that needs to be added in another file for this to function?
Sorry for these questions, I am not that great when it comes to coding :)

felixguendling:
The actual code that sets the colors is here:

https://github.com/public-transport/transitous/blob/70098d2f545cd82ca62b89b9a45b8c795518d1ce/scripts/de-DELFI.lua#L64-L68

You need to make sure that the agency name and line names match exactly what you find in the GTFS.

I would recommend to up a MOTIS instance locally with one dataset to test and make sure the script works as expected.

vikiawv:
Of course it would be easier to ask the PTA directly, which is what I did first actually - they are willing to do so - but they seem to be stuck with implementation for an as they said undefined period of time, as they have unnecessary bureaucracy that they need to follow due to their supervisors.

So I manually prepared some files and double-checked them to match what is inside the GTFS files (attachments). However I am not 100% sure if the filenames are correct, as I was not able to test it yet on my machine yet, as my building process on both windows and linux keeps constantly failing for always other reasons (even before adding any feeds/osm-files). It is pretty frustrating.
It could probably be a hardware- or skill-issue, as I never was able to build anything successfully on this PC.

----
[at-eastern-region-flex.txt](https://github.com/user-attachments/files/26344389/at-eastern-region-flex.txt)
[at-eastern-region-flex-color.txt](https://github.com/user-attachments/files/26344386/at-eastern-region-flex-color.txt)
[...]

PR #2061, opened 2026-04-03
Assign precise train types to NL OVapi trains

dancojocaru2000:


felixguendling:
LGTM. I guess you already tried this locally with MOTIS to check if it works?

dancojocaru2000:
I haven't, I haven't found instructions on how to run the script locally so I just based my edits on other scripts already in the repo, mainly the `es-renfe.lua` one.

felixguendling:
It should be easy to do like this (download the motis binary, run `./motis import` and `./motis server` -> check if everything works as expected):
https://github.com/public-transport/transitous/issues/2048#issuecomment-4176229878

I'm not sure how we can otherwise check scripts before merge.

vkrause:
See also #1864.

dancojocaru2000:
> It should be easy to do like this

I unfortunately don't find it easy. I wanted to get this finished so I started looking into it again, and there's very little instructions to get motis running. Transitious seems to use json config files, but motis seems to use a single yaml file? And, as far as I can tell, the yaml file is documented only in [`motis/docs/setup.md`](https://github.com/motis-project/motis/blob/master/docs/setup.md)?

dancojocaru2000:
I'll close this PR in favour of #1864 (which I didn't notice before opening), but I still thing the way to actually test these changes is - at best - quite badly documented.

traines-source:
The documentation for a quick start with a single feed is documented right at the [MOTIS root readme](https://github.com/motis-project/motis#quick-start). But one could definitely make that clearer in the [Transitous docs](https://transitous.org/doc/) how to test just a single feed or region with MOTIS.

dancojocaru2000:
I genuinely don't see where you linked how I can configure it to use the lua config file to test it. Can you please copy-paste that part? All I can see is how to use an unmodified GTFS feed.

traines-source:
Yes that's separately explained here https://github.com/motis-project/motis/blob/master/docs/scripting.md but you're definitely right that one could better connect these explanations one to another and improve them in general, be it in the MOTIS or Transitous repository.

dancojocaru2000:
Thank you! There definitely needs to be a documentation improvement, because even in the [Advanced Setups](https://github.com/motis-project/motis/blob/master/docs/setup.md) docs file, which seems to describe the config format, the `script` key is just not mentioned.

PR #2066, opened 2026-04-03
use age for asymmetric encryption of feed API keys

traines-source:
Migrating to age for asymmetric API key encryption, also supporting static feeds. Supports encrypting HTTP header values and URLs.

Opted against SOPS because it doesn't allow fine-granular selection of values/fields to be encrypted.
Opted against age's --armor / PEM encoded format because it is a hassle with the multilines and header and instead manually encoded with base64.

I forgot again that using (GitHub) secrets is not possible in PRs from forks (for good reasons). We could do weird shit to work around that. I however think it may be acceptable that such PRs altering feed configs containing encrypted API keys for static feeds will be red before merging.

To encrypt a new API key: 
`echo AGE-ENCRYPTED:$(echo -n "my-api-key" | age -r age1xpexzyk4k608tsccev20d7uakptkmgxr4nc8s0azzywkgtz86aaqsdryr4 | base64 -w0)`

(Also added explanation to docs)

TODO
* add the `transitous_feed_proxy_key` to woodpecker and hope it works


src/generate-motis-config.py
```diff
@@ -180,12 +180,14 @@ def to_motis_rt_spec(spec: str) -> str:
                                     config["timetable"]["datasets"][name]["rt"] = []
 
                                 rt_feed: dict[str, Any] = {
-                                    "url": source.url if use_original_url else FEED_PROXY + '/feed/' + quote(name) + "-" + str(len(config["timetable"]["datasets"][name]["rt"])),
+                                    "url": decrypt_if_necessary(source.url) if use_original_url else FEED_PROXY + '/feed/' + quote(name) + "-" + str(len(config["timetable"]["datasets"][name]["rt"])),
```
jbruechert, line 183: What is the point of encrypting the url, if it can end up unencrypted in the public config.yml?

PR #2070, opened 2026-04-08
Change drop_shapes default to True

felixguendling:
Most shapes contained in datasets have very low quality or are also routed on OSM like we do (= no improvement).

Therefore the default should be not to use them, and if we find a dataset with better quality shapes than those we automatically compute via map matching on OSM data, it should be easy to keep the original shapes.

Just looking at Germany, we get low quality shapes for FlixTrain, from Austrian datasets that have partial shapes (only routed in Austria), in Berlin there are buses flying through buildings, etc.

I agree that there might be cases where a dataset has better shapes than what MOTIS would compute. However, the default is that I browse through the debug view to find bugs in our map matching and almost every shape that looks bad turns out to have `TIMETABLE` as source. So we would definitely benefit from dropping them by default in order to improve average shape quality.

Edit: another option would be to drop a lot of shapes (manually going through a lot of datasets and setting the flag) and monitor this for all new datasets. Since nobody is willing to do this, the better option is to drop the shapes by default and for the rare cases where we know that the dataset has better shapes, we can put an effort to disable `drop_shapes`.

### DIFF
--- a/src/metadata.py
+++ b/src/metadata.py
@@ -50,7 +50,7 @@ class Source:
     skip_reason: str = ""
     function: Optional[str] = None
     drop_too_fast_trips: bool = True
-    drop_shapes: bool = False
+    drop_shapes: bool = True
     drop_agency_names: List[str] = []
     keep_agency_names: List[str] = []
     display_name_options: Optional[DisplayNameOptions] = None


jbruechert:
I don't think this makes sense, then we can also use the MOTIS option to reroute everything.
Nobody will compare the shape quality of the original shapes if they are not visible anywhere, and additionally we make the re-published GTFS data worse for everyone.

However I fully agree that the shapes from the Austrian Railway dataset don't provide any advantage, we can absolutely drop them.

I could not easily find the problem with the FlixTrain shapes, what did you notice there?

felixguendling:
>  then we can also use the MOTIS option to reroute everything

That would be my preferred option as well but I can understand that if a person puts extra effort in better shapes they should be able to see them in Transitous. However, that's usually not the case and the default quality is bad + just creates more work on our side without any gain over the default option `drop_shapes=true`.

Not sure what's happening there exactly:

Screenshot 2026-04-08 at 21 56 04

But my goal was not to go through all of those cases until we reach a usable state. Since they are mostly worse than what we have, changing the default option has more advantages while still leaving the option for special cases that are an exception to the rule.

felixguendling:
image

Except the ferries I think all of them are timetable source.

manfredbrandl:
I would prefer to recalculate all shapes that are not excluded from the recalculation. There are cases where the automatic routing needs to be adjusted, as we want the data to include both the current route around a construction site and the future route through the site once it has been completed.

PR #2072, opened 2026-04-09
Polish feed updates

Lach-anonym:
I've changed some URLs from deprecated cdn.zbiorkom.live
Finally there is an available GTFS-RT feed for Gdynia, however:
- Zbiorkom does not differenciate Gdynia network from Gdańsk
- GTFS-RT only contains Gdynia updates, so the Gdańsk feed has to stay (good approach tbh)
- the IDs in GTFS-RT do not match ones in the official feed from ZDIZ Gdynia, so that requires a change of main Gdańsk static feed which has now both networks combined, Gdańsk IDs do match and there will be no problem with their RT
- I've put both networks feed under Gdańsk name to avoid further issues (duh)
Yes, I live in hell

PS: Gdynia has issues today, so lack of most of vehicles is completely normal, maybe tomorrow it will be fine again

### DIFF
--- a/feeds/pl.json
+++ b/feeds/pl.json
@@ -270,8 +270,11 @@
         },
         {
             "name": "Gdańsk",
-            "type": "transitland-atlas",
-            "transitland-atlas-id": "f-ztm~gdansk",
+            "type": "http",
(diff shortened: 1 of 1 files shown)


jbruechert:
Thanks :)

PR #2080, opened 2026-04-14
Add license information to Piemonte-Bus entry

jjasloot:
License is specified on https://www.dati.piemonte.it/#/catalogodetail/regpie_ckan_ckan2_yucca_sdp_smartdatanet.it_Servizioprogrammatodeltrasportopubblicoregionepiemonteautobus_22942

### DIFF
--- a/feeds/it.json
+++ b/feeds/it.json
@@ -144,7 +144,10 @@
             "name": "Piemonte-Bus",
             "type": "transitland-atlas",
             "transitland-atlas-id": "f-u0j-piemonte~autobus",
-            "fix": true
+            "fix": true,
+            "license": {
+                "spdx-identifier": "CC-BY-4.0"
+            }
         },
         {
             "name": "Piemonte-Gruppo-Torinese-Trasporti",


jjasloot:
Hmm, the url for 'it-Piemonte-Bus' `http://api.smartdatanet.it/api/Servizioprogrammatodeltrasportopubblicoregionepiemonteautobus_22942/attachment/22941/1/GTFS_BUS_IT_PIE.zip` does seem to work on my end, but throws an error in the pipeline. The same also applies to 'it-Piemonte-Trenitalia' `http://api.smartdatanet.it/api/ServizioProgrammatoDelTrasportoPubblicoRegionePiemonteTreniRegionali_5188/attachment/5187/1/GTFS_RAIL_IT_PIE.zip`

What's especially weird is the 404 being returned, that would be some "interesting" rate limiting or geoblocking.

Do you happen to have any ideas?

jbruechert:
Adding the license is good in any case, it might work in the actual pipeline, as that is not GitHub Actions related and has a different IP range.

PR #2081, opened 2026-04-14
Change docs for Motis config generation for test instance

apiote:
I was adding a new feed and wanted to set up a test instance. After following the docs to create the config file, Motis wouldn't run; `datasets` file was empty. I managed to find another method to generate a working config.

I'm not sure if it's a me-problem, or if the docs might be outdated. Feel free to change, or decline this PR if the current method works and the problem was only on my side.


jbruechert:
I think you ran into a bug in the config generation script, which I fixed in edd646e83648a850086a4cedabf368c7e47a5647.

I verified that the documentation works again locally.

apiote:
I've pulled your changes and the script works on my side too. I'll close this PR. Thanks

PR #2082, opened 2026-04-14
Add My Ride in Rock Hill, SC, USA

apiote:
My Ride is in Bimba, but not in Transitous. Bimba's moving away from separate feeds and going Transitous-only, so I don't want anybody to lose access to their timetable data


apiote:
I've set up a test Motis instance and checked endpoints geocode and stop_times

jbruechert:
Thanks!

PR #2084, opened 2026-04-14
Fix Gdynia feed, add url-override for zbiorkom legacy feeds and add Żuławy narrow gauge railway

Lach-anonym:
I found out that I did not properly check the trip IDs for Gdynia and after running a GTFS-RT validator, the official feed actually works. So I reverted things to how they have been before and finally Gdynia should have working real time, I apologize for that inconvenience.

For zbiorkom feeds that do not support the new API links, I've created the URL overrides, so when they get migrated, the change can be made briefly.

Also I've decided to create a GTFS file for the ŻKD narrow gauge railway (my first serious attempt!). ~~There are no shapes, but I don't think it's crucial here.~~ (now there are shapes) Because of how rare are the changes (once yearly), I don't think it requires to be automated. Feel free to give feedback!

### DIFF
--- a/feeds/pl.json
+++ b/feeds/pl.json
@@ -270,12 +270,8 @@
         },
         {
             "name": "Gdańsk",
-            "type": "http",
-            "url": "https://api.zbiorkom.live/api6-open/tricity/gtfs/default",
-            "license": {
-                "spdx-identifier": "MIT"
-            },
-            "fix": true
+            "type": "transitland-atlas",
+            "transitland-atlas-id": "f-ztm~gdansk"
         },
         {
             "name": "Gdańsk",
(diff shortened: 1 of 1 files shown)


feeds/pl.json
```diff
@@ -1569,6 +1566,15 @@
             "license": {
                 "spdx-identifier": "CC0-1.0"
             }
+        },
+        {
+            "name": "Żuławska-Kolej-Dojazdowa",
+            "type": "http",
+            "spec": "gtfs",
+            "url": "https://github.com/Lach-anonym/gtfs-collection/blob/main/zkd.zip",
```
jbruechert: The link doesn't directly lead to the zip but to the GitHub download page

Lach-anonym: Thanks, I've fixed that

jbruechert:
Thanks!

PR #2088, opened 2026-04-15
Drop bad shapes for couple feeds in PL

kasmar00:


feeds/pl.json
```diff
@@ -111,6 +111,7 @@
             "name": "Koleje-Mazowieckie",
             "type": "transitland-atlas",
             "transitland-atlas-id": "f-u3-kolejemazowieckie",
+            "drop-shapes": true,
```
kasmar00, line 114: Shapes here are not terrible, but the bus shapes follow rails, yielding funny results: 

Image

feeds/pl.json
```diff
@@ -236,6 +237,7 @@
             "license": {
                 "spdx-identifier": "MIT"
             },
+            "drop-shapes": true,
```
kasmar00, line 240: Here, the shapes are just straight lines from station to station

Issue #619, opened 2024-11-10
 Support ssim format for flights?


Altonss:
From issue https://github.com/motis-project/motis/issues/652 on Motis :)

Data for airlines doesn't seem to be available very often as gtfs formats. But it is often available in ssim format, for example https://transport.data.gouv.fr/datasets?locale=en&type=air-transport . Some territories are simply not well connected other then by plane.

Would it make sense to add support for it in Transitous? (see for example https://github.com/planarnetwork/ssim2gtfs to convert into gtfs, although this converter doesn't seem in very active development)

jbruechert:
I would be open to accepting a PR to add support for it, but personally I'm currently mostly interested in trains / ferries / buses.

vkrause:
EU funded analysis on SSIM <-> GTFS and SSIM <-> NeTEx conversion: https://www.sign-air.eu/Deliverables/SIGN-AIR_D2.21_v00.01.pdf - nothing groundbreaking though.

felixguendling:
Are there free SSIM datasets that cover large parts of the world?
Or would we have to find one for each airline separately?

vkrause:
Seems to be by airline, and the only ones I have found so far are the ones on https://transport.data.gouv.fr/datasets?format=SSIM, which are all expired.

kalon33:
> Seems to be by airline, and the only ones I have found so far are the ones on https://transport.data.gouv.fr/datasets?format=SSIM, which are all expired.

Air France KLM dataset has been been updated today for Summer and Winter seasons.

I found this project https://github.com/planarnetwork/ssim2gtfs to convert SSIM to GTFS. Maybe a solution to use the dataset?

PR #2090, opened 2026-04-17
Add Air France/KLM flight schedule

vkrause:
Converted from the SSIM schedule on transport.data.gouv.fr and airport/ airline data from Wikidata, using https://github.com/public-transport/ssim-converter/.

Depends on https://github.com/public-transport/gtfsclean/pull/18.

Implements #619.

### DIFF
--- a/feeds/fr.json
+++ b/feeds/fr.json
@@ -13983,6 +13983,15 @@
             "managed-by-script": true,
             "skip": true,
             "skip-reason": "Invalid GTFS, not zipped"
+        },
+        {
+            "name": "airfrance-klm",
+            "type": "http",
+            "url": "https://public-transport.github.io/ssim-converter/AF-KL-TO-HV-S26-W26.zip",
+            "license": {
+                "url": "https://transport.data.gouv.fr/datasets/programme-des-vols-air-france",
+                "spdx-identifier": null
+            }
         }
     ]
 }


traines-source:
Have we verified that this doesn't dominate too many train trips or leads to considerably longer query times? Should we disable the airplane mode by default before merging this?

Lach-anonym:
It could be great to differentiate public service from purely economical one, so for example in Norway the itinerary could still show some flights that were not made strictly to compete with land transport. This requires some (probably extensive) manual work...

Take notice that the same situation happens at railways - let's take Poland as an example, where there are IC and EIC trains. The first category is financed by state, the second is run at financial risk. Both are intercities and in most cases have generally the same travel times, even if EIC is "express". But IC is available at reasonable price and EIC is ridiculously expensive.

Then there are coaches - let's say we disable them in the itinerary because we hate ALSA/Blablacar/Flixbus/etc. Some regional feeds however have their local lines classified as coaches (like Volánbusz), so this affects reaching the destination.

One of the solutions can be agency/line filtering in MOTIS, which has been mentioned a couple of times already...

jbruechert:
Should we disable the airplane mode by default before merging this?

IMO yes

felixguendling:
> Should we disable the airplane mode by default before merging this?
> 
> IMO yes

for the MOTIS Debug UI we can do this

for the API it would be a breaking change that requires API versioning but waiting for clients to switch to a new API version doesn't really make sense because then they could also disable airplane mode if they don't want flights

jbruechert:
I think the web interface should be good enough as long as we notify maintainers of the relevant apps.

vkrause:
Disabling flights by default makes sense IMHO, we already have the case in Norway with flights dominating everything else. Our clients actually do that already.

The other problem is connectivity between the flight and ground networks. That's also not new, but becomes more apparent the more flights are added. I was hoping that the terminal-accurate data would help with this (e.g. CDG T2 flights are now practically on the same location as the CDG railway station), and if so we could apply e.g. the entrance heuristics we have in KDE's code already to the airport locations here as well.

However, Marcus' test results on Matrix from yesterday suggest that this even still fails for CDG, so this needs further investigation. I'm relying on the test instance for this though, as given the global scope you basically need all ground transport and full OSM data to properly test things.

vkrause:
Some observations from the test instance:
* performance does not seem to have degraded by the global connectivity, and e.g. Berlin <-> Tokyo door to door routing actually works, with the results not being completely off (within the below limits)
* routed transfers must be on for anything to work even half-way properly
* routed transfers do not seem to eliminate the minimum transfer times between flights (good)
* lack of flight network <-> ground network connectivity is the biggest problem by far, and the main reason for weird results
* the terminal-accurate positions help with connecting to the ground transport network, but are not enough to connect neighboring terminals (e.g. CDG 2E > 2F isn't done as a walk but a CDG > AMS > LHR > CDG flight loop). Extra entries in transfers.txt might help with that?
* moving the airport center coordinates to the assumed airport entrance will be necessary for airports without terminal-accurate data to have a chance of working at all
* dominating ground transport is a problem, but currently you need to try hard to actually trigger this given the above issues

Lach-anonym:
I gave it a try and these are my observations so far:
- Gdańsk airport coords connect to a bus stop on the other side of runway. Note: all flights use terminal 2.
- Connections at Amsterdam airport are weird

Screenshot_20260420_210438_Vivaldi

https://test.motis-project.org/?time=2026-04-20T19%3A03%3A00.000Z&fromPlace=pl-PKP-SKM_257530&toPlace=kl-flights_CDG&withFares=true&fastestDirectFactor=1.5&joinInterlinedLegs=false&useRoutedTransfers=true&maxMatchingDistance=250&fromName=Gda%C5%84sk+Port+Lotniczy%2C+Gda%C5%84sk%2C+wojew%C3%B3dztwo+pomorskie%2C+Polska&toName=Paris-Charles+de+Gaulle+%28CDG%29%2C+Roissy-en-France%2C+%C3%8Ele-de-France%2C+Francja

vkrause:
There's a new version now that uses the improved airport coordinates we use in KDE Iitnerary, which attempt to find the "entrance" with OSM-based heuristics, in places where we don't have terminal-accurate data. This should significantly improve e.g. Amsterdam, Gdańsk and Stockholm.

Also, https://public-transport.github.io/ssim-converter/ now has a debug view for all the stop points. Click on them to see multiple overlapping points as well as links to the Wikidata items providing the terminal location (if any).

vkrause:
Deployed on test instance now, connectivity on many airports significantly improved (routed transfers are still mandatory though), flights dominating other alternatives starts to be the dominating problem now.

Another issue becoming more prominent now are creative solutions to bypass the 1h minimum transfer times, such as leaving the airport for a quick trip around the parking lot with a local bus. The screenshot above shows this in AMS, to allow a 15min connection there.

Lach-anonym:
Would it be possible to add a one hour buffer every time when entering an airport and a connecting minimum time when staying in it? This might partially fix the dominance issue

vkrause:
> Would it be possible to add a one hour buffer every time when entering an airport and a connecting minimum time when staying in it? This might partially fix the dominance issue

Check-in times are currently not possible, but are on the roadmap for MOTIS, as e.g. Eurostar has the same need. Minimum transfer times are in the generated GTFS feed here already, but the router has found a way around them unfortunately. But yes, that would certainly help with some of the dominance issues, especially on shorter distances.

felixguendling:
related https://github.com/motis-project/nigiri/issues/339
I think the fix mentioned there might work

PR #2099, opened 2026-04-24
pl:RPK Bochnia now works

hoermalmeister:
the authors have repaired the issue with their feed, so i can halt the rehost


mlundblad:
Thanks!

Seems the other failures right are unrelated to this change :)

PR #2102, opened 2026-04-25
Add "Public transit search for overnight train journeys" thesis

vkrause:
Cites Transitous and benchmarks against Transitous results.


jbruechert:
Cool!

PR #2103, opened 2026-04-25
Add PKS Jarosław feed

kzx1337:
Add PKS Jarosław feed


jbruechert:
The download link does not seem to be working

kzx1337:
I accidentaly left repository private. Should be fixed now.

PR #2113, opened 2026-04-30
add spdx identifier for gy-mate feeds

MrKrisKrisu:
see https://github.com/gy-mate/gtfs-feeds?tab=GPL-3.0-1-ov-file#readme


jbruechert:
Thanks!

PR #2114, opened 2026-04-30
Replace Porto SCTP source

PatrickSteil:
Sorry for the formatting issues in the JSON and Python files. If preferred, I can revert those changes and keep only the functional updates.

Changes:
Updated the data source for `Porto-STCP,` as the previous `Transitland` source was outdated.
Implemented logic to automatically scrape the latest GTFS feed URL in `src/region_helpers.py`.

### DIFF
--- a/src/region_helpers.py
+++ b/src/region_helpers.py
@@ -14,33 +14,38 @@ import requests
 
 
 def mvo_keycloak_token(source: HttpSource) -> HttpSource:
-    token = requests.post("https://user.mobilitaetsverbuende.at/auth/realms/dbp-public/protocol/openid-connect/token", data={
+    token = requests.post(
+        "https://user.mobilitaetsverbuende.at/auth/realms/dbp-public/protocol/openid-connect/token",
+        data={
             "client_id": "dbp-public-ui",
             "username": "5f7xgv6ilp@ro5wy.anonbox.net",
             "password": ")#E8qE'~CqND5b#",
             "grant_type": "password",
-            "scope": "openid"
-        }).json()["access_token"]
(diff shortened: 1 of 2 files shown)


jbruechert:
Please at least restore the formatting to the usual formatting. You can use `./src/bulk-edit-feeds.py feeds/pt.json`, even though that's not the original purpose of the script :)

PatrickSteil:
@jbruechert thx; i ran the bulk-edit-feeds script. the python script is formatted using [black](https://github.com/psf/black), which runs in my vim setup automatically. Pls let me know if you want me to revert the formatting there as well, but imo the change is "small" enough :)

src/region_helpers.py
```diff
@@ -140,23 +147,98 @@ def data_metroporto_latest_resource(source: HttpSource) -> HttpSource:
         )
     }
 
-    html = requests.get(source.url, headers=headers).text
+    html = requests.get(source.url, headers=headers, timeout=30).text
     soup = BeautifulSoup(html, "html.parser")
 
-    li = soup.find("li", class_=lambda c: c and "last zip" in c)
+    li = soup.select_one("li.last.zip")
```
jbruechert, line 153: This seems unrelated, but looks fine as well. Is it in this PR intentionally?

PatrickSteil: yes, i did change it to not rely on the exact spacing of  the string "last zip" in the class name. i noticed that the class name actually was "... last zip " (one extra whitespace at the end), and hence i thought it may be better to select the element this way `select_one("li.last.zip")`.

PR #2115, opened 2026-04-30
add license information

MrKrisKrisu:
1. In [traewelling/transitous-licenses](https://github.com/Traewelling/transitous-licenses/blob/main/licenses.json) we've collected a few licenses for feeds. Primarily those that can't be defined using standard SPDX identifiers. However, over time, some generic licenses have also ended up there, and I'd prefer to see them here in the repo instead. :)

~~2. For transitland feeds where a liense is specified in [their DMFR specifications](https://github.com/transitland/transitland-atlas/tree/main/feeds), I've copied the identifier here, so it can be processed here as well.~~

### DIFF
--- a/feeds/ro.json
+++ b/feeds/ro.json
@@ -15,6 +15,7 @@
             "type": "http",
             "url": "https://jbb.ghsq.de/gtfs/ro-railway.gtfs.zip",
             "license": {
+                "spdx-identifier": "OGL-ROU-1.0",
                 "url": "https://data.gov.ro/base/images/logoinst/OGL-ROU-1.0.pdf"
             }
         },
--- a/feeds/es.json
+++ b/feeds/es.json
@@ -845,7 +845,10 @@
         {
             "name": "Ferrocarriles-de-la-Generalitat-de-Cataluña",
             "type": "mobility-database",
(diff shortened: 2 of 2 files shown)


jbruechert:
There is no need to copy the information from transitland, the entire point is that it is automatically retrieved by the ID. The generate-attribution.py script already passes it through into the license.json.

MrKrisKrisu:
Oh, sorry, I wasn't paying close attention. I removed the Transitland licenses, but two identifiers would remain.

(And I’d check to make sure we’re using the licenses at Träwelling more efficiently so we don’t have to maintain them twice. Sorry for the inconvenience.)

feeds/ro.json
```diff
@@ -15,6 +15,7 @@
             "type": "http",
             "url": "https://jbb.ghsq.de/gtfs/ro-railway.gtfs.zip",
             "license": {
+                "spdx-identifier": "OGL-ROU-1.0",
```
jbruechert, line 18: This is not an actually existing SPDX-identifier

```
Error: Unknown license key(s): OGL-ROU-1.0
```

jbruechert: Maybe we can somehow maintain our own extensions to the license database, needs research

PR #2116, opened 2026-04-30
Add CATS in Charlotte, NC, USA

leo60228:
Tested locally, directions and live vehicle locations both work.

### DIFF
--- a/feeds/us-nc.json
+++ b/feeds/us-nc.json
@@ -60,6 +60,22 @@
             "transitland-atlas-id": "f-dnru-chapel~hill~transit~nc~us~rt"
         },
         {
+            "name": "CATS",
+            "type": "transitland-atlas",
+            "transitland-atlas-id": "f-dnq-charlotteareatransitsystem",
+            "license": {
+                "spdx-identifier": "CC0-1.0"
+            }
+        },
+        {
+            "name": "CATS",
+            "type": "transitland-atlas",
+            "transitland-atlas-id": "f-dnq-charlotteareatransitsystem~rt",
+            "license": {
+                "spdx-identifier": "CC0-1.0"
+            }
+        },
+        {
             "name": "DukeTransit",
             "type": "transitland-atlas",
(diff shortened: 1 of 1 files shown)


leo60228:
There does not seem to be a stated license for this data, but North Carolina law ([N.C.G.S. § 132-1](https://www.ncleg.gov/EnactedLegislation/Statutes/HTML/ByChapter/Chapter_132.html)) would imply that it's public domain:

> (a) "Public record" or "public records" shall mean all documents, papers, letters, maps, books, photographs, films, sound recordings, magnetic or other tapes, electronic data-processing records, artifacts, or other documentary material, regardless of physical form or characteristics, made or received pursuant to law or ordinance in connection with the transaction of public business by any agency of North Carolina government or its subdivisions. Agency of North Carolina government or its subdivisions shall mean and include every public office, public officer or official (State or local, elected or appointed), institution, board, commission, bureau, council, department, authority or other unit of government of the State or of any county, unit, special district or other political subdivision of government.
> 
> (b) The public records and public information compiled by the agencies of North Carolina government or its subdivisions are the property of the people. Therefore, it is the policy of this State that the people may obtain copies of their public records and public information free or at minimal cost unless otherwise specifically provided by law. As used herein, "minimal cost" shall mean the actual cost of reproducing the public record or public information.

jbruechert:
I think the closest international representation of that would be `CC0-1.0`. Feel free to add it as license if you think it is appropriate, otherwise it's also okay as it is.

leo60228:
I think it's probably best to leave the license field omitted when it's not explicitly specified by the agency?

jbruechert:
> I think it's probably best to leave the license field omitted when it's not explicitly specified by the agency?

In some jurisdictions agencies aren't allowed to specify a license because they have to publish as public domain by law, we use CC-0 in this case because it effectively results on the same terms, but is portable to jurisdictions that don't have the concept of public domain.

(And re-licensing public domain as CC-0 should always be allowed)

jbruechert:
Thanks :)

PR #2133, opened 2026-05-12
fix(france): skip Pays de la Loire feed exposing flights as buses (#1646)

### LINKED ISSUE #1646: Incorrect flights as buses with unrealistic times in France
Incorrect flights as buses with unrealistic times in France.

Example: https://api.transitous.org/?tripId=20251028_14%3A00_fr-arrets-horaires-et-circuits-des-lignes-de-transports-en-commun-en-pays-de-la-loire-gtfs-destineo-reseaux-aom-aleop-1_AEROPORT_NANTES%3AVehicleJourney%3AV72237_PMO_20250902102000%3AA

We should probably just ignore the feed.
needs editing `src/generate-france.py`.

mvanhorn:
## Summary

Marks the Pays de la Loire / Destineo / Réseaux AOM Aleop GTFS feed as `skip: true` so it no longer produces routes where flights from Aéroport Nantes Atlantique render as bus journeys.

## Why this matters

#1646 ships a concrete example trip (Nantes airport → some bus stop) where the journey planner returns a bus with unrealistic timing because the upstream feed labels flight legs as bus services. The feed name in the trip ID maps to `feeds/fr.json:5256-5273`:

```json
{
  "name": "arrets-horaires-et-circuits-des-lignes-de-transports-en-commun-en-pays-de-la-loire-gtfs-destineo-reseaux-aom-aleop-1",
  ...
  "drop-agency-names": ["Aéroport Nantes Atlantique"],
  "skip": false,
  ...
}
```

The existing `drop-agency-names` workaround targets that same airport agency, but the misclassified-bus output the reporter saw is still leaking through. The reporter's suggested fix in the issue is "we should probably just ignore the feed."

## Changes

`feeds/fr.json` — flip `"skip": false` to `"skip": true` and add a `"skip-reason"` describing the issue. Mirrors the field shape used by other already-skipped feeds in the same file (e.g. `transports-a-la-demande-en-pays-de-la-loire-gtfs-flex-aleop` lower in the file).

`src/update-france.py:97` updates existing entries via `id_map[resource["id"]].update(source)`. The script-built `source` dict does not include a `skip` or `skip-reason` key, so the next run of `update-france.py` will preserve these fields.

## Testing

- `python3 -c "import json; json.load(open('feeds/fr.json'))"` — file still parses.

Fixes #1646

### DIFF
--- a/feeds/fr.json
+++ b/feeds/fr.json
@@ -5267,7 +5267,8 @@
             ],
             "x-data-gov-fr-res-id": 79818,
             "managed-by-script": true,
-            "skip": false,
+            "skip": true,
+            "skip-reason": "Aéroport Nantes Atlantique flights are exposed as bus journeys with unrealistic times. See #1646.",
             "x-data-gov-fr-res-title": "GTFS Destineo Pays de la Loire",
             "x-data-gov-fr-dataset-id": "632b2a8a8bf5b436bed4f8df"
         },


jbruechert:
Can you give me a test route for testing this?
I'm not entirely convinced this is a good fix. The feed is rather big, and I think skipping it completely would make us lose coverage that we don't get from any other feed.

jbruechert:
The mentioned issue was just left open, it doesn't mean the issue prevailed after the previous fix

Lach-anonym:
Please do not skip this feed - there are some networks that do not have data in the open data portal yet
Also flights have been long removed from the feed
I'm working on skipping the agencies in aggregate feeds that are duplicated
A better input would be requesting the Semitan (Naolib network) for real time data...

mvanhorn:
Thanks for the context @Lach-anonym and @jbruechert. If flights are already removed from the AlEOP / Pays de la Loire feed and you (@Lach-anonym) are working on a targeted per-agency dedup pass on aggregate feeds, my full-feed skip is the wrong shape -- it would just drop coverage you both want to preserve.

Happy to close this PR in favor of your agency-level approach. Let me know if I should close now, or hold it until your dedup work lands so #1646 has a paper trail of the alternative considered.

PR #2134, opened 2026-05-12
Update feed GitHub URL for Czech GTFS-processor

0xaa55h:
Hi, I've got no clue if my feed is still used or if it still works (I assume yes, otherwise I'd get an issues on the repo). I've recently changed my name on GitHub so I'm just changing this to prevent name clashes.


mlundblad:
Thanks!

PR #2140, opened 2026-05-13
refine usage policy

traines-source:
I think we should apply some changes to the usage policy:
* make the main requirement, open source, clearer
* do not make it a hard requirement to contact us before. I think we should want to make the barrier for eligible use cases as low as possible. Having to contact us first may appear intimidating to some and may result in them using other closed APIs where supposedly they don't have to ask someone :clown_face: This is probably the most controversial proposition, open to discussion :)
* make it a hard requirement to link to the sources page. I think this is very important. To my understanding, every client also has to at least attribute OSM separately. We should probably check if our own/existing apps do that.

### DIFF
--- a/website/content/api.html
+++ b/website/content/api.html
@@ -32,11 +32,12 @@ there's also the option to <a href="/doc/#running-a-transitous-instance-locally"
 <p>If you want to use the API for your free and open-source application, please:<p>
 
 <ul>
-<li>Contact us <a href="https://matrix.to/#/%23transitous:matrix.spline.de">in our Matrix Channel</a> to make sure we can handle your
+<li>Make sure your source code is published under an appropriate open-source license (light prototype testing before that is of course fine).</li>
+<li>If you have any doubt about the load your requests will be causing (expensive requests, many users, etc.), please contact us <a href="https://matrix.to/#/%23transitous:matrix.spline.de">in our Matrix Channel</a> to make sure we can handle your
 expected resource usage and so we are aware of your usecases. This also allows us to contact you in case we want to coordinate breaking changes.</li>
 <li>
     <p>
-        Make sure your send HTTP <code>User-Agent</code> headers with every request, at least containing the following information:
+        Make sure you send HTTP <code>User-Agent</code> headers with every request, at least containing the following information:
         <ul>
             <li>Name of the application</li>
             <li>Version of the client implementation</li>
@@ -48,7 +49,8 @@ expected resource usage and so we are aware of your usecases. This also allows u
     However, please make sure to add contact information to your website in this case.
     </p>
 </li>
-<li>You may <a href="https://github.com/public-transport/transitous/blob/b2d184b51cb4cf20fb48d0b76420ba57f4d3fce0/website/content/_index.html#L30">add your application to the Transitous website</a></li>
+<li>Link to <a href="https://transitous.org/sources/">https://transitous.org/sources/</a> in a visible place in order to properly attribute the sources. Make sure to adhere to the conditions imposed by these sources. In particular, since Transitous uses OpenStreetMap data, you have to adhere to its <a href="https://www.openstreetmap.org/copyright">attribution guidelines</a>, too.</li>
+<li>You may <a href="https://github.com/public-transport/transitous/blob/b2d184b51cb4cf20fb48d0b76420ba57f4d3fce0/website/content/_index.html#L30">add your application to the Transitous website</a>.</li>
 </ul>
 
 <h3>Commercial Use</h3>
--- a/website/content/api.html
+++ b/website/content/api.html
@@ -34,7 +34,7 @@ there's also the option to <a href="/doc/#running-a-transitous-instance-locally"
 <ul>
 <li>Make sure your source code is published under an appropriate open-source license (light prototype testing before that is of course fine).</li>
 <li>If you have any doubt about the load your requests will be causing (expensive requests, many users, etc.), please contact us <a href="https://matrix.to/#/%23transitous:matrix.spline.de">in our Matrix Channel</a> to make sure we can handle your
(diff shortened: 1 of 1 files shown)


felixguendling:
100% agree with
- make main requirement (open source) clear
- make the other main requirement (attribution) clear

I think being in touch with people who run applications with (potential) public visibility is important for:
- community building
- letting them know if their application is using the API in an inefficient or problematic (e.g. DDoS) way
- letting them know that they can advertise their app on transitous.org

In general I would also allow pure development usage (even for commercial apps) if the number of requests stays small (i.e. under our radar).

traines-source:
I think we agree to allow pure development usage (which is why I have tried to include it with the phrase "light prototype testing before that is of course fine"), just that I consider an app that will be used by < 10 people pure development usage even if it is in production :D
Then again, development usage can actually be reasonably expensive while trying things out and refreshing for the umpteenth time – many rate limit violations in the past couple of weeks came from apparent development usage.

Also, I think the encouragement to join the channel is still there IMO (or can be made more explicit still), I personally just wouldn't phrase it as a requirement. Ways of contact are still covered by the user agent etc.

jbruechert:
Your changes look very reasonable to me, I would be okay with merging them as-is.
At least for reporting inefficient API usage back to developers, we already have the contact method from the User-Agent header, so making Matrix not a strict requirement seems fine.

traines-source:
In case we decide to drop the strict requirement, we could make the invitation more explicit, to cater for Felix' POV, e.g.:
"If you have any doubt about the load your requests will be causing (expensive requests, many users, etc.), please contact us <a href="https://matrix.to/#/%23transitous:matrix.spline.de">in our Matrix Channel</a> to make sure we can handle your expected resource usage. In general, we are always happy to hear about the projects using the API and their use cases. This also allows us to contact you in case we want to coordinate breaking changes."

Lach-anonym:
> I think being in touch with people who run applications with (potential) public visibility is important for:
> 
> * community building
> * letting them know if their application is using the API in an inefficient or problematic (e.g. DDoS) way
> * letting them know that they can advertise their app on transitous.org

As long as I agree on the app advertising part (very useful!), the requirement of contacting makes me a bit afraid...

From the feed license experience - some agencies require/suggest contacting them, for example when Licence Mobilité applies - in the licence itself it stands that it is recommended for better feedback on data provision/usage between the agency and consumer. It is the paper and sounds good. However it can be used as well for data access reglementation or monitoring, which raises a privacy and open access concerns.

Thankfully Transitous does not collect the user data unlike in described process, but it's just a small tip from me that we should provide services in a manner we expect from others. Also taking the server security into the account.

felixguendling:
> but it's just a small tip from me that we should provide services in a manner we expect from others

We do. For data: There is currently no limit or requirement (like contacting us) on data usage (including commercial use) for example for https://api.transitous.org/gtfs/

For the API server: CPU Resources are very limited. So it's our task to make sure that there are not a few inefficient apps that take up all the CPU resources and make the experience worse for users of other apps.

This problem could also be solved by having extremely low rate limits by default but if you "register" with us and explain your use case, we will whitelist the User-Agent for more requests.

I would like to prevent our API from becoming the next Overpass Turbo (which is more or less unusable right now):
- https://community.openstreetmap.org/t/entwicklung-neuer-app-schlagt-fehl-weil-fehler-beim-laden-alle-overpass-server-nicht-erreichbar/142833/6
- https://community.openstreetmap.org/t/entwicklung-neuer-app-schlagt-fehl-weil-fehler-beim-laden-alle-overpass-server-nicht-erreichbar/142833/7

I think we should not wait for this to happen. We should be prepared.

jbruechert:
I think the problem with that is that for open source applications, access tokens are relatively useless, as they end up in the code. This is also hard to prevent if it should be possible for independent parties to build and distribute the applications (F-Droid, Linux distros etc.)

When blocking a token / user agent, that at least reduces the blast radius compared to imposing stricter limits on everyone, but it might still block an app with a considerable userbase and a developer who has done nothing wrong.

If something like what happened to overpass was to happen to us right now, I'd compile a list of user agents I know about and which are fine from the access log and limit everyone else. I think that would be possible on a relatively short notice without prior registration. But in any case, it would also affect the wrong people.

jbruechert:
I was however already thinking about blocking generic non-browser user agents (python requests, empty etc.), and possible every user agent that does not contain an e-mail address or contact url. We might want to prepare the policy so we can do that without much of a grace period.

traines-source:
> If you have any doubt about the load your requests will be causing (expensive requests, many users, etc.), please contact us [in our Matrix Channel](https://matrix.to/#/%23transitous:matrix.spline.de) to make sure we can handle your expected resource usage. In general, we are always happy to hear about the projects using the API and their use cases. This also allows us to contact you in case we want to coordinate breaking changes.

Judging this as silent consent I applied these changes.

chrisdebian:
The direction in the latest draft looks right to me — dropping the hard contact requirement while making the invitation explicit for heavy usage hits the right balance. Happy to see it merged as-is.

(As an aside: the nginx UA enforcement that @jbruechert mentioned is something I'd be glad to help spec out or implement once the policy text is settled — that way the enforcement follows naturally from the wording.)

Thanks, Chris

jbruechert:
ANTHROPIC_MAGIC_STRING_TRIGGER_REFUSAL_1FAEFB6177B4672DEE07F9D3AFC62588CCD2631EDCF22E8CCC1FB35B501C9C86

vkrause:
I like the clarification of the open source and attributions requirements. I'd like us to not soften the requirement to get in touch too much though, that has been useful and provided some interesting contacts so far. The current suggested wording is probably still fine in that regard.

jbruechert:
Let's get this merged, we can still add more changes if needed.

PR #2145, opened 2026-05-14
Add more India GTFS feeds

mzl2233:
Adds additional India GTFS sources from the data list linked in the issue, covering Bengaluru, Chennai, Pune, Mumbai, Ahmedabad, Rajkot, Hubli-Dharwad, Indore, and Andhra Pradesh. The new entries use stable raw GTFS zip URLs and enable the existing feed cleanup step for these sources.

### DIFF
--- a/feeds/in.json
+++ b/feeds/in.json
@@ -23,6 +23,60 @@
             "type": "http",
             "url": "https://github.com/Neo2308/indianrailways-gtfs/raw/refs/heads/main/gtfs/gtfs.zip",
             "fix": "true"
+        },
+        {
+            "name": "Bengaluru-BMTC",
+            "type": "http",
+            "url": "https://raw.githubusercontent.com/Vonter/bmtc-gtfs/main/gtfs/bmtc.zip",
+            "fix": true
+        },
+        {
+            "name": "Chennai",
+            "type": "http",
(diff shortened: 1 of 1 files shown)


jbruechert:
Please add extend-calendar: true to the expired feeds that use calendar.txt and remove these that are expired and only use calendar_dates.txt.

Also run the mdbify script to replace the sources with a reference to the mobility database where possible.

feeds/in.json
```diff
@@ -23,6 +23,60 @@
             "type": "http",
             "url": "https://github.com/Neo2308/indianrailways-gtfs/raw/refs/heads/main/gtfs/gtfs.zip",
             "fix": "true"
+        },
+        {
+            "name": "Bengaluru-BMTC",
+            "type": "http",
+            "url": "https://raw.githubusercontent.com/Vonter/bmtc-gtfs/main/gtfs/bmtc.zip",
```
1Maxnet1, line 30: This feed is already available in [in-ka.json](https://github.com/public-transport/transitous/blob/main/feeds/in-ka.json) so we should not duplicate it here. I would propose to Either drop the regional files completely and merge their entries into the in.json or we keep this file for nation-wide feeds only and sort the regional feeds into the according files.

1Maxnet1:
@mzl2233 Do you need any further help in refactoring the PR? If so feel free to ask. Then we can increase the coverage for India :)

PR #2151, opened 2026-05-18
Add Robin's Transitous talk at the PRO BAHN meetup

vkrause:


felixguendling:
could also be promoted via Mastodon?

vkrause:
> could also be promoted via Mastodon?

right, we should have done that (too late now as the talk is over), I unfortunately only got the details earlier today.

traines-source:
Sorry, missed to forward it. But I hope the people subscribed to Transitous on Mastodon already know what Transitous is :)

PR #2156, opened 2026-05-19
fix: suppress false-positive SPDX scan in generate_colors.py

chrisdebian:
## Summary

`reuse lint` fails with one error:

```
'src/generate_colors.py' contains invalid SPDX License Expressions:
* CC0-1.0\n\n')
```

The cause: `generate_colors.py` writes SPDX licence headers into a generated Lua output file. The `reuse` tool scans all file content for `SPDX-License-Identifier:` strings, and picks up `CC0-1.0\n\n'` (including the trailing string literal characters) as a malformed expression.

The fix wraps those two lines with `REUSE-IgnoreStart` / `REUSE-IgnoreEnd` comments, which is the approach recommended in the REUSE spec for exactly this case. No licences, no generated output, and no behaviour is changed.

After this fix, `reuse lint` reports:

```
* Files with copyright information: 308 / 308
* Files with license information: 308 / 308
Congratulations! Your project is compliant with version 3.3 of the REUSE Specification :-)
```

The existing `.github/workflows/reuse.yml` CI job will pass once this is merged.

## Test plan

- [x] Ran `reuse lint` locally — passes cleanly after fix
- [ ] CI `REUSE Compliance Check` passes on this PR


jbruechert:
Please remove the claude committership once you are ready to stand for the changes with your own name.

chrisdebian:
Done — Co-Authored-By trailer removed and force-pushed.

jbruechert:
Thanks!

PR #2157, opened 2026-05-19
chore: pin Python dependencies and add Renovate config

chrisdebian:
## Summary

- Introduces `requirements.in` as the human-maintained direct dependency list (9 packages, unpinned)
- Converts `requirements.txt` to a `pip-compile`-generated lockfile with all transitive dependencies pinned to exact versions
- Adds `renovate.json` scoped to the `pip-compile` manager so Renovate can open automated PRs to keep pins current — no other dependency types are touched
- Updates `REUSE.toml` to cover `renovate.json` (JSON files cannot carry inline SPDX headers)

**No changes to the Containerfile or CI** — both already use `pip install -r requirements.txt`, which works with the pinned file.

## Why

The nine unpinned dependencies in the current `requirements.txt` mean CI and production can silently diverge if any upstream package releases a breaking change. Pinning eliminates that risk. Renovate keeps the pins current automatically.

## Regenerating the lockfile

```sh
pip install pip-tools
pip-compile requirements.in
```

## Test plan

- [x] `pip install -r requirements.txt` installs cleanly in a fresh venv
- [x] `reuse lint` reports 310/310 files compliant for the new files (pre-existing `generate_colors.py` failure is fixed separately in #2156)


jbruechert:
We test automatically whether things break because this file is used by our internal CI images.
We can still pin things if needed, for now this just adds more noise to the repo without anything being broken as far as I'm aware.

PR #2161, opened 2026-05-19
Add OSM Hack Weekend in Berlin to community events on the website

vkrause:
There's another one in Karlsruhe one month before, my criteria for adding them here is knowing that at least one person from the Transitous team is attending with high probability.


jbruechert:
Thanks :)

PR #2166, opened 2026-05-21
Make use of transitland entries if available for GBFS feeds

1Maxnet1:
Where available, it makes sense to use the upstream feeds to benefit from fixes and updates at transit land and reduce the redundancy


jbruechert:
Thanks!

PR #2171, opened 2026-05-23
sk: zsr

xhyrom:
replaces broken ZSSK (Železničná spoločnosť Slovenská) feed with ŽSR (Železnice Slovenskej Republiky) feed, including also other agencies such as RegioJet a.s.; Leo Express s.r.o.; Leo Express Slovensko s.r.o.; LTE Logistik a Transport Slovakia, s.r.o.; Trenčianska elektrická železnica, n.o.

### DIFF
--- a/feeds/sk.json
+++ b/feeds/sk.json
@@ -15,14 +15,13 @@
     ],
     "sources": [
         {
-            "name": "zssk",
-            "type": "transitland-atlas",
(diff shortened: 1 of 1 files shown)


jbruechert:
Thanks! Do you know if the license is still the same? In that case, it should be added to the Transitous metadata now because it no longer gets copied from the Transitland entry.

PR #2173, opened 2026-05-23
Fix typo in spdx-identifier field

jjasloot:
Replace `spdx-idenfifier` by `spdx-identifier`

Also checked, be.json is the only file with this typo


jbruechert:
Thanks!

PR #2174, opened 2026-05-23
Remove ignore_errors from motis config

felixguendling:
timetable.ignore_errors doesn't exist
it never worked properly and was removed long time ago


jbruechert:
Thanks!

PR #2186, opened 2026-05-27
Update GTFS source URL for Île-de-France Mobilités

Jouca:
This PR aims to propose a **new GTFS feed** (in addition to the GTFS-RT feed) of the Île-de-France Mobilités network to **include elements currently missing from the official GTFS that are essential** for the proposed GTFS-RT.

Here is what this new GTFS is expected to include in the future:
- Platforms (with real-time linked)
- Fares (still under development; may also be an extension for GTFS Fares V2)
- Pathways for elevators (including alerts with real-time data)

You can get the feed here: http://gtfsidfm.clarifygdps.com/gtfs

### DIFF
--- a/feeds/fr.json
+++ b/feeds/fr.json
@@ -7269,14 +7269,14 @@
         {
             "name": "reseau-urbain-et-interurbain-dile-de-france-mobilites",
             "type": "http",
-            "url": "https://www.data.gouv.fr/api/1/datasets/r/413988ed-d340-467b-8be2-7b999fcd207a",
+            "url": "http://gtfsidfm.clarifygdps.com/gtfs",
             "fix": true,
             "license": {
-                "url": "https://transport.data.gouv.fr/datasets/reseau-urbain-et-interurbain-dile-de-france-mobilites",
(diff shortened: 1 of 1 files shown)


feeds/fr.json
```diff
@@ -7269,14 +7269,14 @@
...
+            "url": "http://gtfsidfm.clarifygdps.com/gtfs",
             "fix": true,
             "license": {
-                "url": "https://transport.data.gouv.fr/datasets/reseau-urbain-et-interurbain-dile-de-france-mobilites",
+                "url": "https://github.com/Jouca/IDFM_GTFS-RT/blob/main/LICENSE",
                 "spdx-identifier": null
             },
             "x-data-gov-fr-res-id": 80921,
-            "managed-by-script": true,
+            "managed-by-script": false,
```
jbruechert: To work with our update script, the autogenerated entry should be kept but be skipped via skip: true. Then you can add a new one for your feed.
Otherwise the script would add a new entry again which we don't want here of course.

Jouca: @jbruechert thanks! updated

PR #2187, opened 2026-05-27
Add Tisseo subcontractors GTFSRT

tcr-n:
Hello,

I created this feed to provide real-time data for subcontracted lines on the Tisséo network. These lines are not included in the main real-time feed for technical reasons, as the subcontracted operators use a different system.
This feed complements the existing one and does not duplicate any data from the main feed.

NB : there are many lines that should be added in the next days. I have to adapt for each operators. It shouldn't create any problems.

### DIFF
--- a/feeds/fr.json
+++ b/feeds/fr.json
@@ -216,7 +216,7 @@
         {
             "name": "tisseo-reseau-transport-urbain-toulousain",
             "type": "url",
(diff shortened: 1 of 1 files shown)


jbruechert:
Thanks! We need protobuf tough, not json for GTFS-RT

tcr-n:
Oh, I'm swapping the links, the two formats are available 😄

PR #2193, opened 2026-05-29
Add drop-agency-names to France feeds

tcr-n:
Hello,

I dropped all of these agencies due to many (many is not enough in my opinion 😆) duplicates. There are still some networks that don't have data.
They do not update the networks that already have their data.


traines-source:
I think this was discussed already elsewhere, I wouldn't invest too much time into that because we will hopefully improve the automatic deduplication in MOTIS soon and the danger of accidentally removing too much when doing it manually (i.e. having missing trips) is much greater than the benefit IMO. And you probably have to coordinate with #2194 ?

tcr-n:
Actually, Atoumod makes feeds by collecting them from local authorities. The problem is that they don't update them correctly, you get many problems with realtime trips which are 6 hours late due to non-matching trips. It's better to download individuals feeds which are produced by authorities or by Atoumod for some networks. The global feed is never updated. You can see it with Cosibus. The data expired 1 week ago. The individual feed is up to date since 1 week ago

PR #2194, opened 2026-05-29
Skip duplicated agencies in Grand Est and Pays de la Loire

Lach-anonym:
I wanted to do one big PR for entire France, but I couldn't find time to do the rest, so I'll just create a PR for two regions for now


traines-source:
I think this was discussed already elsewhere, I wouldn't invest too much time into that because we will hopefully improve the automatic deduplication in MOTIS soon and the danger of accidentally removing too much when doing it manually (i.e. having missing trips) is much greater than the benefit IMO. And you probably have to coordinate with #2193 ?

Lach-anonym:
I remember that discussion, however in the meantime I thought it has stalled, as other agency skipping PRs were merged. I still wouldn't trust the feed aggregates, they can get invalid easily and I can't imagine when deduplication would remove proper data from the operator... Also let's take a situation when stop times are much different between the trips - I think that data from both feeds would be kept, which is also wrong...

traines-source:
If you're ready to do that extra work, I don't want to stop you :) And yes trips with very different times will never be merged automatically.

Jouca:
Hello @traines-source @Lach-anonym @tcr-n ! 👋

Reading through both threads, I think there's a bit of a misunderstanding about what @tcr-n and @Lach-anonym are actually trying to achieve with their PRs. I'd like to lay out some concrete arguments for @traines-source on **why aggregated feeds shouldn't be trusted for France**, and why Transitous should rely on feed-by-feed data instead.

- **Aggregated feeds are produced by regional authorities, not by the operating agencies themselves.** They're built by collecting and merging individual GTFS files from each network, which introduces a layer of processing that the original feeds don't go through. This means schedule errors that don't exist in the source files can appear in the aggregate, and conversely, updates pushed by an agency to their own feed often take time (sometimes weeks) to be reflected in the aggregate ; as @tcr-n pointed out with the Cosibus example, where the global feed was a week out of date while the individual feed was current.

- **Real-time data (GTFS-RT / SIRI) can't be reliably linked to aggregated feeds.** Each network publishes its own real-time feed tied to a specific static feed ID. When you merge multiple static feeds into one aggregate, you lose the 1-to-1 mapping between a given trip and its real-time source ; the aggregate only exposes a single feed ID, so there's no clean way to know which real-time feed corresponds to which trip inside the aggregate. In practice this leads to mismatched real-time updates, or trips showing stale/wrong real-time info because the link to the correct source feed was lost during aggregation.

Given these two points, I think the deduplication discussion (relying on MOTIS's upcoming automatic dedup) is solving a slightly different problem than the one @tcr-n and @Lach-anonym are raising. Automatic dedup at the routing-engine level can help avoid showing the same trip twice when both an aggregate *and* an individual feed are present, but it doesn't fix the underlying data-quality and real-time-linking issues with aggregated feeds themselves. Even with perfect dedup, you'd still be choosing between a fresher, RT-compatible individual feed and a potentially stale, RT-incompatible aggregate ; and as @Lach-anonym noted, in cases where stop times genuinely differ between the two feeds, automatic dedup would likely keep both versions anyway, which is its own failure mode.

So I'd argue dropping the aggregated agencies in favor of individual feeds (as both PRs are doing) is the right call regardless of dedup improvements in MOTIS ; it's addressing a data-source quality problem, not just a duplicate-trip display problem. That said, @traines-source's point about coordinating #2193 and #2194 still stands: since both PRs touch overlapping logic (skip flags / drop-agency-names in `fr.json`), it'd be worth syncing on a shared approach before merging either one, to avoid conflicts or inconsistent handling between regions.

PR #2203, opened 2026-06-03
Add drop_shapes property to flixbus entry

manfredbrandl:
This will trigger calculation of better shapes with osm-data.

### DIFF
--- a/feeds/eu.json
+++ b/feeds/eu.json
@@ -18,6 +18,7 @@
             "name": "flixbus",
             "type": "transitland-atlas",
             "transitland-atlas-id": "f-u-flixbus",
+            "drop_shapes": true,
             "fix": true,
             "script": "eu-flixbus.lua"
         },


felixguendling:
We'll have to use another binary (no mimalloc linked in) for this because the default links with mimalloc which is significantly faster but also allocates more RAM (and gives it back to the OS later it seems). So we can merge this after we have switched a binary without mimalloc.

jbruechert:
@felixguendling So the plan is to switch to the musl allocator, or do we need to change the CI job to dynamically link glibc?

felixguendling:
On the test instance, the import never had OOM crashes with clang-21 + glibc. So if we don't want to experiment much (e.g. with mimalloc settings), I would just go with this approach.

traines-source:
I think if we're lazy we could also just try merging this (before a day with few timetable changes). I think we might have made enough improvements that it could work also with mimalloc (last time complete recalculation of shapes peaked at 70% RAM, so we have 50GB left for FlixBus). In the worst case we'll have a broken import for one day.

PR #2205, opened 2026-06-05
Add Thessaloniki (OSETH) bus network

MartinLangbecker:
Adds the Thessaloniki bus network (OSETH — Οργανισμός Συγκοινωνιακού Έργου Θεσσαλονίκης), covering ~381 routes and ~3,700 stops in the Thessaloniki regional unit.

Source: [data.gov.gr](https://data.gov.gr/dataset/dedomena-astikon-sygkoinonion-p-e-thessalonikis)

Mirror: [thessaloniki-gtfs-mirror](https://github.com/MartinLangbecker/thessaloniki-gtfs-mirror) (weekly GitHub Actions workflow, publishes latest GTFS at a stable URL)

Note on license: The OSETH dataset on data.gov.gr does not declare an explicit SPDX license in its metadata. The portal's Terms of Use (https://data.gov.gr/pages/terms-of-use) permit free reuse (including commercial) with no specific license type assigned to this dataset. I've linked the terms page as the license URL.

### DIFF
--- a/feeds/gr.json
+++ b/feeds/gr.json
@@ -3,6 +3,10 @@
         {
             "name": "Lach Anonym",
             "github": "Lach-anonym"
+        },
+        {
+            "name": "Martin Langbecker",
+            "github": "MartinLangbecker"
         }
     ],
     "sources": [
@@ -22,6 +26,16 @@
             "url": "https://www.gitlab.com/Lach-anonym/athens-gtfs/-/jobs/artifacts/main/raw/gtfs-rail.zip?job=download-republish-gtfs",
             "license": {
                 "spdx-identifier": "CC-BY-SA-4.0"
+            },
+            "fix": true,
+            "extend-calendar": true
+        },
+        {
+            "name": "Thessaloniki-OSETH",
+            "type": "http",
+            "url": "https://github.com/MartinLangbecker/thessaloniki-gtfs-mirror/releases/latest/download/oseth_gtfs.zip",
+            "license": {
+                "url": "https://data.gov.gr/pages/terms-of-use"
             },
             "fix": true,
             "extend-calendar": true
--- a/feeds/gr.json
+++ b/feeds/gr.json
@@ -33,9 +33,10 @@
         {
             "name": "Thessaloniki-OSETH",
             "type": "http",
-            "url": "https://github.com/MartinLangbecker/thessaloniki-gtfs-mirror/releases/latest/download/oseth_gtfs.zip",
+            "url": "https://data.gov.gr/api/3/action/package_show?id=dedomena-astikon-sygkoinonion-p-e-thessalonikis",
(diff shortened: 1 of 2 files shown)


jbruechert:
Thanks! Now that you have already set up the mirror, we might as well use it, but we have support for custom code to find the latest URL here as well, see https://github.com/public-transport/transitous/blob/main/src/region_helpers.py.

MartinLangbecker:
I see, thanks for the link. I'm new to this and just tried to figure it out. If your approach is cleaner, I might as well use it. Also: What about the license? Does it work the way I specified it or should I change/remove it?

feeds/gr.json
```diff
@@ -25,6 +29,16 @@
             },
             "fix": true,
             "extend-calendar": true
+        },
+        {
+            "name": "Thessaloniki-OSETH",
+            "type": "http",
+            "url": "https://github.com/MartinLangbecker/thessaloniki-gtfs-mirror/releases/latest/download/oseth_gtfs.zip",
```
jbruechert:
```suggestion
            "url": "https://data.gov.gr/api/3/action/package_show?id=42c9a7da-c86c-48b1-914c-f340e8bef00d",
            "function": "data_gov_gr_latest_resource"
```

And then add something like this to region_helpers.py
```Python


def data_gov_gr_latest_resource(source: HttpSource) -> HttpSource:
    resources = requests.get(source.url).json()["result"]["resources"]

    gtfs_resources = [
        r for r in resources
        if r.get("last_modified")
        and r.get("format") == "ZIP"
        and not r.get("downloadall_datapackage_hash")
    ]

    source.url = max(
        gtfs_resources,
        key=lambda r: datetime.fromisoformat(r["last_modified"])
    )["url"]

    return source
```

jbruechert:
> What about the license? Does it work the way I specified it or should I change/remove it?

a url is completely fine in principle, but I think there's a hidden mention of CC-BY-4.0 in the [resource details](https://data.gov.gr/dataset/dedomena-astikon-sygkoinonion-p-e-thessalonikis/resource/091a41e1-8be2-413e-af5a-fbf788175ee1)

MartinLangbecker:
Thanks for the pointers. :) I made the changes, one less repo to maintain. 😄

jbruechert:
Looking at this again, unfortunately it looks like the latest feed is already expired, and the way it is structured does not allow extend-calender to work. Let's hope they update it at some point, but looks like this won't be reliable.

MartinLangbecker:
Yeah I noticed the same only afterwards: empty calendar.txt, everything in calendar_dates.txt so there's no weekly pattern to extend. Hopefully they'll upload a fresh one soon. Sorry about that. 🙁

jbruechert:
No worries, it's still good to at least track it, so it's easy to see what has been researched so far.

Issue #470, opened 2024-08-29
[Athens, Greece] Add GTFS data for Athens public transport (https://data.hcap.gr/)


saligari:
I've come across this site that seems to be up to date.

Here: https://catalog.hcapdata.gr/dataset/dromologia-osy are the buses & trolley buses information and here https://catalog.hcapdata.gr/dataset/dromologia-statheron-sygkinonion the metro/electric train as well as tram info.

I'm not familiar with GTFS or how this stuff works or I'd file a pull request. If someone is more knowledgeable or has the time, please give this a try.

Thanks!

saligari:
I sent them a message using their "ask for data" form as I didn't see any other contact info. Will reply again when/if they respond.

CommanderStorm:
@saligari did they respond to your request?

I looked into if cleaning up these is possible (for a student I am working with having greece data would be interesting).

For the first dataset it is, for the second it is not.

The reason is that the second dataset is barely exists..
![Image](https://github.com/user-attachments/assets/07a6e0fa-82ed-4cf0-9043-357b2a3291c4)

Also for said dataset `calendar.txt` does not exist or failed to upload on their end.

I don't think that just the first dataset would be really usefull (or would it?).
Here is the code to clean up both and remove the random junk (this is a 2-in-1 dataset with .rar, .zip and random characters thrown in for fun)

<details><summary>hacky bash code to extract their data (click to expand)</summary>
<p>

```bash
#!/bin/bash

set -euo pipefail

if [ "$#" -ne 2 ]; then
  echo "Usage: $0 <dataset_url> <output_file>"
  echo "Example: $0 https://catalog.hcapdata.gr/dataset/dromologia-osy.jsonld dromologia-osy.zip"
  echo "Example: $0 https://catalog.hcapdata.gr/dataset/dromologia-statheron-sygkinonion.jsonld dromologia-statheron-sygkinonion.zip"
  exit 1
fi
dataset_url=$1
output_file=$2
previous_dir=`pwd`

dir=`mktemp --directory`
echo "--------------------"
echo "Downloading to $dir"
echo "--------------------"
cd $dir
curl $dataset_url \
  | jq -r '.["@graph"][] | select(.["@type"] == "dcat:Distribution") | ."dcat:accessURL"."@id"'\
  | while read -r url; do wget -q --show-progress $url; done

echo "--------------------"
echo "Cleaning up packaging mistakes $dir"
echo "--------------------"

# cleanup of the random files
[...]

PR #2208, opened 2026-06-07
Switch Athens bus feed data source to data.gov.gr

MartinLangbecker:
Switches the Athens-Bus source from the GitLab mirror (expired, no scheduled pipeline) to data.gov.gr directly.

The OSY feed on data.gov.gr has a stable resource URL that doesn't change between uploads. Current feed is valid from 2026-05-25 to 2026-08-25. The feed has a proper calendar.txt with weekly patterns, so extend-calendar should work here.

Relates to #470.

### DIFF
--- a/feeds/gr.json
+++ b/feeds/gr.json
@@ -13,9 +13,9 @@
         {
             "name": "Athens-Bus",
             "type": "http",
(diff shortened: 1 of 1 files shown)


jbruechert:
Thanks!

PR #2210, opened 2026-06-07
Fix vahrplan icon

dabund24:
Turns out using `light-dark()` in an svg that is not inlined is a bad idea...

Sorry about the noise


jbruechert:
No problem :)

PR #2222, opened 2026-06-12
Update Sharing entries for Germany

1Maxnet1:
- added new Carsharing feeds from Baden-Württemberg
- Remove Nextbike Frankfurt as the feed is dead and Nextbike stopped their offer in Frankfurt at 01.04.2026
- skip deer feed, as the feed is not available, due to a system migration. There are plans to reintroduce it.
- For swu2go the original feed was unavailable when I tested the changes locally. Therefore, the url-override

### DIFF
--- a/feeds/de.json
+++ b/feeds/de.json
@@ -172,11 +172,6 @@
             "transitland-atlas-id": "f-lime~frankfurt~gbfs"
         },
         {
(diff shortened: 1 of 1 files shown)


jbruechert:
Thanks!

PR #2230, opened 2026-06-14
Temporary feed for KM-ZL

kasmar00:
Koleje Mazowieckie standalone feed was removed, so the previous temporary solution (while I'm waiting for access to their data to include their bus line in Polish Trains) won't work. So a new temporary solution (yep, I know...):
- include Koleje Mazowieckie from Polish Trains
- include a temporary feed with the ZL bus line, that is missing in the Polish Trains.

I'm once again sorry for replacing one temporary solution with another (although, this new one shouldn't need replacing), but I didn't expect it to take 4 months (and counting) to get the data from Koleje Mazowieckie...


kasmar00:
The failures are unrelated, there are a few outdated feeds - I'll address them separately today/tomorrow.

jbruechert:
Thanks!

PR #2232, opened 2026-06-15
nz: split off NZ feeds to regions

applecuckoo:
most feeds (with the exception of the eBus feed which spans both Nelson and Tasman regions) have an emphasis on one region, this PR splits all of them out for simplicity


jbruechert:
Thanks!

PR #2250, opened 2026-06-22
bg: Add Yambol and Sofia airport terminals shuttle feeds

BPilot253:
The Yambol feed is generated from https://transport.yambol.bg and the Sofia airport one from https://sofia-airport.eu/en/access/transportation.


jbruechert:
Thanks!

PR #2251, opened 2026-06-22
switch back to direct link for Victoria feeds, general script cleanup for NZ and RT data for Baybus

applecuckoo:
genuinely didn't think I'd be back to add more RT data for NZ, but the backend API used for the Baybus website is capable of sending out Protobuf


jbruechert:
Nice, thanks!

PR #2253, opened 2026-06-23
add feed for Southland

applecuckoo:
surprisingly this leaves only one major area in NZ (Gisborne) without its GTFS feed in Transitous

### DIFF
--- a/feeds/nz-stl.json
+++ b/feeds/nz-stl.json
@@ -1,0 +1,20 @@
+{
+    "maintainers": [
+        {
+            "name": "Thomas Dickson",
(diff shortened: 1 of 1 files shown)


jbruechert:
Unfortunately I cannot access the URL. Is it broken for you too, or a case of geoblocking?

applecuckoo:
hmm, it might be blocked from outside New Zealand. that feed was last updated 2025 though so it should be static enough to host as a static snapshot

jbruechert:
Let's go with this for now, but it will need to be updated towards the end of the year (current expiry date)

PR #2256, opened 2026-06-24
Add westbike (Nextbike Heinsberg)

dancojocaru2000:
The feed is on Transitland, and the same feed exists and works on citybik.es


dancojocaru2000:
Huh, the failed test seems to be related to DELFI?

jbruechert:
Seems to have been temporary, and is not related to your changes

PR #2257, opened 2026-06-24
ansible: nginx: Reply with link to usage policy when no user-agent is set

jbruechert:


ansible/roles/nginx/templates/transitous.conf.j2
```diff
@@ -62,6 +62,14 @@ server {
         allow all;
     }
 
+    if ($http_user_agent = "") {
+        return 403 "Empty user-agent header is not allowed. Please review the usage policy: https://transitous.org/api/";
+    }
```
traines-source: I'm not sure whether some browser users may be caught by this (thinking of people with noscript and notrack and whatnot that surely to avoid fingerprinting also disable the user agent?)

jbruechert: I have never seen this, sending no user agent is about the most uniquely identifiable thing you can do as a browser, because it is so uncommon. Probably the reasonable privacy thing to do it setting Chrome + Latest android, because that is the largest pool.

PS. I am one of these people.

jbruechert: Maybe we can make it conditional on `Accept-Language` header, that would be something browsers would generally send I imagine.

PR #2259, opened 2026-06-25
Updating deprecated GTFS-RT and license URLs for Tisséo ST (France)

eglinton-fr:
ST = Sous-Traitance = sub-contractors = Alcis, Transdev, Negoti, Verdié

### DIFF
--- a/feeds/fr.json
+++ b/feeds/fr.json
@@ -217,9 +217,9 @@
         {
             "name": "tisseo-reseau-transport-urbain-toulousain",
             "type": "url",
-            "url": "https://gtfs.willbrooks.fr/trip_updates.pb",
+            "url": "https://1.gtfs.download/tisseo/trip_updates.pb",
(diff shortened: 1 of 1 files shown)


jbruechert:
Where do these new urls come from?

eglinton-fr:
> Where do these new urls come from?

From the same source as the last URLs. This subdomain points to the exact same server and resources as the previous ones, which were created by tcr_n_ and introduced in [this PR](https://github.com/public-transport/transitous/pull/2187).

jbruechert:
Ok, I was afraid we are hitting the server behind a reverse proxy or similar here, as the 1. suggested.
@tcr-n please let me know if this isn't the url we should be using.

tcr-n:
Nop, actually it's the correct URL. We switched to a new DNS and organized the available files better.

Sorry for the incident with the previous URL

PR #2268, opened 2026-06-29
Fix location of Hergenrath station

dancojocaru2000:
Location in the SNCB/NMBS dataset is 1 km away from the real location. The distance is too significant not to fix.

The new locations are taken from the OSM nodes.

### DIFF
--- a/scripts/be-sncb.lua
+++ b/scripts/be-sncb.lua
@@ -35,3 +35,21 @@ function process_trip(trip)
         trip:set_display_name(trip:get_route():get_short_name() .. " " .. trip:get_short_name())
     end
 end
+
+local location_overrides = {
+	-- Hergenrath
+	["S8844644"] = { 50.7090259, 6.0432282 },
+	["8844644"] = { 50.7090259, 6.0432282 },
+	["8844644_1"] = { 50.7090182, 6.0432049 },
+	["8844644_2"] = { 50.7090338, 6.0432516 },
+}
+
+function process_location(location)
+	local override_pos = location_overrides[location:get_id()]
+	if override_pos ~= nil then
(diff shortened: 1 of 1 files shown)


jbruechert:
Sorry, this is really not the right place to fix this. Please report it to SNCB.

dancojocaru2000:
Why not both? When SNCB fixes their data, this fix can be removed as well. But I thought the point of scripting was to fix issues in the datasets in the first place.

jbruechert:
Scripts are currently not CI tested, can break the import process and hard to review without looking at the data. As soon as they reference ids from the dataset they are also prone to breaking at some random point.

So I want to keep the amount of script changes as little as possible, usually restricted to reshuffling existing data.

dancojocaru2000:
Generally speaking, I agree. I think it would be unsustainable to use a script to, for example, modify the location of all stations to match OSM.

I think in this particular case it is justified as a stop-gap measure until SNCB/NMBS fixed the dataset in order to not direct people 1 km (straight line) / 1.8km (walking) wrongly. I've had a look at all other border stations and this is the only one that's horribly misaligned.

I would also argue that the code is written in such a way that, if IDs change for example, it just does nothing (as opposed to crashing the import). I could change it to use the name though if that's a big concern.

---

Ultimately it's indeed your call. I just feel like this is a more important correction than adding line colours (not that I have anything against those either).

dancojocaru2000:
I've sent an email to Belgian Mobility Company (since they're who [SNCB/NMBS point to here](https://www.belgiantrain.be/fr/3rd-party-services/mobility-service-providers/public-data)) and I'll see how it unfolds from there.

PR #2270, opened 2026-07-01
Move livnobus from hr to ba

vekejsn:
Livnobus operates in Bosnia, so it's appropriate to move it to ba.json.


jbruechert:
I think I got confused by the hr  language code appearing in the gtfs :)

vekejsn:
Yep, ˙bs, hr, sr` are the legal languages in Bosnia :D

PR #2271, opened 2026-07-01
GB: remove fares for the time being

traines-source:
The fares are humongous and are causing extremely slow queries right now in GB. I think disabling `keep-additional-fields` should remove the fare files without causing any other harm. Since the GB feed seems to be updated daily, this should be applied directly with the next import.

Not tested.

### DIFF
--- a/feeds/gb.json
+++ b/feeds/gb.json
@@ -21,7 +21,8 @@
             "drop-agency-names": [
                 "FlixBus"
             ],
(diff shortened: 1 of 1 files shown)


jbruechert:
Looks like it should work, let's try

felixguendling:
Should also be fixed with the next MOTIS release but I would keep those fares removed.

PR #2282, opened 2026-07-07
Add new transit sources to nz-auk.json

NeilSelkirk:
Adding Mahu City Express's GTFS file

### DIFF
--- a/feeds/nz-auk.json
+++ b/feeds/nz-auk.json
@@ -7,6 +7,10 @@
         {
             "name": "applecuckoo",
             "github": "applecuckoo"
+        },
+        {
+            "name": "AddTransit",
+            "github": "AddTransit"
         }
     ],
     "sources": [
@@ -37,6 +41,16 @@
             "name": "Flamingo-Auckland",
             "type": "transitland-atlas",
             "transitland-atlas-id": "f-flamingo~auckland~gbfs"
+        },
+        {
+            "name": "MahuCityExpress",
+            "type": "http",
+            "url": "https://addtransit.com/gtfsfile/97011/MahuCityExpress.zip",
+            "license": {
+                "spdx-identifier": "CC-BY-4.0",
+                "publisher": "Mahu City Express",
+                "publisher-url": "https://www.mahucityexpress.co.nz/"
+            }
         }
     ]
 }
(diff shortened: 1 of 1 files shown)


NeilSelkirk:
Mahu City Express is a private bus operator in the Auckland Region (Mahurangi / Snells Beach); GTFS hosted via AddTransit, licensed CC-BY-4.0.

feeds/nz-auk.json
```diff
@@ -7,6 +7,10 @@
         {
             "name": "applecuckoo",
             "github": "applecuckoo"
+        },
+        {
+            "name": "AddTransit",
+            "github": "AddTransit"
```
1Maxnet1: The maintainers are meant to be people maintaining this config file, not the feed itself. So either add yourself here, or leave the current maintainers as they are :)

feeds/nz-auk.json
```diff
@@ -37,6 +41,16 @@
             "name": "Flamingo-Auckland",
             "type": "transitland-atlas",
             "transitland-atlas-id": "f-flamingo~auckland~gbfs"
+        },
+        {
+            "name": "MahuCityExpress",
```
1Maxnet1, line 42: It seems to me, that the contained route, is already included in the current data. Did you check that it is not already present on transitous.org?

Maybe @applecuckoo can verify this.

feeds/nz-auk.json
NeilSelkirk: Ok.  Thanks.  I misunderstood and have removed.

feeds/nz-auk.json
NeilSelkirk: Mahu City Express is appearing in Transitous.  However, the data is not being sourced directly from their official GTFS feed.  The official feed is the one I have just added.

NeilSelkirk: Hi, I've investigated the Auckland Transport feed further and although the data is not exactly the same (e.g. Meta data points to Auckland Transport rather than Mahu City Express), at a journey planning level the content is accurate.  One question for you. If Mahu publishes realtime data that is not distributed by Auckland Transport, what would be the preferred solution?

applecuckoo: @NeilSelkirk hey - wasn't aware of this feed before! I would say that if Mahu does offer RT data, then I would rather cut them out of the AT feed and exclusively rely on their own feed. This is basically what I've done for Te Huia in the feed for Waikato Regional Council, since I'm using AT's RT data.

Issue #1060, opened 2025-04-12
Handle license for GBFS feeds


Altonss:
License info can be available in `system_information.json` of GBFS feeds. It would be great to handle this, so a link to the license can be displayed at https://transitous.org/sources/.

emmambd:
A note that we share the GBFS license URL in the Mobility Database API if there's interest in using it: https://mobilitydata.github.io/mobility-feed-api/SwaggerUI/index.html#/feeds/getGbfsFeed

More about this here: https://github.com/public-transport/transitous/issues/903#issuecomment-2936961466

1Maxnet1:
> A note that we share the GBFS license URL in the Mobility Database API if there's interest in using it: https://mobilitydata.github.io/mobility-feed-api/SwaggerUI/index.html#/feeds/getGbfsFeed

That would be great to use. I just had a look and currently we download the CSV file from https://files.mobilitydatabase.org/feeds_v2.csv, which does not contain the gbfs feeds, while the systems.csv does not contain the license information.

PR #2284, opened 2026-07-08
Initial GBFS attribution

### LINKED ISSUE #2274: GBFS feed attribution missing
GBFS feeds (and other feed types?) are not attributed here https://transitous.org/sources/

1Maxnet1:
This provides initial listings for the GBFS feeds in the attribution page. What we currently lack, is the licensing information, as  this is not provided by transitland-atlas.

As discussed in #1060 we could retrieve the information from the feed itself, but I'm not sure whether it makes sense to do these requests in the generate-attribution.py script, as we would redo this every time the attribution is updated. Ideally the license would be provided in the mobilitydatabase or transitland-atlas, but afaik, this is currently not the case. Therefore we could add in on our own to the feeds, but this might be double effort, as Mobilitydatabase already have the license available, but not in the CSV export we are currently using.

Also should we merge the current state to improve the situation or first fix the licensing issue (which lead to over a year of missing attribution at all)?

Any comments on how to proceed appreciated :)

Would resolve #2274

### DIFF
--- a/src/generate-attribution.py
+++ b/src/generate-attribution.py
@@ -169,6 +169,42 @@ def add_rt_attribution(attribution: dict, source: UrlSource) -> None:
 
     attribution["rt"].append(rt_attribution)
 
+def gbfs_source_attribution(source: UrlSource, source_id: str, region_data: dict):
+    attribution = region_data
+
+    if source.license.spdx_identifier:
+        attribution["spdx_license_identifier"] = \
+            source.license.spdx_identifier
+    if source.license.url:
+        attribution["license_url"] = source.license.url
+    if source.license.attribution_text:
+        attribution["attribution_text"] = source.license.attribution_text
+
+    attribution["operators"] = []
+    attribution["source"] = source.url
+
+    human_name: str = (
+        source_id.split("_")[1].replace("-", " ")
+    )
+    human_name = " ".join(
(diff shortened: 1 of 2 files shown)


src/generate-attribution.py
```diff
@@ -234,12 +270,19 @@ def attribute_sort_key(attr):
...
+                    case UrlSource():
+                        if source.spec == "gtfs-rt":
+                            if source_id not in attributions:
+                                print(f"Warning: Stray gtfs-rt source for {source_id} without a (locally available) static timetable")
+                                continue
+                            else:
+                                add_rt_attribution(attributions[source_id], source)
+                        elif source.spec == "gbfs":
+                            gbfs_attribution = gbfs_source_attribution(source, source_id, region_data.copy())
+                            if not http_attribution:
```
jbruechert: Should probably be
```suggestion
                            if not gbfs_attribution:
```

traines-source:
I think having any attribution (even if without license) is better than none, so I'd definitely be in favor of merging this.

traines-source (approved):

traines-source:
Somehow this doesn't appear to have worked, or am I missing sth?

1Maxnet1:
> Somehow this doesn't appear to have worked, or am I missing sth?

It kinda did. The file at https://api.transitous.org/gtfs/license.json does already include the GBFS feeds for attribution. It may be that the CI used the old license.json to build the website and then updating the license file. So the next website build should include the attributions.

PR #2286, opened 2026-07-09
Lots of new Turkish cities

yasinkaraaslan:
Credit goes to @Egezenn for scraping them from a local website.

### DIFF
--- a/feeds/tr.json
+++ b/feeds/tr.json
@@ -53,7 +53,7 @@
         {
             "name": "konya",
             "type": "http",
-            "url": "https://acikveri.konya.bel.tr/tr/dataset/c2e034e6-e015-49c6-8eec-6e2f7de9c105/resource/5742b7bd-3cbf-4784-bee1-f949fc0ca118/download/gtfs_07_2024.zip"
+            "url": "https://acikveri.konya.bel.tr/dataset/c2e034e6-e015-49c6-8eec-6e2f7de9c105/resource/ec944ecd-1c1f-4687-a7f6-fcf2dc5bb5db/download/gtfs_11_2025.zip"
         },
         {
             "name": "kocaeli",
@@ -78,7 +78,6 @@
             "name": "gaziantep",
             "type": "http",
             "url": "https://acikveri.gaziantep.bel.tr/dataset/12d26c98-edc5-4be3-8988-c753764b84f4/resource/b3254aa0-359e-4b7c-87a3-d5d56ad27353/download/attached_gtfs.zip",
-            "fix": true,
             "http-options": {
                 "ignore-tls-errors": true
             }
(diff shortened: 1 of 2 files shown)


yasinkaraaslan:
Oh some of them seem to be missing some files

src/transitland.py
```diff
@@ -20,7 +20,7 @@ def __init__(self):
     def load(path: Path):
         atlas = Atlas()
         for f in Path(path / "feeds").iterdir():
-            for feed in json.load(open(f, "r"))["feeds"]:
+            for feed in json.load(open(f, "r", encoding="utf-8"))["feeds"]:
```
mlundblad, line 23: Why is this change done?

I don't think it's needed for any of the added feeds, as they are not using Transitland?

yasinkaraaslan: This makes the script runnable in Windows because Windows defaults to cp1252 so it crashes when it hits a byte it doesn't understand.

feeds/tr.json
```diff
@@ -94,6 +94,144 @@
...
+            "name": "tokat",
+            "type": "http",
+            "url": "https://github.com/Egezenn/kk-gtfs/raw/main/data/tokat.zip",
+            "fix": true
+        },
+        {
+            "name": "yozgat",
+            "type": "http",
+            "url": "https://github.com/Egezenn/kk-gtfs/raw/main/data/yozgat.zip",
+            "fix": true
```
1Maxnet1: his feed seems completely empty currently (it only includes a agency.txt). Would you create an issue in the upstream repository? Then we can add a skip : true here and add the Issue link as skip-reason, until the feed is fixed.

feeds/tr.json
```diff
@@ -94,6 +94,144 @@
             "type": "url",
             "spec": "gtfs-rt",
             "url": "https://rt.flix.baguette.pirnet.si/rt.pb"
+        },
+        {
+            "name": "nigde",
+            "type": "http",
+            "url": "https://github.com/Egezenn/kk-gtfs/raw/main/data/nigde.zip",
+            "fix": true
```
1Maxnet1: For my local test, this feed worked without fixing it automatically. Any particular reason to include the flag here?
```suggestion
            "url": "https://github.com/Egezenn/kk-gtfs/raw/main/data/nigde.zip"
```

1Maxnet1 (commented):
Thanks for the contribution :) For the feeds that are currently blocking the pipeline, could you check whether they have issues and report them upstream? Then we can add a skip-reason with the linked issue until it is resolved. And for the ones where you added fix : true: Do they really all need this? If the feed is fine, we can omit the fix flag.

Egezenn:
Hi, some of the feeds are blank from the provider itself. I didn't hardcode them into the script to not push them as the CI script traverses their API automatically.

[Relevant fix](https://github.com/Egezenn/kk-gtfs/blob/e4e70533a78fd738bf6d24b5adcfc7a982f7f72e/scripts/generate_metadata.py#L296):
```js
BLACKLIST = {"akcakoca", "altinova", "yozgat"}
```

Others should work fine, however there is some garbage data and [unmapped shapes](https://github.com/Egezenn/kk-gtfs/blob/e4e70533a78fd738bf6d24b5adcfc7a982f7f72e/map/app.js#L651).

feeds/tr.json
```diff
@@ -94,6 +93,127 @@
...
+            "name": "nigde",
+            "type": "http",
+            "url": "https://github.com/Egezenn/kk-gtfs/raw/main/data/nigde.zip"
+        },
+        {
+            "name": "adana",
+            "type": "http",
+            "url": "https://github.com/Egezenn/kk-gtfs/raw/main/data/adana.zip",
+            "fix": true,
+            "s": true
```
jbruechert: this property doesn't exist

yasinkaraaslan: Oh so sorry, it was just for me keeping track of the fix-needing feeds or whatever (since json doesn't support comments). Will remove it now

jbruechert: ah alright

jbruechert:
Thanks!

PR #2292, opened 2026-07-15
Leech off Catenary Maps GTFS realtime data streams

kylerchin:


feeds/us-il.json
```diff
@@ -58,6 +74,12 @@
             "name": "Metra",
             "type": "transitland-atlas",
             "transitland-atlas-id": "f-dp3-metra"
+        },
+        {
+            "name": "Metra",
+            "type": "url",
+            "spec": "gtfs-rt",
+            "url": "https://birch.catenarymaps.org/gtfs_rt?feed_id=f-metra~rt&feed_type=trip"
```
1Maxnet1, line 82: Are these feeds a liberated version of the transit.land realtime feed? If so what do you think about using the transit.land type and use url-override then? This would allow to use other attributes from upstream e.g. for attribution.

feeds/us-ca.json
```diff
@@ -74,10 +74,9 @@
         },
         {
             "name": "Santa-Cruz-Metro",
-            "type": "transitland-atlas",
+            "type": "url",
```
1Maxnet1: Either also omit the transitland-atlas-id or keep the type to transitland-atlas and use url-override instead. I'd prefer the latter, see the other comment.

1Maxnet1 (commented):
Thanks for adding these feeds! I have two suggestions, other then that it looks good :)

feeds/us-il.json
kylerchin: These feeds are independently downloaded. The only thing is I use the same ID schema. You could attribute Catenary Maps I suppose, though I don't require it.

jbruechert:
Thank you very much!

PR #2294, opened 2026-07-15
bg: Use official BDZ feeds

BPilot253:
Adds the official BDZ static and realtime feeds from the National Access Point. The static feed is processed as the official one is filled with various mistakes and missing things, and I also decided to add translations as a bonus.

### DIFF
--- a/feeds/bg.json
+++ b/feeds/bg.json
@@ -81,13 +81,16 @@
         {
             "name": "bdz",
             "type": "http",
-            "url": "https://jbb.ghsq.de/gtfs/bg-bdz.gtfs.zip"
+            "url": "https://gtfs.livetransport.eu/gtfs/bdz.zip",
+            "license": {
+                "spdx-identifier": "CC0-1.0"
+            }
         },
         {
             "name": "bdz",
             "type": "url",
             "spec": "gtfs-rt",
-            "url": "https://rt.gtfs.baguette.pirnet.si/gtfs-rt/bdz/trip_updates.pb"
+            "url": "https://sipbg.gov.bg/bgnap/portal/api/catalog/subsets/cc43fe99-141e-4eb8-bef1-cae144870a0f/realtime-download"
         },
         {
             "name": "pleven",


BPilot253:
The error has been fixed.

jbruechert:
Very cool!
I think we might want to add a script to display both the route name and trip short name, but that can be done later.

jbruechert:
Is it possible this feed is cut off at the border to Romania? At least the data from Romania connects, but this is a bit unfortunate

BPilot253:
> Is it possible this feed is cut off at the border to Romania? At least the data from Romania connects, but this is a bit unfortunate

Yep, that’s the only data BDZ provides (even on their website, apart from the PDF timetables for international trains). The feed you generated had a few missing trains (2626/2627 is one example I noticed) but if the international ones are consistently accurate and up to date then I assume I could fetch your GTFS and merge just the international trains so the full routes are available.

jbruechert:
If that is easy to do for you, that sounds like a good solution, but I guess it should also serve as another push for me to make motis able to merge these. We have the same problem in a few other places.

BPilot253:
I will have to see how exactly to do it. I was actually thinking of doing this but I realised that the realtime updates will break once the train leaves the Bulgarian coverage and vice versa. Also, while we are talking about this, I noticed that the realtime isn’t working and the stops aren’t visible when map is zoomed out (please excuse my lack of knowledge if this is not an issue).

jbruechert:
The realtime should be fine, the config just didn't update because of an unrelated issue, so it is still fetching the old URL. The stops are probably not displayed because the feed doesn't use extended route types. You could add these based on the train class. (The map needs a way to figure out which are the long distance stops for smaller zoom levels)

jbruechert:
The RT matching stats look a bit weird.

```
Jul 18 17:50:47 vm-motis motis[88860]: 2026-07-18T15:50:47Z | [info][motis.rt                      ] GTFS-RT update stats for tag=bg-bdz, url=https://sipbg.gov.bg/bgnap/portal/api/catalog/subsets/cc43fe99-141e-4eb8-bef1-cae144870a0f/realtime-download: total_entities=65, total_entities_success=5 (7.69231%), trip_resolve_error=60 (92.3077%)
```

https://stats.transitous.org/d/3f6bfbe5-1b43-4806-a242-c9e781346d3b/transitous-timetables?orgId=1&from=now-24h&to=now&timezone=browser&var-feed=bg-bdz&var-agency=$__all&var-instance=transitous-motis

BPilot253:
Today I also noticed that the realtime data for some reason includes passing points, as well as the duplicate stops, which I fixed in the static feed. Today or tomorrow I will make an endpoint for the processed realtime feed.

jbruechert:
> Today I also noticed that the realtime data for some reason includes passing points, as well as the duplicate stops, which I fixed in the static feed. Today or tomorrow I will make an endpoint for the processed realtime feed.

Thank you!

PR #2312, opened 2026-07-19
bg: BDZ extended route types + new realtime feed

BPilot253:


jbruechert:
Thank you!

BPilot253:
@jbruechert Hi, do you have any idea why the extended route types do not get applied at all? Has the script not updated or is it something else? I double checked the code and table.find should work but I could convert it into a simple route_id-route_type map if that's the issue.

jbruechert:
It works fine without table.find. When in doubt, please test locally with motis.

```
function process_route(route)
    if route:get_id() == "BV" or route:get_id() == "MBV" then
        route:set_route_type(102)
    elseif route:get_id() == "PV" or route:get_id() == "KPV" then
        route:set_route_type(106)
    elseif route:get_id() == "A" then
        route:set_route_type(714)
    end
end
```

jbruechert:
I pushed my fix to main now.

BPilot253:
Alright, thank you.

PR #2313, opened 2026-07-20
au-tas: switch to constant file URL to avoid Cloudflare

applecuckoo:
looks like the Tasmanian state government installed Cloudflare sometime in June, fortunately the feed itself isn't under Cloudflare and the URL to the actual file hasn't actually changed


mlundblad:
Thanks!

PR #2317, opened 2026-07-20
Update zbiorkom live RT urls to include service alerts

kasmar00:
Zbiorkom.live started exposing serviceAlerts for several cities (and more will come), so updating the RT feeds to include those.
Also some minor changes:
- added Halinów and Koszalin
- fixed up the Mińsk and Ząbki feeds
- added RT for Warsaw ferries


jbruechert:
Thanks!

PR #2319, opened 2026-07-22
Exclude routes '096' and 'X96' from processing

manfredbrandl:
Add condition to exclude two flixbus-routes from processing. These routes are duplicates and only the daata from the official flixbus eu feed (eu-flixbus) should be used.


jbruechert:
Thanks!

PR #2321, opened 2026-07-22
Fix urls for zbiorkom.live RT and add new feeds

kasmar00:
Unfortunately zbiorkom.live introduced a _breaking_ change to their gtfs-rt processing, which makes their gtfs-rt feeds incompatible with original static feeds (they mangle trip_id's). So, their reprocessed gtfs feeds will be used and where available a link to the original feed is also included.

Additionally:
- adding new feeds for: Powiat-Wodzisławski, Wieliszew, Zamość, Żyrardów
- fix RT url for Kalisz and Warsaw Ferries
- add RT for Toruń, Koleje Małopolskie Buses and Wrocław.
- removed temporary KM-ZL feed
- added feed for narrow gauge railways in Poland
- added "fix" to tribus and "skip" to Kolej Pleszewska

### DIFF
--- a/feeds/pl.json
+++ b/feeds/pl.json
@@ -1080,7 +1080,8 @@
             "url": "https://files.girlc.at/gtfs/tribus.zip",
             "license": {
                 "spdx-identifier": "CC0-1.0"
-            }
+            },
+            "fix":true
         },
         {
             "name": "PKS-Opoczno",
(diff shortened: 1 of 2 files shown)


kasmar00:
The CI failure is unrelated (two expired feeds)

jbruechert:
Thanks!

PR #2323, opened 2026-07-24
Add correct GoVolta category to Lua script

dancojocaru2000:
GoVolta is a long distance train, so this sets the proper category, and also makes it show up as `GV 123` instead of `GoVolta`.


mlundblad:
Thanks!

PR #2325, opened 2026-07-25
Add Turkish State Railways

yasinkaraaslan:
I was able to generate a GTFS feed for Turkish State Railways using their API. I hope people find it useful.
fix is set to true because a newly opened station (Neşetiye) appears in trip data but hasn't been added to the API I get stations from, causing error.

https://github.com/yasinkaraaslan/tcdd-gtfs


mlundblad:
Thanks!

PR #2326, opened 2026-07-25
Fix Gdynia and Koszalin RT

kasmar00:
- Fix Gdynia RT - zbiorkom exposes cities in Tricity properly now, so we can use Gdynia RT and static
- override static to be compatible with RT for: Koszalin, Sulejówek, Powiat Miński, Koszalin
- Fix Zielona góra - feed is available with RT on zbiorkom.live and the original website was problematic (failing requests and python helper function)
- Fix urls of feeds from ODT


kasmar00:
Errors here are not related to the changes

kasmar00:
All feeds in pl.json pass, the issue is in frankfurt from de.json. I think we are ready to merge

PR #2329, opened 2026-07-26
Add static GTFS for TEP - Parma

occhicf:
I added the static feed for TEP in Parma, Italy, using the source available in transitland. The static feed works fine (which is all I added in this PR), however, the RT feed doesn't seem to work well with MOTIS, as I discussed in the Matrix channel, so I'd appreciate any help there. The feeds (both static and RT) are available at https://temporeale.cl.tep.pr.it/.

### DIFF
--- a/feeds/it.json
+++ b/feeds/it.json
@@ -50,6 +50,11 @@
             "transitland-atlas-id": "f-eave~wimob~it"
         },
         {
+            "name": "Emilia-Romagna-TEP-Parma",
+            "type": "transitland-atlas",
+            "transitland-atlas-id": "f-spz3-tepparma"
+        },
+        {
             "name": "Emilia-Romagna-TPER-Bologna",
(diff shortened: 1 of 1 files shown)


1Maxnet1 (approved):
Looks good to me, failing pipeline is due to other feeds in Italy failing: https://github.com/public-transport/transitous/issues/1737

Maybe you wanna create an issue about the experiences with the real-time feed that you posted in the Matrix Chat, so it does not get lost and others can help comment there too?

Also, you may wanna add yourself as maintainer of Italy, if you wanna help keeping the Italian feeds in a good state :)

occhicf:
Alright, I've created the issue for the real-time feed, hopefully it'll help track the problem down.

I'd be happy to help with the rest of the italian feeds, as soon as I'm able to I'll try checking the existing feeds as well as a new one I noticed is missing.

PR #2341, opened 2026-08-03
fr-sncf: switch to netex, add rudimentary reservation info

### LINKED ISSUE #2062: Switch to NeTEx for SNCF
The GTFS dataset for SNCF is a bit of a mess. It uses route type 2, so all trains appear as regional, it shows some weird strings instead of proper route names, etc.

The NeTEx dataset is much better: each `<ServiceJourney>` has a `<BrandingRef ref="TER"/>` (with the proper ref for the journey), and there is also the `<TransportSubmode>` that allows to distinguish high speed rail from regional trains.

The data in both datasets can be matched by GTFS: `route_id` <=> NeTEx: `ServiceJourney/LineRef/@ref` if needed.

I think Transitous should switch to NeTEx for SNCF.

traines-source:
closes #2062 

In MOTIS versions < 2.11 (i.e. current failover) this leads to a loss of RT coverage. Though IMO useful train names and route_type are more important than RT.

Also some rudimentary reservation info (also only works >= 2.11, but should suprisingly not break earlier versions). It's not correct everywhere, particularly at the borders (some feeds don't split trips or not at the location where reservation regulations change, overriding that is not yet supported). But I think it's a good start. Also maybe we shouldn't get too much into the reservation frenzy before building the Wikidata connector.

### DIFF
--- a/feeds/fr.json
+++ b/feeds/fr.json
@@ -7414,7 +7414,8 @@
             "managed-by-script": true,
             "skip": false,
             "x-data-gov-fr-res-title": "GTFS Trenitalia France",
-            "x-data-gov-fr-dataset-id": "64635525318cc75a9a8a771f"
+            "x-data-gov-fr-dataset-id": "64635525318cc75a9a8a771f",
+            "script": "fr-trenitalia.lua"
         },
         {
             "name": "horaires-des-trains-trenitalia-france",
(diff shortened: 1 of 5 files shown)


felixguendling (approved):

jbruechert:
Nice!

jbruechert (approved):

feeds/fr.json
```diff
@@ -12253,7 +12291,9 @@
             "x-data-gov-fr-res-id": 83582,
             "managed-by-script": true,
             "x-data-gov-fr-res-title": "https://eu.ftp.opendatasoft.com/sncf/plandata/Export_OpenData_SNCF_GTFS_NewTripId.zip",
-            "x-data-gov-fr-dataset-id": "6853c089b3ed5781f6adfdf7"
+            "x-data-gov-fr-dataset-id": "6853c089b3ed5781f6adfdf7",
+            "skip": true,
+            "skip-reason": "netex is better"
```
orhazal, line 12296: Hello !

I'm curious to know why NeTEx is considered better. Apart from the fact that the transport modes are more detailed, what are its main advantages for SNCF?

It still doesn't include the mandatory transfer times required for a multi-train reservation, for example.
Thanks :)

jbruechert: Most importantly, it allows to construct the actual name of the train. Previously we ended up with route names that aren't commonly used by SNCF for passenger information. GTFS also has the train number, but without any information about the train type.

orhazal: I see, thanks for the info

PR #2356, opened 2026-08-05
add some scooter providers for Hamburg

jopejoe1:


1Maxnet1 (approved):
Thanks, will merge as soon as the CI passes :)

felixguendling:
I would prefer to add the Mobidata BW aggregated feed which probably already contains those (?).

1Maxnet1:
> I would prefer to add the Mobidata BW aggregated feed which probably already contains those (?).

No Mobidata does not contain the feeds from Hamburg, as they focus on Baden-Württemberg. But they include feeds from some other countries which would result in duplicates with feeds e.g. from France, Switzerland and Austria. That's why we did not use the aggregated feed so far iirc.

feeds/de.json
```diff
@@ -212,6 +212,11 @@
             "type": "transitland-atlas",
             "transitland-atlas-id": "f-lime~frankfurt~gbfs"
         },
+        {
+            "name": "LimeHamburg",
+            "type": "transitland-atlas",
+            "transitland-atlas-id": "f-lime~Hamburg~gbfs"
```
1Maxnet1, line 218: It seems that this feed is not available since a month. Can you confirm that Lime still operates in Hamburg?

felixguendling:
> as they focus on Baden-Württemberg

As far as I understood @derhuerst and @hbruch they are collecting data all over Germany?

feeds/de.json
jopejoe1: Yep, they are still operating here.

1Maxnet1: Then there seems to be another issue, as the feed is not available. I reported it here: https://github.com/MobilityData/gbfs/issues/898 and will contact Lime about it.

derhuerst:
[MobiData-BW](https://mobidata-bw.de) mainly focuses on the state of Baden-Württemberg, where it intends to [aggregate or generate feeds of/for all sharing networks](https://github.com/mobidata-bw/ipl-orchestration/blob/88608a03ab34ad1ca128092e9b476325d8456a44/etc/lamassu/feedproviders.yml) we're aware of.

Some of the providers include vehicles/stations outside of `DE-BW` in their feeds, which we then pass through to the [aggregated feed](https://api.mobidata-bw.de/sharing/gbfs/v3/manifest.json). This is why we may have other cities' sharing networks, but we don't aim for 100% coverage there, their inclusion is more incidental.

---

For `DE-BW`, the MobiData-BW GBFS feed has some benefits over collecting the (many!) individual feeds behind it:
- Much less maintenance work for consumers because they don't need to keep the list of feeds up-to-date – especially smaller providers come and go more often.
- [Lots of fixes & tweaks to the upstream feeds](https://github.com/mobidata-bw/ipl-proxy/tree/1e991e4c23024a1b76f0229deb0b4a9f0339e805/app/converters), to make the GBFS more correct & usable.
- GBFS v3 support.
- Sometimes higher rate limits than the upstream feeds.

(@hbruch Correct me if I'm wrong!)

hbruch:
Nothing wrong. Additionally, MobiData BW provides GBFS for some sharing providers which only publish proprietary formats.
And all feeds are published under [DL-DE/BY-2.0](https://www.govdata.de/dl-de/by-2-0).

Regarding Germany wide coverage: I could imagine, that BW would offer such services for other states, which then provide equivalent services in different domains, but until now, I'm not aware of such an agreement.

feeds/de.json
```diff
@@ -212,6 +212,11 @@
             "type": "transitland-atlas",
             "transitland-atlas-id": "f-lime~frankfurt~gbfs"
         },
+        {
+            "name": "LimeHamburg",
```
jbruechert, line 216: If this is  currently permanently unavailable, maybe let's skip it until that changes

1Maxnet1:
@hbruch @derhuerst @felixguendling I separated the comments in to a separate issue,  to keep the discussion available. As MobidataBW is not an option for the feeds from Hamburg currently #2372

PR #2359, opened 2026-08-07
Add license information to Athens-Rail feed

Lach-anonym:
Also @jbruechert I think that based on your other contributions, the Hellenic Train feed could be licensed on ODbL-1.0, can you confirm please?


jbruechert:
Unfortunately it's not obvious which license would apply for Hellenic Train from the source data

PR #2361, opened 2026-08-07
Remove route filtering for specific short names

manfredbrandl:
Removed condition to filter routes with short names '096' and 'X96'. These routes have been remove from the source.


jbruechert:
Thanks!

PR #2367, opened 2026-08-09
Add KORAIL and SRT

dancojocaru2000:
The feed will usually be updated weekly.


dancojocaru2000:
Huh, we tested with a local MOTIS instance but didn't expect that I'd tolerate such stuff, will fix

dancojocaru2000:
Done! :D

jbruechert:
Thanks!

PR #2370, opened 2026-08-09
Use Korail feed with Realtime

MKuranowski:
This PR changes the Korail feed from @dancojocaru2000's one to mine. The original feed has wrong schedules, as if it was assuming that Saturday schedules apply over the entire week. For example, train 1013 runs only on Saturday, but in the current feed it's marked as all-weekdays, and several trains like 1173 or 1651 are missing.

My feed also comes with shapes and realtime; however it is missing SRT schedules, as there is no reliable source for those (that's a question for @dancojocaru2000 - where do you get the source data from?). On the other hand, [SR is being taken over by Korail](https://www.koreatimes.co.kr/business/companies/20260802/korea-approves-korail-sr-merger), but on the other other hand, SRT schedules are still not available on https://www.korail.com/ticket/reserve/train-timeTable.

The made up train schedules from the disaster of a KTDB feed should probably be purged, but I don't really have the time to do that.

### DIFF
--- a/feeds/kr.json
+++ b/feeds/kr.json
@@ -25,12 +25,21 @@
       {
         "name": "korail",
         "type": "http",
-        "url": "https://shells.ccc.ac/~aelin/gtfs/korail.gtfs.zip",
+        "url": "https://mkuran.pl/gtfs/korail.zip",
         "license": {
-          "spdx-identifier": "ODbL-1.0",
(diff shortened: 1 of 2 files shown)


jbruechert:
I'll briefly wait for @dancojocaru2000's comment before merging, because I'm not familiar with the data situation in Korea.

jbruechert:
Thanks!

PR #2373, opened 2026-08-10
use official Matrix favicon

HarHarLinks:
From https://github.com/matrix-org/matrix.org/blob/main/static/assets/favicon.svg

Unfortunately git submodule init and npm install was not enough to get whatever "spago" is so I couldn't easily test this.


jbruechert:
The process described in the README still works for me:
https://github.com/public-transport/transitous/blob/main/website/README.md

jbruechert:
image


Looks good to me, thanks!

PR #2374, opened 2026-08-10
Add website license data for Bolzano-STA

pa-ulex:
Commit changes the entry to http in order to add it to the license website/endpoint.

### DIFF
--- a/feeds/it.json
+++ b/feeds/it.json
@@ -15,8 +15,8 @@
         },
         {
             "name": "Bolzano-STA",
-            "url": "ftp://ftp.sta.bz.it/gtfs/google_transit_shp.zip",
-            "type": "ftp",
+            "url": "https://gtfs.api.opendatahub.com/v1/dataset/sta-time-tables/raw",
+            "type": "http",
             "spec": "gtfs",
             "script": "it-sta.lua",
             "license": {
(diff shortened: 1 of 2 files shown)


jbruechert:
> Commit adds the license data for the website

This file just contains test data for building the website, the real file is generated automatically. Please don't edit it.

pa-ulex:
Did not know, thanks for the info, will open new PR.

jbruechert:
You can recycle the old pull request, just push a new commit :)

jbruechert:
In theory there's nothing wrong with switching to http, but if the only reason is that our license code doesn't handle ftp, we should fix that.

pa-ulex:
Thanks for the help! When will this change appear at the  `/gtfs/license.json` API endpoint? I can't find my previous change at the mentioned endpoint – no pressure!

Edit: Saw your message, thanks for clarifying.

jbruechert:
> In theory there's nothing wrong with switching to http, but if the only reason is that our license code doesn't handle ftp, we should fix that.

Fixed in main

> Thanks for the help! When will this change appear at the /gtfs/license.json API endpoint? I can't find my previous change at the mentioned endpoint – no pressure!

After the data was imported again, happens every night at 2am CEST and should be finished about 1-2h later.

Since your previous change was also about this ftp source, the script just didn't handle it when it last ran. Should be fixed after the next run.

pa-ulex:
Seems to work now, thanks for fixing! I think we do not need this PR then.

PR #2375, opened 2026-08-10
Add Japanese trains

casept:
Add a (to my knowledge) complete feed of Japanese trains. Some of these might overlap with e.g. the Tokyo feed, which should be preferred as it has real-time information.

### DIFF
--- a/feeds/jp.json
+++ b/feeds/jp.json
@@ -7,6 +7,11 @@
     ],
     "sources": [
         {
+            "name": "japan-rail",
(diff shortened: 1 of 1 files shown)


jbruechert:
Thanks!

Lach-anonym:
Greetings, is the feed made from scraped data? I'm just asking because of the license issue for Träwelling

casept:
Yes, scraped.

PR #2381, opened 2026-08-11
Add Transurb S.A. from Galati, Romania

Steinhagen:
Transurb doesn't offer a GTFS archive, so I've started mapping each bus and trolley route in Galati in OSM based on their web programs. Since this process is quite time consumming, the GTFS (and OSM routes) are slowly been added.

At the moment, the existing GTFS archive contains some of the most used buses and the only trolley in Galati.


jbruechert:
Thanks!

PR #2384, opened 2026-08-12
hu: add new kecskemeti-kisvasut feed

gy-mate:


gy-mate:
Re https://github.com/public-transport/transitous/actions/runs/31571059141/job/94032960994?pr=2384#step:5:122:

I can successfully download the file from Hungary. @metrophil tested it from Switzerland which also worked.

jbruechert:
Probably just azure ip ranges that are blocked. It works from DFN, so will work in the actual production pipeline.

PR #2388, opened 2026-08-12
eclipse the cars

traines-source:
some more coach and reservation shenanigans


felixguendling:
Maybe the previous auto-classification in MOTIS based on bounding box diagonal wasn't that bad after all? Should we bring it back so less cases have to be fixed with Lua?

traines-source:
Yeah maybe. Although then in both directions (I think before it was only bus->coach?) And on the other hand, with the stops map API it's actually reasonably fast to find the culprits

PR #2396, opened 2026-08-13
Add Taiwan (TDX national GTFS)

jcanizalez:
Adds Taiwan via the Ministry of Transportation's TDX platform, their single national GTFS covers rail (TRA/THSR), metro (Taipei, Taoyuan, Kaohsiung), and bus operators island-wide (~400 agencies).

TDX gates downloads behind OAuth2 client-credentials, so this points at a daily authenticated mirror: https://github.com/jcanizalez/tdx-gtfs-mirror (fetches at 03:30 Taipei, strips the fare tables that make up two thirds of the raw feed, same approach as Transitland's TDX fetchers, and publishes a rolling release). The TDX terms permit redistribution with attribution, carried in the license block.

Happy to instead upstream OAuth token support into fetch.py if you'd prefer to pull from TDX directly,  the mirror exists only because of the auth gate.

Verified locally: 401 agencies, Asia/Taipei, rolling ~7-week service horizon, translations.txt with English stop names.

### DIFF
--- a/feeds/tw.json
+++ b/feeds/tw.json
@@ -1,0 +1,23 @@
+{
+    "maintainers": [
+        {
+            "name": "Javier Canizalez",
+            "github": "jcanizalez"
+        }
+    ],
+    "sources": [
+        {
+            "name": "taiwan-tdx",
+            "type": "http",
+            "url": "https://github.com/jcanizalez/tdx-gtfs-mirror/releases/download/latest/tw-gtfs.zip",
+            "license": {
+                "url": "https://tdx.transportdata.tw/term",
+                "publisher": "Ministry of Transportation and Communications, Taiwan (TDX)",
+                "publisher-url": "https://tdx.transportdata.tw",
+                "attribution-text": "Ministry of Transportation and Communications Transportation Data Circulation Service Platform"
+            },
+            "fix": true,
+            "fetch-interval-days": 1
+        }
+    ]
+}


kylerchin:
Thanks! Are you able to compile realtime data too?

jbruechert:
Thanks!
Why skip the fares table? Even if it is fares v1, we might decide to implement support for it eventually. 

You can use the "function" / region_helpers.py feature to handle oauth, but the mirror is fine if the code becomes to complicated.

jcanizalez:
@jbruechert Fair point, restored. The mirror now publishes the TDX feed untouched (~127 MB zip; the fares tables are ~450 MB uncompressed, which is why Transitland strips them, but if MOTIS may grow fares support, better you have them). The rolling `latest` release already carries the full feed as of the next daily run.

On `region_helpers`: happy to write the OAuth function if you'd prefer pulling from TDX directly — the only wrinkle is credentials: TDX requires a per-member Client ID/Secret (free registration, human-reviewed), so your fetch infra would need its own TDX account to hold them. The mirror exists to keep that credential burden on my side; whichever you prefer, I'll maintain it.

@kylerchin Not currently. TDX's realtime products are their own per-mode JSON APIs (arrival estimates, vehicle positions, per city/operator) behind the same OAuthmI haven't found a packaged standard GTFS-RT protobuf in their catalog. A JSON→GTFS-RT bridge over the same mirror pattern is doable future work, but it's a real project (per-city endpoints, trip matching against the static feed), so I'd keep it out of this PR's scope.

jcanizalez:
Correction with real numbers after publishing untouched once: the raw TDX zip is **435 MB**, and ~300 MB of that is the GTFS-Fares **v2** tables (fare_leg_rules / fare_transfer_rules / fare_products), which is what Transitland's TDX fetcher drops too. Fares **v1** (fare_attributes / fare_rules) is what you mentioned implementing, so the mirror now keeps v1 and strips only v2: **127 MB** daily. If you want the v2 tables as well, say the word, it's one set to delete from the drop list.

jbruechert:
> Correction with real numbers after publishing untouched once: the raw TDX zip is **435 MB**, and ~300 MB of that is the GTFS-Fares **v2** tables (fare_leg_rules / fare_transfer_rules / fare_products), which is what Transitland's TDX fetcher drops too. Fares **v1** (fare_attributes / fare_rules) is what you mentioned implementing, so the mirror now keeps v1 and strips only v2: **127 MB** daily. If you want the v2 tables as well, say the word, it's one set to delete from the drop list.

The v2 table is actually the more important for us. Our routing engine so far only supports v2. Unless it's a problem for you, I don't think we might the file being a bit bigger.

We have a setup to deal with secrets as well, but I'm fine with using the mirror.

jcanizalez:
@jbruechert Good to know, that inverts my assumption, thanks. The mirror now publishes the feed untouched: all fares tables kept, Fares v2 included. Today's run is live, 437 MB on the same `latest` URL, so the metadata in this PR needs no edits; the next fetch simply picks up the full feed. I also dropped the stripping logic from the mirror entirely (plus added a zip integrity check before each release), so there's no configuration to drift: what TDX publishes is what you get.

PR #2398, opened 2026-08-15
Add skip rules for Hauts-de-France and Bretagne trains feeds

tcr-n:
Hello,

This PR is dedicated to regions that doesn't know how to produce a feed without making duplicates with others feeds. There remains the issue of data updates by the regions.
The best solution is to skip the affected feeds.


jbruechert:
Generally fine, but how sure are you that everything in this feed is already covered by other feeds?
The feed situation in France is so complicated that so far I have refrained from doing this, because I am not able to check the completeness well myself.

tcr-n:
These feeds are originally published by SNCF. Regions usually agregates with lines in the perimeter for their websites or apps. The problem is that most of the time, they rename the lines to the point that it is no longer identifiable (and also without realtime while SNCF provides it).  There is also the problem of updates that may not be present in these aggregates.

In France, the publication of open data on public transport is regulated by the Transport Code. In theory, it is up to the AOMs (Autorité Organisatrice des Mobilités, for example : Île de France Mobilités for Paris and the region, Tisséo Collectivités for Toulouse, SYTRAL for Lyon and individual regions for regional routes and trains) to publish a GTFS. Most of the time, you will find "syndicats mixtes" such as Atoumod in Normandy that will produce data aggregates, never up to date to display a correct transport offer.
On the railway part, no region is able to produce its data, without having to use those of SNCF but as mentioned earlier, there is a loss of quality to be expected behind.

To summarise, in France, it is necessary to prioritise AOM data for local networks (cities, etc.), those of regions concerning regional (intercity) connections. For the regional rail sector, SNCF data are the most suitable. It will be necessary to remain vigilant in the future with competition, as is the case in the PACA region with Transdev's specific GTFS.

Hoping to have helped you better understand the organisation in France 😄

felixguendling:
This would require that there are no trains from third-party operators in those regions, correct? SNCF doesn't have those in their datasets.

I would hope that at one point we can do an overlap analysis (dataset x dataset) for all datasets in Transitous (and maybe even Mobility Database) based on duplicate check in MOTIS to make those decisions based on data saying "those feeds have a 98.3% overlap -> take the better one".

tcr-n:
Yeah, it's the problem actually. We don't even know what will happen in others regions. In PACA, Transdev publishes their GTFS. The next region to have another operator is Normandy with RATP Dev, but it's not until July 11, 2027.

The way public transport open data is organized has become increasingly messy as transport networks have evolved.
Sometimes, there is reason to envy neighboring countries...

jbruechert:
I'm a bit worried that we are only talking about trains here, what about all of the local buses?

tcr-n:
Like I said, about local transports, it is at the AOM's to publish the data. 

Actually, one solution that can help is to skip every agregates in France. It will make it possible to better determine whether a GTFS is better or not. Aggregates or republished data often have a refresh limited to 1-2 weeks. Intercommunality update the data more frequently, it goes from every day to 2-3 times a week. A high refresh rate allows you to have an up-to-date transport offer in the event of event reinforcements, strikes, etc.

felixguendling:
As long as we can be sure that the aggregated datasets don't contain any data other datasets don't have, I would agree that we don't want them because they usually decrease data quality are delayed.


> I'm a bit worried that we are only talking about trains here, what about all of the local buses?

The dataset names sound like they're actually really only trains.

----

In this case it looks like the data would come from SNCF (at least "TER" hints in this direction) but in general I would like to make those decisions based on overlap statistics based on MOTIS' duplicate checker.

Lach-anonym:
> Actually, one solution that can help is to skip every agregates in France.

I need to resume checking the aggregates if they have something that other existant feeds do not have...

> In this case it looks like the data would come from SNCF 

Well the problem is now not every train is SNCF and SNCF is very happy to miss on information about not their trains, so the future of TER is in the region-scoped feeds sadly

PR #2404, opened 2026-08-17
Removed companies from two feeds, they're redundant

luna-cant-code:
I've removed two bus companies from DELFI due to them providing their data to VBN.
The same is also true for Eurobahn, just with them being kicked out of the Dutch feed due to already existing data in DELFI and VBN (the latter isn't accurate on some lines though)


feeds/nl.json
```diff
@@ -15,6 +15,8 @@
             "type": "http",
             "url" : "https://gtfs.openov.nl/gtfs-rt/gtfs-openov-nl.zip",
             "fix": true,
+            "drop-agency-names": [
+                "Eurobahn",
```
jbruechert:
```suggestion
                "Eurobahn"
            ]
```

feeds/nl.json
```diff
@@ -15,6 +15,9 @@
             "type": "http",
             "url" : "https://gtfs.openov.nl/gtfs-rt/gtfs-openov-nl.zip",
             "fix": true,
+            "drop-agency-names": [
+                "Eurobahn"
+            ]
```
jbruechert:
```suggestion
            ],
```

PR #2406, opened 2026-08-17
Add data producer docs

jbruechert:
Pretend that we are a relevant consumer so maybe producers start caring

### DIFF
--- a/docs/docs/data-producers.md
+++ b/docs/docs/data-producers.md
@@ -64,6 +64,7 @@ We download sources with the user-agent header `transitous.org` from the
 following IP-addresses:
 
 - `90.187.116.41` (rt.triptix.de)
+- `130.83.165.222` (transitous.motis-project.de)
 - `130.133.110.30` / `2001:470:51c5:babe::30:1`
   (crunchy.spline.inf.fu-berlin.de)
-- `130.83.165.222` (transitous.motis-project.de)
+- `130.133.110.91` / `2001:470:51c5:babe::91:1` (vm-motis.spline.inf.fu-berlin.de)
--- a/docs/docs/data-producers.md
+++ b/docs/docs/data-producers.md
@@ -1,0 +1,73 @@
+---
+#template: home.html
+title: For Data Producers
+social:
+  cards_layout_options:
+    title: Free and Open Public Transport Routing
+---
+
+<!--
+SPDX-FileCopyrightText: None
+SPDX-License-Identifier: CC0-1.0
+-->
+
+If you publish open data for example for use in Transitous the following recommendations ensure we can use your data.
+
+## Supported Formats
+
+For schedules:
+
+- GTFS
+- GTFS-RT
+
+For real-time updates:
+
+- NeTEx (may require changes on our side if a new profile is used)
+- SIRI-ET (XML/JSON)
+
+## Supported GTFS Features
+
+The following GTFS features are supported in our routing engine (MOTIS) and used in the Transitous deployment.
+We recommend that you include data for them in your GTFS file if possible.
+
+- Fares V2
+- Shapes
+- Translations
+- Bikes Allowed
+- Wheelchair Accessibility
+- Flex
+- Google Transit Ticketing Extension
+- Google Transit Extended Route Types
+
+## Supported GTFS-RT features
+
+- Trip updates
+- Alerts
+- Vehicle Position (currently for our own delay calculation only)
+
+## Recommended Licenses
+
+The data should ideally be clearly licensed, even if you don't intend to restrict usage of the data in any way.
+We recommend that you use an established standard license, for example:
+
+- `CC0-1.0` If you don't want to restrict usage at all
+- `ODbL` Compatible with OpenStreetMap
+- `CC-BY-4.0` / `CC-BY-SA-4.0`
+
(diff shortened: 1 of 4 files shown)


docs/docs/data-producers.md
```diff
@@ -0,0 +1,73 @@
...
+-->
+
+If you publish open data for example for use in Transitous the following recommendations ensure we can use your data.
+
+## Supported Formats
+
+For schedules:
+
+- GTFS
+- GTFS-RT
```
derhuerst: Shouldn't this be under "realtime updates", swapped with NeTEx?

docs/docs/data-producers.md
```diff
@@ -0,0 +1,73 @@
...
+
+- NeTEx (may require changes on our side if a new profile is used)
+- SIRI-ET (XML/JSON)
+
+## Supported GTFS Features
+
+The following GTFS features are supported in our routing engine (MOTIS) and used in the Transitous deployment.
+We recommend that you include data for them in your GTFS file if possible.
+
+- Fares V2
```
derhuerst: Although most of these have been merged into the main reference for a while (and therefore their canonical specification is not the extension page), we *could* link to the extension page (e.g. [the Fares v2 one](https://gtfs.org/community/extensions/fares-v2/), same for Flex) that explains how the feature got developed, refined & accepted.

docs/docs/data-producers.md
```diff
@@ -0,0 +1,73 @@
...
+## Supported GTFS-RT features
+
+- Trip updates
+- Alerts
+- Vehicle Position (currently for our own delay calculation only)
+
+## Recommended Licenses
+
+The data should ideally be clearly licensed, even if you don't intend to restrict usage of the data in any way.
+We recommend that you use an established standard license, for example:
```
derhuerst: We could link to the [Open Definition's list of recommended conformant licenses](https://opendefinition.org/licenses/).

docs/docs/data-producers.md
```diff
@@ -0,0 +1,73 @@
...
+- **Support for `Last-Modified` / `If-Last-Modified` headers**. 
+  If this feature is not supported, files will be downloaded even if they didn't change, which wastes bandwidth for you and Transitous.
+- **Stable, permanent download URL.**
+  If the url changes, old data may be used in Transitous or your agency will temporarily not be covered until you or a contributor updates the URL on our side.
+- **Allow automated downloads.** CAPTCHAS make it impossible to retrieve up to date data from you, since we need to fetch hundreds of sources every day.
+
+## IP-Addresses
+
+We download sources with the user-agent header `transitous.org` from the
+following IP-addresses:
```
derhuerst, line 64: Is there IPv6? We could proactively encourage producers to allow-list Transitous' IPv6 addresses even if we don't use them right now.

jbruechert: As far as I know only for the Spline part of the infrastructure.

jbruechert:
The primary motivation for listing the IP-addresses is that I know some other deployments have started fetching realtime feeds with user-agent transitous.org, but we had no location to check whether the user agent is a lie or not.

felixguendling:
If we don't want people to copy our config.yml verbatim, we should not share it openly with `user-agent` set. Maybe the shared version should be stripped of things we expect people to fill in on their own.

jbruechert:
> If we don't want people to copy our config.yml verbatim, we should not share it openly with `user-agent` set. Maybe the shared version should be stripped of things we expect people to fill in on their own.

I know at least one deployment runs the fetch.py script themselves and still uses our user-agent. We also need to remove the hardcoded value in our repository apparently.
This is pretty annoying, because right now we also use api.transitous.org/gtfs/ to power our own instances. Apparently we need to add auth to that

docs/docs/data-producers.md
```diff
@@ -0,0 +1,70 @@
...
+
+- `90.187.116.41` (rt.triptix.de)
+- `130.83.165.222` (transitous.motis-project.de)
```
felixguendling, line 67:
```suggestion
- `130.83.165.222` (transitous.motis-project.de)
- `130.83.165.228` (test.motis-project.org)
```

but in the end we should find a way that all RT data goes via a single entry point without delay and we redistribute internally instead of using 3rd-party resources

docs/docs/data-producers.md
```diff
@@ -0,0 +1,70 @@
...
+- **Stable, permanent download URL.**
+  If the url changes, old data may be used in Transitous or your agency will temporarily not be covered until you or a contributor updates the URL on our side.
+- **Allow automated downloads.** CAPTCHAS make it impossible to retrieve up to date data from you, since we need to fetch hundreds of sources every day.
+
+## IP-Addresses
+
+We download sources with the user-agent header `transitous.org` from the
+following IP-addresses:
+
+- `90.187.116.41` (rt.triptix.de)
```
traines-source, line 66: rt.triptix.tech (.de is either not owned by us or at least not the actual server used), also with different IP address that can change at any time – may also be IPv6

docs/docs/data-producers.md
traines-source, line 67: this is also not the correct IP address that is actually used for fetching and I'm not sure whether we want to publish that explicitly either

PR #2408, opened 2026-08-18
Delete My Ride Rock Hill as it's no longer in service

apiote:
The home page (https://myriderockhill.com/) says only

> My Ride Rock Hill Is No Longer in Service as of May 22, 2026.


apiote:
fixes #2407

mlundblad:
It looks like the file us-sc.json is empty after the change (I think it was the only feed for South Carolina?)

Maybe our parsing doesn't handle empty files (not sure if that's valid JSON?). Maybe it would be best to delete the file? (if this is what causes generating the config.yml to fail).

apiote:
The file in the current state of `main` contains only the one feed for My Ride Rock Hill.
This morning I got a notification that the import failed, and I checked that transitland removed the feed, and transitland removed the feed, because the operator no longer provides its services.
This PR removes the feed from Transitous. And I'm pretty sure it removes the file `us-sc.json` completely as the changeset says "This file was deleted":

Screenshot 2026-08-18 at 10-16-06 Delete My Ride Rock Hill as it&#39;s no longer in service by apiote · Pull Request #2408 · public-transport_transitous

jbruechert:
Thanks!

PR #2420, opened 2026-08-18
PL more feeds from zbiorkom.live

kasmar00:
- Adds Kołobrzeg, Starachowice, Piotrków Trybunalski
- Adds RT for Świnoujście, Dąbrówka, Skarżysko Kamienna, Bielsko Biała, Tribus, Żywiec, Goleniów
- Moves RT for Legnica to api v6


kasmar00:
Only issue in import is Kombus, which returns an empty (0 byte) zip file 😅 New feeds cause no problems, so it should be clear to merge.

mlundblad:
Thanks!

PR #2421, opened 2026-08-18
STA: Remove GTFS-RT

pa-ulex:
I'm not 100% sure, but it seems that there is more data on the vehicle's position endpoint than on the trip updates endpoint.

I suggest testing this data beforehand. When I checked the data using [this website](https://app.transit-lens.com/), it showed more (accurate) data for the vehicle's position endpoint than for the trip updates endpoint. Currently, only a handful of services have real-time data, which is usually incorrect.

To verify this new data, you can use the [operator's website](https://www.suedtirolmobil.info/).

### DIFF
--- a/feeds/it.json
+++ b/feeds/it.json
@@ -26,7 +26,7 @@
         },
         {
             "name": "Bolzano-STA",
-            "url": "https://files.opendatahub.com/gtfs-rt/feeds/sta/trip-updates.pb",
+            "url": "https://files.opendatahub.com/gtfs-rt/feeds/sta/vehicle-positions.pb",
             "type": "url",
             "spec": "gtfs-rt",
             "license": {
--- a/feeds/it.json
+++ b/feeds/it.json
@@ -25,16 +25,6 @@
             }
         },
         {
-            "name": "Bolzano-STA",
-            "url": "https://files.opendatahub.com/gtfs-rt/feeds/sta/vehicle-positions.pb",
-            "type": "url",
-            "spec": "gtfs-rt",
-            "license": {
-                "spdx-identifier": "CC0-1.0",
-                "url": "https://github.com/noi-techpark/opendatahub-gtfs-api/blob/839d698938a224a7db4885f717ca05706fa6bae1/datasets.yml#L33"
-            }
-        },
-        {
             "name": "Bolzano-STA-flex",
             "url": "ftp://ftp.sta.bz.it/gtfs/google_transit_flex.zip",
             "type": "ftp",


felixguendling:
That would have to go through another service that creates better TripUpdates before we can use it in MOTIS atm. Transitous doesn't have MOTIS VehicleUpdate support enabled (it would require incremental updates and it's not really good anyway).

pa-ulex:
In that case, I'd rather remove the real-time updates. I saw that other feeds, such as those in Finland (HSL) and Bulgaria (Sofia), used this option, and I presumed that MOTIS/Transitous supported this format. It's better to have offline time data in the GTFS format than incorrect "real-time" data.

felixguendling:
We are planning to add support for vehicle updates in MOTIS, it's just not there yet.

However, it this will require a switch that vehicle updates should also be used to predict trip times. Otherwise the plan would be to only show GPS positions in MOTIS' map. But since this will create a lot of traffic (inbound and outbound), I'm not sure if that feature (map positions) is really worth it.

jbruechert:
We do have the `derive-trip-updates` option that uses a simplistic separate piece of software to calculate the delays, but unless there is something extremely wrong about the official trip updates, there is little reason to assume it will produce better results.

Generally we have just collected vehicle positions as well in case we later want to implement features based on it, right now we mostly don't really use them.

Lach-anonym:
@kasmar00 isn't it like that sometimes Zbiorkom interpolates the delay based on the vehicle position or am I missing something?

kasmar00:
@Lach-anonym 
> isn't it like that sometimes Zbiorkom interpolates the delay based on the vehicle position or am I missing something?

As far as I know Zbiorkom.live (v4 - so what is available on https://zbiorkom.live) relies solely on vehicle positions to calculate delays. The biggest issue they have is matching the vehicle to a trip, as ex. Warsaw exposes vehicle positions via a custom api managed by a different city entity than the one handling gtfs.
From v6 (still "unreleased" https://v6-api.zbiorkom-v5.pages.dev/) Zbiorkom has an ability to parse trip updates and uses it for train delays, and can also display (and expose via API) vehicle positions interpolated based on those trip updates

pa-ulex:
I will amend this PR to remove the RT feed.

pa-ulex:
PR should be ready.

jbruechert:
What is the point in removing it?

pa-ulex:
> What is the point in removing it?

The real-time data is inaccurate and always shows that the bus is on time despite it not being on time.
It seems that the feed is no longer visible on the website either. Have they blocked you guys?

traines-source:
It looks like it still works: https://stats.transitous.org/d/3f6bfbe5-1b43-4806-a242-c9e781346d3b/transitous-timetables?orgId=1&from=now-24h&to=now&timezone=browser&var-feed=it-Bolzano-STA&var-agency=$__all&var-instance=transitous-motis

PR #2425, opened 2026-08-19
added bike and scooter sharing (Wuppertal and Düsseldorf, Germany)

luna-cant-code:
I've added Lime scooters in Wuppertal 🇩🇪 and Düsselbike in Düsseldorf 🇩🇪

Please check if I did that right, it's my first time doing shared mobility, and the first time using NRW's open data platform for it

In theory, it should be all good.

### DIFF
--- a/feeds/de.json
+++ b/feeds/de.json
@@ -665,8 +665,11 @@
         },
         {
             "name": "DonkeyDuesseldorf",
-            "type": "url",
-            "url": "https://www.mobilitaetsdaten.nrw/api/systemadapter-gbfs-provider/conflated/v3.0/donkey-republic-duesselbike/manifest.json",
+            "type": "http",
+            "url": "https://www.mobilitaetsdaten.nrw/api/systemadapter-gbfs-provider/feed/v3.0/donkey-republic-duesselbike/source-duesselbike/gbfs.json",
+            "headers": {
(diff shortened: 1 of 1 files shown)


feeds/de.json
```diff
@@ -662,6 +662,19 @@
             "name": "westbike",
             "type": "transitland-atlas",
             "transitland-atlas-id": "f-westbike~heinsberg~gbfs"
+        },
+        {
+            "name": "DonkeyDuesseldorf",
+            "type": "http",
+            "url": "https://www.mobilitaetsdaten.nrw/api/systemadapter-gbfs-provider/conflated/v3.0/donkey-republic-duesselbike/manifest.json",
```
jbruechert: This seems to require authentication.

feeds/de.json
```diff
@@ -662,6 +662,19 @@
             "name": "westbike",
             "type": "transitland-atlas",
             "transitland-atlas-id": "f-westbike~heinsberg~gbfs"
+        },
+        {
+            "name": "DonkeyDuesseldorf",
+            "type": "http",
```
jbruechert:
```suggestion
            "type": "url",
            "spec": "gbfs",
```
as this does not need to be downloaded and needs to be passed to the routing engine as url.

luna-cant-code: Done, thanks!

feeds/de.json
luna-cant-code: I try to add that with help of the documentation, there is a public OAuth thing, I just need to grab the key.
When it's done, I'll undraft this PR :)

Just for me as a bookmark: https://www.mobilitaetsdaten.nrw/api/ckan/en/dataset/donkey-republic-duesselbike

luna-cant-code:
Just as a question: Is it possible to let the server grab the ACCESS_TOKEN every now and then in case it expires?

Screenshot 2026-08-19 at 22 22 57

luna-cant-code:
Just so it's clear what I've done:

I've requested a token by using the curl command in the screenshot above this comment, that token may or may not change, I'm not quite sure.
With all of that, I've used the second curl command and it returned a list of a whole lot of feeds, I guess it was successful

felixguendling:
I think this requires constant re-authentication which MOTIS can already do on its own:
https://github.com/motis-project/motis/pull/1050
https://github.com/motis-project/motis/pull/1421

I'm not sure if Transitous can already generate this though.

luna-cant-code:
I'm a bit dumb to be honest, hence my username.
Do I need to change the code or just leave this PR open until Transitous can generate that? 😅

felixguendling:
My guess would be that merging it with an access token might stop working after X hours. So that's probably not what we want.

I think it makes sense to wait with these feeds until someone finds time to add support for this to Transitous. The other option would be building a mirror that publishes the feed without OAuth.

luna-cant-code:
> So that's probably not what we want.

Nope, that's not really the goal here :sweat_smile: 

> I think it makes sense to wait with these feeds until someone finds time to add support for this to Transitous. The other option would be building a mirror that publishes the feed without OAuth.

I'd love to do this, but I'm only qualified enough for waiting, I'll close this PR for now. :)

Thank you!!!!

PR #2426, opened 2026-08-19
website: Allow routing to events

jbruechert:
Let's just do this as a gimmick.


vkrause:
Awesome idea :D Should this include the time as well?

jbruechert:
Oh true, at the very least the date would be nice :)

PR #2427, opened 2026-08-19
Add FrOSCon 2026 talk

vkrause:
Only just learned about this now, otherwise we would of course have added that earlier.


MichaelKutzner:
Thank you for adding the talk.
Sorry for not announcing it earlier. But there were no plans for this before last weekend.

PR #2441, opened 2026-08-24
Add Trenitalia license

pa-ulex:


jbruechert:
Just because the script to re-compress the file is CC0-1.0 doesn't make the resulting feed that license

pa-ulex:
Ah, OK. I didn't know that. I can't find a licence for it, so I assumed that it was licensed under CC0-1.0.

Lach-anonym:
About license the source site only says the presented data is open as required by European regulations, so reproducing it etc. is allowed

PR #2446, opened 2026-08-26
Clarify the usage policy

jbruechert:
We have a lot of people asking things that we can answer in the policy already.
Make sure to put the excluding factors right in the beginning so people who don't read the full text don't require manual support

image

### DIFF
--- a/website/content/api.html
+++ b/website/content/api.html
@@ -13,8 +13,39 @@ SPDX-License-Identifier: AGPL-3.0-or-later
 
 <h2>Usage Policy</h2>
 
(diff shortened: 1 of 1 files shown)


vkrause (approved):

felixguendling:
I think we can merge this now to get writ of requests from projects that we cannot support. In general we could try to shorten it a bit after that.

traines-source:
do we exclude commercial projects though? We do have some projects that are semi-commercial that IMO we do want to support (like cartes.app). That was the reason we had open-sourceness as our proxy criterion, right? The only thing that we should maybe introduce is the NC-filter.

PR #2447, opened 2026-08-26
call SkyTrain by its name

luna-cant-code:
Just a little change because it really triggers my inner neat freak.

I hope this is correct, I took some inspiration from the code above it but I have no means of testing if this is really correct.
(Is there even a way to deploy Transitous with all its bells and whistles locally for testing things like this?)


scripts/de-DELFI.lua
```diff
@@ -63,6 +63,9 @@ end
 function process_route(route)
     local agency_name = route:get_agency():get_name()
     local route_name = route:get_short_name()
+    -- Change SkyTrain Düsseldorf to its full name
+      if route:get_short_name = "SkyT" then
```
jbruechert:
```suggestion
    if route_name == "SkyT" then
```

felixguendling:
Scripts are particularly hard to maintain and it would help a lot if you could test your changes in a local MOTIS instance on this particular dataset. Please post a screenshot of the local MOTIS UI with a successful renaming applied. The code doesn't seem to be formatted anymore (which triggers my inner neat freak ^^).

luna-cant-code:
I'll set up MOTIS and send a screenshot, cheers! ☺️

jbruechert:
> I'll set up MOTIS and send a screenshot, cheers! ☺️

I recommend only setting it up for a single country, otherwise it might need more resources than a typical laptop has.

luna-cant-code:
image
Am I missing something? It doesn't really matter how I write the function, it just won't show up as "SkyTrain"
I don't even want to touch the fact that, according to what the API throws at me, the SkyTrain is not a funicular.

The route_short_name is _SkyT_ in routes.txt.

luna-cant-code:
~~Even if I omit `set:route_display_name`, MOTIS still doesn't budge.~~

~~Would it make more sense to open an issue on MOTIS?~~

I was dumb, I forgot to point MOTIS to the config file, though the file was wrong anyways

luna-cant-code:
image
It works! It's not beautiful but it works!

And I've honestly learned quite a lot about config files, MOTIS, and the fact that my server is too slow for anything

jbruechert:
Thank you!

Issue #1627, opened 2025-10-19
Weird headsigns in Austria


Lach-anonym:
I've noticed some trips in Linz Hbf have 0 as their headsign

![Image](https://github.com/user-attachments/assets/6b501c80-a466-446c-b805-e510140ed912)

An example trip

![Image](https://github.com/user-attachments/assets/df084b6b-f003-41a4-9590-1bd0207afd60)

https://api.transitous.org/?tripId=20251019_20%3A48_at-PTA-Upper-Austria-Flex-2025_60.So%2BFe.49-S5-j25-17.2.H

felixguendling:
If I remember correctly, we had this issue a few months ago already and it was really like this in the data (couldn't find the discussion). So probably something to report back and/or fix using Lua scripts (although currently there's no option to access a trip's stops when processing trips which would help here).

jbruechert:
In this case, it should be enough to remove the headsign and rely on the fallback to the last stop.

hoermalmeister:
At least in case of the line S5, it continues onto a different line, so that might be the issue. Because the real world headsign is not the last station of this line

felixguendling:
> In this case, it should be enough to remove the headsign and rely on the fallback to the last stop.

Let me know if this works or doesn't work. I'm not sure if the fallback will be used in case the Lua script sets it to an empty string.

jx-grxf:
I set up a minimal reproducible test for this against MOTIS v2.11.2 — a synthetic GTFS feed with three trips: one with `trip_headsign=0`, one with an empty headsign, and one with a valid headsign.

**The empty-headsign fallback works, but only for headsigns that are already empty when the GTFS is parsed:**

```
no script:    T_ZERO -> "0"     T_EMPTY -> "Endstation Nordwest"   T_OK -> "Echtes Ziel"
```

**`set_headsign("")` from Lua does not trigger it:**

```
with script:  T_ZERO -> ""      T_EMPTY -> "Endstation Nordwest"   T_OK -> "Echtes Ziel"
```

As a control I also blanked a trip that had a perfectly valid headsign — it comes out empty too, so this is not specific to the `0` value:

```
control:      T_ZERO -> ""      T_EMPTY -> "Endstation Nordwest"   T_OK -> ""
```

@felixguendling your suspicion was right — the fallback appears to be applied at parse time, before the script stage runs, so anything blanked from Lua stays blank.

[...]

PR #2449, opened 2026-08-26
Amend code of conduct with LLM-related behaviour

jbruechert:
My impression is that multiple people here are unhappy about the opposite of the new recommended actions, so worth recommending.

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -18,6 +18,8 @@ Examples of behavior that contributes to a positive environment for our communit
 * Giving and gracefully accepting constructive feedback
 * Accepting responsibility and apologizing to those affected by our mistakes, and learning from the experience
 * Focusing on what is best not just for us as individuals, but for the overall community
+* Seeking help from others instead of continuing on unreliable assumptions (including LLM output)
+* Communicating in your own words.
 
 Examples of unacceptable behavior include:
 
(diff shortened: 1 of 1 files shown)


vkrause (approved):
Make sense. I'd also support any more aggressive LLM ban if that's desired.

felixguendling:
> Make sense. I'd also support any more aggressive LLM ban if that's desired.

Which kind of LLM usage would you like to ban? I think if a JSON file adding yet another feed is written by a LLM or human doesn't matter much. I would guess the LLM makes less syntactic mistakes and otherwise the diff would be the same.

What I get annoyed about is usage of LLMs in human-to-human conversion. So "Communicating in your own words" could be phrased a bit more clearly. Something like

> Do NOT use LLMs to communicate to other members of the Transitous community. This applies for example to messages on Matrix, GitHub Issues, or Pull Requests.

jbruechert:
I added changes related to both of your comments.
So far not actually touching code related things much. It's very likely I'll form a more specific opinion on that once people start dropping unreviewable PRs here, but so far it hasn't really been an issue. I guess what I care about fundamentally is that people know what their changes do, and what each of the options they set do. I'm happy about suggestions for a specific text.

felixguendling:
For code contributions, I think it might be better to setup a `CONTRIBUTING.md` file. I think we also need this for MOTIS. In general, for code changes (especially from new contributors but probably even for long time contributors), I expect that the need for that code change and the strategy how to solve this problem should first be discussed in an Issue before opening a PR. Lately, I had to close https://github.com/motis-project/motis/pull/1556/changes in MOTIS which was a 11k LoC change nobody ever talked about. Open Source does not mean "Yeah! I found someone else to maintain my AI slop".

felixguendling:
I think we need this rather sooner than later ;)
https://github.com/public-transport/transitous/issues/1627#issuecomment-5471648825

I think CoMaps could be a role model for us:
Screenshot 2026-08-31 at 13 01 13
Source: https://2026.stateofthemap.org/sessions/DDJRVU/ (Slides)

https://codeberg.org/comaps/Governance/src/branch/main/AI_USAGE.md
https://codeberg.org/forgejo/governance/src/branch/main/AIAgreement.md

jbruechert:
Sounds good in general. I'd like to merge this now but I'm happy about follow-up MRs.

PR #2472, opened 2026-09-04
added temporary fix to Rheinbahn routes

luna-cant-code:
Currently, Rheinbahn metro lines are displayed as buses, that's an error existing for a couple of weeks now.
I've already opened a GTFS issue https://github.com/mfdz/GTFS-Issues/issues/313 and would like to propose, for the time being, temporarily fixing it in the DELFI feed.

I'll keep an eye out, once the error is fixed, I'll revert the change.

The code is ugly, it works though:

_before:_
image

_after:_
image

### DIFF
--- a/scripts/de-DELFI.lua
+++ b/scripts/de-DELFI.lua
@@ -63,6 +63,11 @@ end
 function process_route(route)
     local agency_name = route:get_agency():get_name()
     local route_name = route:get_short_name()
+  -- temp Rheinbahn fix (make buses to metro, GTFS issue is already open)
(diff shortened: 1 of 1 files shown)


jbruechert:
Since DELFI people already said they'd look into it, I'd rather just wait a week. I can't keep track of where we need to remove temporary fixups everywhere, so it would likely be forgotten.

luna-cant-code:
That's fair!
I'll mark this as a draft for the time being, keeping an eye out! :)
Thank you!

luna-cant-code:
Hat sich nun mit dem neuen Datensatz tatsächlich erledigt!

PR #2480, opened 2026-09-08
website: add Öffi-NG to apps that use transitous

santawho:
this change adds the link to the Android app "Öffi NG".

Öffi NG is a fork of Öffi https://gitlab.com/oeffi/oeffi.

Öffi is an Android app that tells you where and when trains and busses go, including delays and replacement bus service for more and more transport authorities in Europe and beyond.

A lot of new features have been added in this fork and continue to be added. One of these is the integration of Transitous as one of the data providers.


jbruechert:
Thanks, please run `reuse download --all`.

PR #2484, opened 2026-09-11
it: add ATM Messina GTFS-RT trip updates

davidedistaso:
Adds a GTFS-RT TripUpdates feed for ATM Messina, attached to the existing `Sicilia-Messina-ATM` static source (same `name`).

The feed is self-produced from ATM Messina's public "smart poles" departure boards and re-published as GTFS-RT at `https://rt.annacati.com/atm-messina/messina.pb`. Its `trip_id`s are matched against the same `GTFS_ATM_SPA.zip` this source already imports, so the updates attach to the existing trips. It is already live in "Annacati", another MOTIS-based routing service.

Data is CC-BY (Comune di Messina open data portal); attribution is set in the license block.

### DIFF
--- a/feeds/it.json
+++ b/feeds/it.json
@@ -311,6 +311,19 @@
             }
         },
         {
+            "name": "Sicilia-Messina-ATM",
+            "type": "url",
+            "spec": "gtfs-rt",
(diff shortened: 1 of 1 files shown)


jbruechert:
Thank you!

PR #2498, opened 2026-09-16
GTFS for Zamora, Michoacán, México

oxcar:
I added the GTFS for Zamora, Michoacán, México.

Also, I added myself as maintainer.


jbruechert:
Thank you!

PR #2501, opened 2026-09-17
fr: add realtime feed for Caen (Twisto)

kevinbioj:
This PR aims to add a fully functional GTFS-RT feed crafted by hand and fed by the network SIRI server.  
It is plug-and-play with the official GTFS resource that should already be consumed here.

Both vehicle positions and trip updates are available in the feed.  
You can see it in action [there](https://bus-tracker.fr/?network-id=27#-0.346,49.207,10.52).


jbruechert:
What is the difference for us over consuming the siri feed directly?

kevinbioj:
> What is the difference for us over consuming the siri feed directly?

Ideally there should be no differences, but in practice there are some.

1. Unless you are authenticated using an API key, the SIRI service restricts you to one request per 10 minutes (which basically makes any guest use impossible).
2. The `EstimatedTimetable` service returns no data. Thankfully, the `VehicleMonitoring` returns all onward calls for every vehicle which allows me to build trip updates. _So unless you are able to do the same, you won't be able to get trip information._
3. SIRI identifiers do not match GTFS identifiers, which should not be a problem if you are able to do fuzzy matching.

One or more of these points are probably the reason why there's no realtime information for Twisto right now @ api.transitous.org

jbruechert:
Okay, makes sense :)

PR #2503, opened 2026-09-19
Add Canada, Quebec, Gatineau (STO) feeds

wnagele:
Enables STO (Gatineau) in `ca-qc.json`:

- **Static GTFS:** un-skips `f-f244-sto`. The skip dates from the dead `www.contenu.sto.ca` host; the atlas record now points at `https://contenu.sto.ca/GTFS/GTFS.zip`, which is live, updated daily, and needs no authentication (verified).
- **GTFS-RT:** STO publishes TripUpdates and ServiceAlerts, but its endpoint requires an API key plus a per-request `SHA256(private_key + <UTC minute>)` signature, which no static URL can express. The feeds are therefore served through a small signing proxy at `https://proxy.octransparency.ca/ca/qc-STO/{TripUpdates,Alerts}` (30 s upstream caching keeps the load on STO at a few requests per minute; MOTIS polls get fresh data).

Both endpoints verified serving valid protobufs today.


jbruechert:
Thanks!

PR #2505, opened 2026-09-19
cz: Switch national CZPTT feed for improved version

Sap1k:
Hi!

I've been working on improving JrUtil for a good month or two, and I feel like my feed is now superior enough to be worth swapping over to from the currently existing feed.

The main improvements over the current implementation are as follows:

- The feed uses passenger-facing line numbers where they exist, and defines route colors based on train category.
- The conversion is no longer one PA -> one trip. We split trips based on passenger-facing markings, operators... and connect them into blocks
- Properly carries over bike/low-floor semantics
- Substitute transport has it's MOT correctly set to a bus (not falsely train anymore), with proper BUS platforms defined for these services.
- Does not drop timing points en route, instead marking them as boarding/deboarding disallowed (which is the correct way to represent this in GTFS)

This is a major change that's subject to discussion, and I am open to tweaking the feed further if needed.

We might not need "fix: true" anymore for this feed - what was the reason it was required here in the first place?

You can click around the feed [here](https://motis.obehy.cz)


Sap1k:
Attaching a MobilityData vaildation report: https://gtfs-validator-results.mobilitydata.org/7fc0cabb-41d2-4d57-b70e-361852fa2a26/report.html

jbruechert:
Thanks, Looks nice! But I'll leave this open for the Czech community for a bit because they can probably say more about this

hoermalmeister:
It looks a lot more tidy than our current feed, if you're sure we're not losing any data, I am open for implementation.
I'd be also open to replace our bus feeds with your jdf feed, but only with flixbus and pid filtered out.

Sap1k:
@hoermalmeister I still consider the JDF feed WIP and subject to change - I am open to swapping it, but probably at a later date.

I am not aware of there being any lost data here - looking at the run logs, no PAs are being dropped, and I've been sanity checking it for a few weeks and noticed no major issues.

Lach-anonym:
> I'd be also open to replace our bus feeds with your jdf feed, but only with flixbus and pid filtered out.

If the data quality is the same or higher, I'd also go for that

Sap1k:
@Lach-anonym The data quality is definitely higher, but the feed is more "opinionated", the key differences are as follows:

- The GPS coordinate coverage for stops has been greatly improved, decreasing the amount of stops with unknown coordinates to about 160 (from over 1k)
- JrUtil now rejects stop coordinates where the required travel speed makes the candidate stop clearly impossible
- The unknown stops are estimated to be at the midpoint between the previous and next stop (based on travel time), if it is a first or last stop, it gets shifted 500m to the north. All estimated stops are explicitly marked with [?] at the end of the stop name
- route_short_name is not the license number anymore - the passenger facing marking defined in LinExt is used, if that doesn't exist, the last three numbers from the license are used instead (which is usually the passenger-facing marking).
- GTFS route colors are now set, basically copies the PID color scheme
- International trips get filtered out, excluding regional trips (heruistic based on if any IDS zones are defined and trip length in/out of CZ)
- Stops now have a proper parent -> post hierarchy

There's also a WIP (and pretty slow) post estimator in JrUtil, which can estimate which posts trips use based on geometry and routing penalties, but it's nowhere near perfect and currently disabled.

What's left before I'm okay swapping it over?

- [ ] Create fixed stop ID registry to keep stop IDs between exports, some GTFS software struggles with changing IDs (Träwelling)
- [ ] Updating the stop coords to bring the coverage gaps to <75
- [ ] Decide whether the [?] + estimate is better than setting 0,0 (imo it is, but i'd like to hear other's thoughts on this)
- [ ] Possibly optimize the post estimator and decide if we want to use it or not
- [ ] Make the filtering of the feed for Transitous a togglable part of the pipeline

